PR #5, opened 2021-09-08
Merge the ElementData feature branch

timspainNERSC:
Merge the fully implemented ElementData branch into develop.


einola:
Minor comment: shouldn't the ``.clang-format`` file be in the root? It's in ``core/src`` now

timspainNERSC:
> Minor comment: shouldn't the `.clang-format` file be in the root? It's in `core/src` now

More like there shouldn't be a copy of `.clang-format` in both `core/src`and the root.

auraoupa:
I don't mind where it is as long as we change the workflow accordingly, for the moment it looks for it in src. That's been replaced by core right ?

timspainNERSC:
Since the workflow works correctly with the files as they are, I think we are good?

(I made a change that was not clang-format approved and the check failed as expected. And then I force pushed the commit revert. >:-) )

physics/src/modules/include/NextsimPhysics.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+    //! Calculate the new ice formed this timestep on open water
+    void newIceFormation(const PrognosticData& prog, const ExternalData& exter, PhysicsData& phys);
+    //! The thickness of newly created ice in the current timestep
+    inline double newIce() const { return m_newice; };
+
+    //! Total ice-atmosphere heat flux [W m⁻²]
+    inline double QIceAtmosphere() const { return m_Qia; };
+    //! Ice to ocean heat flux [W m⁻²]
+    inline double& QIceOceanHeat() { return m_Qio; };
```
einola, line 59: Is this better than declaring ``m_Qio`` public? You do this quite a lot, but to me the assignment ``phys.QIceOceanHeat() = 0.`` (for example) is then a bit counterintuitive, because you're assigning something to the result of a function call. But maybe that's just my Fortran background?

einola (approved):
This looks very nice :)

I've put in some comments we should discuss in the new year, but I don't think they're anything we should worry about for this PR. It's mostly for my benefit or something that could become an issue we deal with later.

einola:
> Since the workflow works correctly with the files as they are, I think we are good?
> 
> (I made a change that was not clang-format approved and the check failed as expected. And then I force pushed the commit revert. >:-) )

Sounds to me like we're good :)

auraoupa (approved):
I couldn't say about the code itself but on the continuous integration's point of view it's all good to me ! Sorry it took me so long to approve the merge, I thought it was done already ...

PR #28, opened 2021-10-29
Issue22: Timer infrastructure

timspainNERSC:
Implement more useful timer infrastructure.


pminak:
@timspainNERSC , I've tested Timer, and it works well, wall and cup times was important for us.
Hierarchy of time measurements is nice in your implementation, we don't have it. 
Now for the sake of tests we use your Timer in our example1.

Would it be possible for you to add, number of function call (activations) and time per call (readPerActivations),
see stopwatch.hpp in feat_dynamics.

We plan to remove our and use your time measurement infrastructure.

timspainNERSC:
I've added number of calls and wall time per call to the report. Feel free to take a look. Is that what you were envisaging?

pminak (approved):

PR #31, opened 2021-11-01
Issue15 ci workflow

auraoupa:
The workflow to check if the code  is clang formatted, to compile and to run the tests is now in place so we should run it on the develop branch


timspainNERSC (changes requested):
Is there a way of avoiding repeating the set of tests to be run in both the `tests-on-ubuntu` and `tests-on-mac` `run` sections? It would be better to not have to update the list of tests in two places if possible.

auraoupa:
Yes sure it makes sense !

timspainNERSC (changes requested):
I think it's definitely better to test on both ubuntu and mac, especially if they are using different compilers. It would be better to have both sets of testing with the testing steps repeated (as before) than to remove one merely to avoid code redundancy.

einola:
yes, I think commit 263ac6a (remove double testing) is a misunderstanding. We should test on both mac and ubuntu, so I'd like that commit to be reversed.

I think Tim's point was that it would be nice to avoid code repetition in the two test cases, but they are so different that I don't see that this is possible.

auraoupa:
ok sorry for the misunderstanding, I will reverse this commit.
maybe instead of launching each test one by one we could make a file that lists all the test and that will be run on each platform ? this way when adding a new test , we just have to add it to this file ?

einola:
> ok sorry for the misunderstanding, I will reverse this commit.

No worries :)

> maybe instead of launching each test one by one we could make a file that lists all the test and that will be run on each platform ? this way when adding a new test , we just have to add it to this file ?

Yes, that sounds very good! And more like what @timspainNERSC had in mind, I think.

timspainNERSC:
Yes, that is more what I was thinking of.

I believe that Catch provides a way of managing individual tests in a test suite. This may be a way of achieving this goal, but I have never used it, so I don't know whether it is appropriate for our situation.

einola:
> I believe that Catch provides a way of managing individual tests in a test suite. This may be a way of achieving this goal, but I have never used it, so I don't know whether it is appropriate for our situation.

Ok. I suggest we approve and merge now and put up an issue about checking if Catch is a better way to do this.

timspainNERSC (approved):

auraoupa:
I will investigate the Catch solution, in the mean time I'm manually adding the tests present in develop to the CI script, if that is ok with you ?

timspainNERSC:
That will be absolutely fine.

PR #45, opened 2021-12-07
WIP: Physics subdirectory

timspainNERSC:


einola (commented):
Not sure if you want comments now - but you'll get them anyway! :)

I like the structure and I think it's quite clear, especially for the ``physics`` directory. What I find a bit confusing is the naming of the ``src`` directory. I'd rename it something like ``core``, ``base``, or even ``infrastructure`` - something to indicate that this is the model infrastructure, contrary to the physics and dynamics, that will live in their own directories.

timspainNERSC:
I did think that the current `src` directory should be renamed to parallel the `physics`and `dynamics` directories, but could not come up with a good name last week. I think `core` is just right, though.

timspainNERSC:
Any outstanding comments can go on PR #5 .

PR #61, opened 2022-01-05
Issue60 Physics Module

### LINKED ISSUE #60: NextsimPhysics should be a module
It turns out that including NextsimPhysics as a template argument isn't that pleasant to use. However, it should be usable as a regular module with only a few changes.

timspainNERSC:
Implements the physics calculation class as a module, rather than a template parameter.

Will close #60.


einola (approved):
Looks like a very nice solution to me!

PR #68, opened 2022-02-03
Add logging class and library

timspainNERSC:


timspainNERSC:
Definitely still current!

PR #70, opened 2022-02-04
add link to install guide

dorchard:
I thought this might be useful?


einola (approved):
Good point - thanks! :)

PR #73, opened 2022-02-07
Use CMake imported target for Boost dependency

athelaf:
When both variables and imported targets are provided for
packages, projects should prefer the latter due to their
superior robustness and better integration with CMake’s
transitive dependency features.


timspainNERSC (commented):
This change will need to be made to all the tests in core/test and physics/test that use Boost. I think this would be better done as part of this PR than separately.
The change looks great otherwise.

athelaf:
Created a new pull request for issue71.

PR #74, opened 2022-02-10
Develop

timspainNERSC:
Merge the Minimum Viable Model state of `develop` into `main` to be tagged as the 0.1.0 release.


auraoupa:
I can approve it or maybe @einola you want to have a look before ?

timspainNERSC:
@auraoupa Could you take a look at it to and see if you think there is enough documentation in the new code? @einola  and I were just discussing this, so I'm fairly sure that he is happy with it from a structure & logic point of view.

auraoupa:
Ok I'll take time this afternoon to have a long look at it

auraoupa:
Ok so I've merged develop into docs so it triggers all the documentations : doxygen : https://nextsimdg.github.io/nextsimdg/index.html and readthedocs (especially the API in it) : https://nextsim-dg.readthedocs.io/en/latest/api/library_root.html. Nothing strikes me but you can have a look too ! Maybe it would be a good timing to describe the structure of the model and a minimal test case in https://github.com/nextsimdg/nextsimdg/blob/docs/docs/getting_started.rst ?

auraoupa (approved):

Issue #78, opened 2022-02-14
Crash with no command line options


einola:
The model crashes with an unintelligible error if no options are supplied:

```
Skopje:build einola$ ./nextsim 
libc++abi: terminating with uncaught exception of type netCDF::exceptions::NcException: NetCDF: Malformed URL
file: ncFile.cpp  line:88
Abort trap: 6
Skopje:build einola$ 
```

I suggest we also print the help message if no command-line options are supplied (in addition to printing it if we call the model with -h or --help).

PR #79, opened 2022-02-14
Address issue where model crashes without command line options (#78)

einola:
I added a simple check for the length of ``m_arguments`` in ``CommandLineParser.cpp``. This means the help message is printed when no command line options are given.

This is just me familiarising myself better with the code :)


timspainNERSC (approved):
Excellent.
(And that's a rather embarrassing thing to have missed on my part!)

PR #83, opened 2022-02-21
New module system

### LINKED ISSUE #59: Modules should be separable
While writing the test for `PrognosticData` I came across the need to include all the modules from the physics subdirectory when just trying to write code only involving core classes.

Some way needs to be available to only use the core (or only the physics) modules.

Currently this is fixed by using alternative hand-written `ipp` files when compiling the tests.

timspainNERSC:
Python free and more self-contained.

Will close #59.


timspainNERSC:
Well, python free in the important sense. I am writing a script to auto generate the boiler-plate heavy module class files.

core/src/modules/README.md
```diff
@@ -1,75 +0,0 @@
-# ModuleLoader
```
einola, line 1: Don't we still want a README here?

einola (changes requested):
Looks good to me. But I miss that README file. There should still be one to tell developers how to develop a new module.

core/src/modules/README.md
timspainNERSC: Almost certainly.

einola (approved):

PR #84, opened 2022-02-22
Isssue81 netcdf cxx discovery

draenog:
This pull request tries to fix the issue of discovery of netcdf-cxx from cmake when the library was not installed using cmake (which is common on HPC systems). There are still some corner cases when the process is not able to find netcdf-cxx but it should not introduce any new regressions.


timspainNERSC (approved):
Looks great

CMakeLists.txt
```diff
@@ -4,14 +4,22 @@ set(CMAKE_CXX_STANDARD 17)
...
-    set(NSDG_NetCDF_Library "netcdf_c++4")
+    find_package(netCDF REQUIRED)
+    if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Darwin")
+        set(NSDG_NetCDF_Library "netcdf-cxx4")
+    else()
+        set(NSDG_NetCDF_Library "netcdf_c++4")
+    endif()
 endif()
 
+
```
einola, line 22:
```suggestion
```

PR #87, opened 2022-02-24
Feat dynamics perf

athelaf:
I have eliminated unnecessary virtual functions and achieved performance improvements in the projections functions by manually performing common sub-expression elimination that the compiler cannot perform (the compiler cannot access that the involved functions have no side effects). I also applied Clang formatting.


pminak (approved):

pminak:
@athelaf perhaps you could help us with clang formatting. 

We included clang format in CMakeLists.txt, here
https://github.com/athelaf/nextsimdg/blob/343973a418a8e7ddc809d3b8132b7814ca83a1cd/dynamics/CMakeLists.txt#L75

The difference in format seems to be fact the we do not specify line length at 100, but use default 80. 
What would be your suggestion and how do you force clang format yourself?

Issue #82, opened 2022-02-18
Self-contained modules


einola:
We should try and have the modules as self-contained as possible. If a model contains a single physical process, then different formulations or parameterisations of that process may have different complexities. This leads to different module implementations for the same process (like albedo calculation) that can have

 - a different number of variables that need to be advected (prognostic variables)
 - a different number of variables that don't need to be advected, but could be interesting to output (diagnostic variables)

Ideally, a module would do the following:

```
Get inputs (probably by reading PrognosticData/ElementData on input)
Do physics and modify prognostic and diagnostic variables
Broadcast diagnostic outputs (to XIOS)
Return modified diagnostic variables (presumably by modifying the input PrognosticData/ElementData)
```

timspainNERSC:
It turns out I needed to read the issue text more carefully. PR #83 is an important step towards solving this, but far from the only component.

timspainNERSC:
ModelModule implements enough of this to close for now, I think.

PR #90, opened 2022-03-11
WIP: ModelModule

timspainNERSC:
Modularizing the model components to achieve the goals in issue #82. Moving away from the ElementData concept to a more traditional set of 2d arrays (ModelArray).

Different from but related to the modules defined in `Module.hpp`.


dorchard:
https://github.com/nextsimdg/nextsimdg/blob/4c518967a403407d8731b962347b256f377d2335/core/src/ModelArray.cpp#L220 appears to have a bug? I think this should be

        size_t dLast = d[7];

timspainNERSC:
> https://github.com/nextsimdg/nextsimdg/blob/4c518967a403407d8731b962347b256f377d2335/core/src/ModelArray.cpp#L220
> appears to have a bug? I think this should be
> 
> ```
>     size_t dLast = d[7];
> ```
> 
> (unless I have misunderstood the code!)

It was, of course, a copy-paste error. Corrected.

dorchard:
> > https://github.com/nextsimdg/nextsimdg/blob/4c518967a403407d8731b962347b256f377d2335/core/src/ModelArray.cpp#L220
> > 
> > appears to have a bug? I think this should be
> > ```
> >     size_t dLast = d[7];
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (unless I have misunderstood the code!)
> 
> It was, of course, a copy-paste error. Corrected.

If the `const size_t* d` parameter came first then it would be easier to collapse these into a single variadic function instead (avoids any temptation for such mistakes and then gives you arbitrary generalisation and a smaller code footprint).

core/src/ModelArray.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+ModelArray& ModelArray::operator=(const double& fill)
+{
+    m_name = std::to_string(fill);
+    setData(fill);
+
+    return *this;
+}
+
```
athelaf, line 64: Hi Tim, I noticed all the math operators return new ModelArray instances. I was wondering if it would be useful to add in-place operators as well.

core/src/ModelArray.cpp
```diff
@@ -0,0 +1,332 @@
...
+// Simple 1d indexing
+inline size_t indexr(const size_t* dims, const size_t nDims, size_t i) { return i; }
+
+// Special case for 2d indexing
+inline size_t indexr(const size_t* dims, const size_t nDims, size_t i, size_t j)
+{
+    return i * dims[1] + j;
+}
+
+// General case for up to 8d indexing
```
athelaf, line 193: If this function is supposed to compute the linear index in multiple dimensions, I think it does not produce correct results for more than 2 dimensions (might be wrong here). Also, I suggest the use of variadic templates in this case. For example, you can write something like (there is a way to write this without recursion):

`// Base case
template<typename T>
T index(const T *dims, T ndims, T v) {
  return v;
}

template <typename T, typename... Args>
T index(const T *dims, T ndims, T first, Args... args)
{
    // FIXME: check number of dimensions match number of indices
    T dim_prod = 1;
    for (int i = 0; i < ndims - 1; i++)
        dim_prod *= dims[ndims - i - 1];
    return first*dim_prod + index(dims, ndims - 1, args...);
}`

core/src/ModelArray.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+// Simple 1d indexing
+inline size_t indexr(const size_t* dims, const size_t nDims, size_t i) { return i; }
+
+// Special case for 2d indexing
+inline size_t indexr(const size_t* dims, const size_t nDims, size_t i, size_t j)
+{
+    return i * dims[1] + j;
+}
+
```
athelaf, line 192: If this function is supposed to be implementing linear indexing in multi-dimensional arrays, I think it produces wrong results for more than 3 dimensions. Also, I suggest you variadic templates to implement this. For example:

```
// Base case
template<typename T>
T index(const T *dims, T ndims, T v) {
  return v;
}

template <typename T, typename... Args>
T index(const T *dims, T ndims, T first, Args... args)
{
    // FIXME: check number of dimensions match number of indices
    T dim_prod = 1;
    for (int i = 0; i < ndims - 1; i++)
        dim_prod *= dims[ndims - i - 1];
    return first*dim_prod + index(dims, ndims - 1, args...);
}

```
There is a better way to implement this without recursion.

core/src/ModelArray.cpp
athelaf, line 193: I suggest we use variadic templates to implement this. For example:

// Base case
template<typename T>
T index(const T *dims, T ndims, T v) {
  return v;
}

template <typename T, typename... Args>
T index(const T *dims, T ndims, T first, Args... args)
{
    // FIXME: check number of dimensions match number of indices
    T dim_prod = 1;
    for (int i = 0; i < ndims - 1; i++)
        dim_prod *= dims[ndims - i - 1];
    return first*dim_prod + index(dims + 1, ndims - 1, args...);
}

There is a better way to implement this without recursion.

core/src/ModelArray.cpp
```diff
@@ -0,0 +1,332 @@
...
+        stepLength *= dims[1];
+    case 1:
+    default:
+        out += stepLength * i;
+        return out;
+    }
+    return 0;
+}
+
+size_t indexr(const ModelArray::Dimensions& loc, const size_t* dims)
```
athelaf, line 229: This one assumes loc.size()=8, otherwise it will crash correct?

core/src/ModelArray.cpp
```diff
@@ -0,0 +1,332 @@
...
+    case (6):
+        return (*this)(dims[0], dims[1], dims[2], dims[3], dims[4], dims[5]);
+    case (7):
+        return (*this)(dims[0], dims[1], dims[2], dims[3], dims[4], dims[5], dims[6]);
+    default:
+        return (*this)(dims[0], dims[1], dims[2], dims[3], dims[4], dims[5], dims[6], dims[7]);
+    }
+}
+
+const double& ModelArray::operator()(size_t i, size_t j) const
```
athelaf, line 256: With the variadic template approach I think we can then replace all the operator() calls with a single call like this:

```
template <typename... Args>
const double& operator()(Args... args)
{
    //return m_data[index(dimensions().data(), dimensions().size(), args...)];
    return m_data[index(dimensions().data(), args...)];
}
```

athelaf: Need to check if it is legal to overload the operator though.

athelaf: My understanding is you can.

core/src/ModelComponent.cpp
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @date Feb 28, 2022
+ * @author Tim Spain <timothy.spain@nersc.no>
+ */
+
+#include "include/ModelComponent.hpp"
+
+namespace Nextsim {
+
+std::map<std::string, ModelComponent*> ModelComponent::registeredModules;
```
athelaf, line 12: I suggest to use unordered_map if ordering is not required for correctness, as it has better time complexities for most operations.

athelaf: Same with set and unordered_set

core/src/include/ModelArray.hpp
```diff
@@ -0,0 +1,456 @@
...
+    //! Retuns the (enum of) the ModelArray::Type of this.
+    const Type getType() const { return type; }
+
+    /*!
+     * @brief Sets the number and size of the dimensions of a specified type of
+     * ModelArray.
+     *
+     * @param type The type of array the dimensions are to be specified for.
+     * @param dim The per-dimension size to be set.
+     */
```
athelaf, line 150: What is the motivation for this being static? Will all model arrays of each type have the same dimensions?

core/src/ModelArray.cpp
timspainNERSC: Oh, yes it will. Fixed.

core/src/ModelArray.cpp
timspainNERSC: I have adapted this for a new version of indexr

core/src/ModelArray.cpp
timspainNERSC: Done

athelaf: Hi Tim, by in-place I meant to add non-const operators that modify the ModelArray's internal data in-place and do not create a new ModelArray instance. This might be useful as well.

PR #98, opened 2022-05-12
Issue96 example names

auraoupa:
I cannot commit to tag version v0.1.0 so I would like to merge the changes in the documentation (not only about the simple example but also some work has been done for the front page) and the simple example to the main branch so I can create a new tag maybe v0.1.0b or v0.1.1 ? Or maybe no tag is necessary, tell me what you prefer


docs/index.rst
```diff
@@ -6,7 +6,37 @@ neXtSIM_DG : the next generation sea-ice model with a Discontinuous Galerkin met
 Introduction
 ------------
 
-`neXtSIM_DG` is a C++ model the simulates the sea ice ...
+**neXtSIM is a Lagrangian dynamical-thermodynamical sea-ice model that faithfully represents large-scale sea ice thickness, concentration and drift patterns** [1]_ [2]_ **, statistical properties of pack ice deformation and drift** [3]_ [2]_ **, diffusion and dispersion** [4]_ [5]_ **; and additionally exhibits a strong capacity for short-term sea ice forecasting at Pan-Arctic scales** [6]_ **. neXtSIM was developed to explore the role of sea-ice mechanics and dynamics in determining the large-scale behaviour of Arctic sea ice. As such, it has served as a testing and development ground for the brittle sea-ice rheologies developed over the past decade** [7]_ [8]_ [9]_ [2]_ **.** The initial development of the model was focused on understanding and representing as well as possible the impact of the new physics developments, which was the main rationale for the Lagrangian advection scheme used. As the new physics developments matured it became evident that these could impact not only the ice, but also the ocean and atmosphere below and above it [10]_.
```
timspainNERSC, line 9: Starting the sentence, paragraph and section with an unexpanded, uncapitalized acronym looks really odd. Perhaps 'The neXt generation Sea Ice Model (neXtSIM) is a Lagrangian…'

timspainNERSC (changes requested):
I think this is all fine (especially at this stage of the project), with the exception of the very start of the Introduction.

docs/index.rst
auraoupa: Yes I agree, I've updated it. If the next release is near, no need to do one for this then ! I just will make sure these changes are taken by merging it to main.

PR #104, opened 2022-06-02
Issue103 ci

auraoupa:
I failed to manage our compiling environment with only conda on both platforms.
At least I was able to fetch a particular version of catch2 for mac platform which should work with current developments but further work will be needed in the future
Did also point to a specific version of clang-format

### DIFF
--- a/.github/workflows/clang-compile-tests.yml
+++ b/.github/workflows/clang-compile-tests.yml
@@ -42,7 +42,7 @@ jobs:
         conda install libgcc
         conda install netCDF4=1.5.7
         conda install -c conda-forge boost=1.78.0
-        conda install -c anaconda cmake=3.23.1
+        conda install -c anaconda cmake=3.22.1
         conda install -c conda-forge catch2=2.9.13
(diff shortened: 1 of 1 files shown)


timspainNERSC:
Those raw github hashes look really brutal. These commands look better to me
```
brew extract --version=2.13.9 catch2 homebrew/cask
brew install catch2@2.13.9
```
*if* they work in the CI environment.

The commands are taken from this Stack Overflow answer: https://stackoverflow.com/a/66458452

timspainNERSC:
> ```
> brew extract --version=2.13.9 catch2 homebrew/cask
> brew install catch2@2.13.9
> ```

Does a version of the above commands work for clang-format, too? The relevant version looks to be 14.0.4

```
brew extract --version=14.0.4 clang-format homebrew/cask
brew install clang-format@14.0.4
```

timspainNERSC (approved):
Much neater, and hopefully less fragile.

auraoupa:
This is much nicer indeed
I was trying with catch2 first to see if that was working and now I've put it for clang-format as well

PR #130, opened 2022-08-23
ConfigOutput

timspainNERSC:
A (slightly) more complex DiagnosticOutput class that allows the frequency and initial output time to be controlled by configuration, as well as the contents of the files.


einola (approved):
Great - thanks :)

PR #137, opened 2022-08-31
Issue133 Time for parameters

### LINKED ISSUE #133: Passing time values to parameterizations
Parameterizations (such as ice albedo) might be time dependent. Currently they have no canonical way of accessing or being passed the current model time.

Fix this.

timspainNERSC:
Actually just IIceAlbedo. and it doesn't do anything there. Yet.

Closes #133


physics/src/modules/include/IIceAlbedo.hpp
```diff
@@ -21,6 +23,12 @@ class IIceAlbedo {
      * @param snowThickness The true snow thickness on top of the ice.
      */
     virtual double albedo(double temperature, double snowThickness) = 0;
+
+    /*!
+     * Sets the time parameter for the implementation, if it is time dependent.
+     * @param time The desired TimePoint.
+     */
+    virtual void setTime(const TimePoint& tp) { }
```
einola, line 31: Wouldn't something along those lines make sense?

```suggestion
    virtual void setTime(const TimePoint& tp) { M_tp = tp; }
    
    private:
        TimePoint M_tp;
```

einola (commented):
Question if we should make ``tp`` (or rather ``M_tp`` then) a class variable for ``IIceAlbedo``?

einola (approved):

physics/src/modules/include/IIceAlbedo.hpp
timspainNERSC: That's a detail for each implementation to solve. None of the existing implementations are time dependent, so they inherit the do-nothing Base class version of the function.

einola: Fair enough!

PR #139, opened 2022-08-31
Issue135 Remove the ocean from restart files

### LINKED ISSUE #135: Restart files can't
Since RectGridIO unconditionally reads `sst` and `sss` fields, and the written restart files do not contain these fields, trying to read a restart file written by the model fails. 

Fix this. By removing the restart file dependency on SST and SSS. These should be provided by OceanState implementations instead.

timspainNERSC:
Make it so that restart files can restart the model.

Closes #135


run/config_simple_example.cfg
```diff
@@ -1,5 +1,5 @@
 [model]
-init_file = init_rect30x30.nc
+init_file = init_rect3030.res.nc
```
einola, line 2: Why did you change this? Was it a mistake?

```suggestion
init_file = init_rect30x30.res.nc
```

einola (changes requested):
A surprising change in ``run/config_simple_example.cfg``. Otherwise fine.

einola (approved):

Issue #129, opened 2022-08-23
Configured simple atmosphere and ocean state


einola:
At the moment we have "dummy" atmosphere and ocean state modules. The DummyAtmosphereState has a hard-coded atmosphere state and DummyOceanState reads its state from the restart file.

We also need slightly more sophisticated dummy states, where the values are read in from a config file.

timspainNERSC:
How sophisticated do you need? Constant values? Spatially uniform monthly samples? `rand() / a - T0`?

einola:
Let's start with constant values, read in from a config file :)

The most complicated variant I can imagine is someone writing an Atmosphere or OceanState with an analytical forcing that can take in some optional parameters. But in this case the "constant values from options" version can act as a template.

PR #141, opened 2022-09-01
Issue129 configclimatology

timspainNERSC:
Climatologies that provide constant values, but can be configured.

Closes issue #129


einola (changes requested):
... again the change in ``run/config_simple_example.cfg`` :)

einola (approved):

timspainNERSC:
The python script `make_init.py` is supposed to regenerate the restart file used in `config_simple_example.cfg`. The names of the generated file, the restart file in the script and the file committed to the repo should also match. I have chosen the name `init_rect30x30.nc`, which I think is a revert.

PR #142, opened 2022-09-01
Issue140 Ice calculation

### LINKED ISSUE #140: Initial ice depends on a call to aoState.update()
Correctly filling the true ice thickness at the start of the time step requires a call to `AtmosphereOceanState.update()`. Is this the correct place for it?

This arose when @einola added a new fixed fluxes implementation of `IFluxCalculation`. The lack of update on the initial true ice thickness (i.e. the `ModelArray` accessed as `ProtectedArray::HTRUE_ICE`) means it is never filled with data unless `AtmosphereOceanState.update()` is called.

Since fixed fluxes should probably be feasible without updating the atmosphere & ocean, where would be the correct place for this calculation in code and to hold the array in the program structure?

The same applies to `ProtectedArray::HTRUE_SNOW`, but the two will almost certainly move together.

timspainNERSC:
Move the calculation of the true (ice averaged) ice and snow thicknesses from `AtmosphereOceanState` to `IceGrowth` as the latter might not be called. If `IceGrowth::update()` is not called then the ice thicknesses are probably not needed.

Closes #140

### DIFF
--- a/physics/src/IceGrowth.cpp
+++ b/physics/src/IceGrowth.cpp
@@ -42,7 +42,6 @@ IceGrowth::IceGrowth()
 
     registerProtectedArray(ProtectedArray::HTRUE_ICE, &hice0);
     registerProtectedArray(ProtectedArray::HTRUE_SNOW, &hsnow0);
-
 }
 
 void IceGrowth::setData(const ModelState::DataMap& ms)
@@ -132,8 +131,8 @@ void IceGrowth::update(const TimestepTime& tsTime)
     // Copy the ice data from the prognostic fields to the modifiable fields.
     // Also divide by c_ice to go from cell-averaged to ice-averaged values.
     cice = cice0;
-    hice = hice0 = hSnowCell / cice0;
-    hsnow = hsnow0 = hIceCell / cice0;
+    hice = hice0 = hIceCell / cice0;
+    hsnow = hsnow0 = hSnowCell / cice0;
 
     iFluxes->update(tsTime);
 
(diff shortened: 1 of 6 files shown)


physics/src/IceGrowth.cpp
```diff
@@ -124,12 +127,14 @@ ConfigMap IceGrowth::getConfiguration() const
...
 {
-    iFluxes->update(tsTime);
+
     // Copy the ice data from the prognostic fields to the modifiable fields.
     // Also divide by c_ice to go from cell-averaged to ice-averaged values.
     cice = cice0;
-    hice = hice0;
-    hsnow = hsnow0;
+    hice = hice0 = hIceCell / cice0;
+    hsnow = hsnow0 = hSnowCell / cice0;
```
einola, line 135: Shouldn't we catch the ``cice0 = 0`` case here? The suggestion below has perhaps a bit many ``=`` signs to be readable? o_O

```suggestion
    hice = hice0 = cice0 == 0. ? 0. : hIceCell / cice0;
    hsnow = hsnow0 = cice0 == 0. ? 0. : hSnowCell / cice0;
```

einola (changes requested):
Potentially a division by zero in one place. Otherwise good.

physics/src/IceGrowth.cpp
timspainNERSC: A good point, but these variables are whole arrays, not scalar values, so using the plain ternary operator won't work.

einola: Well, I suppose it'll have to be a loop then.

timspainNERSC: Implemented as an `overElements` loop.

einola (approved):

PR #144, opened 2022-09-06
Merge in newmesh

timspainNERSC:
Just taking a look at the conflicts/other damage


winzerle (approved):
* createdistortedrectangle.py script is just for testing. Does not need to stay in the repo
* dgInitial is most likely obsolete. I will check
* I introduced Nextsim::FloatType. Wanted to discuss this with you. I guess it would be good to have an option to change between float and double. What do you all think? And you to do it? best: choose inc make
* stopwatch.hpp/cpp shape be replaced by the next sim-timer

PR #155, opened 2022-09-08
Issue151 integration

pminak:


timspainNERSC:
It looks like `ubuntu-latest` is not the latest version of Ubuntu, which does have Eigen 3.4.

Indeed, it is still 20.04LTS `focal`, which only has Eigen version 3.3.7.

dynamics/CMakeLists.txt
```diff
@@ -1,113 +1,16 @@
...
-
-################################################## Compile the test cases
+add_subdirectory(applications)
 add_subdirectory(tests)
 
-################################################## Compile larger application examples
-add_subdirectory(applications)
 
+set(NextsimSources
+    "${NextsimSources}"
```
timspainNERSC, line 11: If you are not (yet) changing the entries in `NextsimSources`, these lines should be removed as redundant. (Also lines 14-16)

dynamics/src/CMakeLists.txt
```diff
@@ -7,12 +12,16 @@ set (INC "${CMAKE_CURRENT_SOURCE_DIR}/include")
...
-INCLUDE_DIRECTORIES(${Nextsim_INCLUDE_DIRS})
-
-add_library(${NextsimDGLib} ${Nextsim_Sources})
-
-target_link_libraries (${NextsimDGLib} Eigen3::Eigen)
-
 
+set(NextsimSources
+    "${NextsimSources}"
+    "${Sources}"
```
timspainNERSC, line 18: Again, remove these set( PARENT_SCOPE) commands.

timspainNERSC (changes requested):
I don't think we should include the set(PARENT_SCOPE) commands until they are actually being used.

einola:
Looks like @pminak has addressed the issues raised by @timspainNERSC - right guys?

dynamics/src/CMakeLists.txt
timspainNERSC: Okay, fine.

timspainNERSC (approved):

PR #161, opened 2022-09-20
Added OpenMP to CMakeLists.txt

pminak:


timspainNERSC (approved):
That looks right for OpenMP in CMake, as far as I remember.

PR #163, opened 2022-09-22
Issue160 New ModelArrayRef

### LINKED ISSUE #160: Rework ModelArrayRef
Given the changes likely required by #154, `ModelArrayRef` should not be dependent on `ModelComponent`.

This can be achieved by removing direct references to the `ModelComponent` arrays of `ModelArray`s and replacing them with second template argument pointing to a backing store. This will make the variable declaration a little less clean:
```
ModelArrayRef<ProtectedArray::H_ICE> hice;
```
will become something like
```
ModelArrayRef<ModelComponent::H_ICE, ModelComponent::ProtectedArrays> hice;
```
instead.

Also, the enumeration will likely have to be a trad enumeration instead of a C++ enum class, unless there is something clever to be done with class inheritance there. Sadly.

timspainNERSC:
Frees ModelArrayRef from dependence on ModelComponent, at the expense of more syntax.

closes #160


core/src/include/ModelArrayRef.old.hpp
```diff
@@ -0,0 +1,265 @@
+/*!
```
einola, line 1: did you want to keep this file in the repository - or is it an accidental commit?

einola (approved):
Looks sensible.

Check if you want to keep ``ModelArrayRef.old.hpp`` in the repository before merging.

core/src/include/ModelArrayRef.old.hpp
timspainNERSC: Of course I didn't want to keep it

Issue #154, opened 2022-09-08
Move snowfall to IFluxCalculation


einola:
For my work on issue #127 having the shared variable ``snowfall`` in ``IFluxCalculation`` makes more sense than having it in ``AtmosphereState``. I think this is true in general. Can we move ``snowfall``?

timspainNERSC:
Snowfall makes sense to me as something the atmosphere itself is doing. Instead, might it make sense to have a total precipitation flux or even E-P areal mass flux in the implementations of IFluxCalculation? This seems more 'flux-y' and would parallel the existing `subl` sublimation mass flux.

einola:
Yes, the problem is that snowfall kind of belongs in both places. You get either

 - Fluxes (total radiative and turbulent and evaporatoin) + precipitation (rain + snow)
 - Atmosphere state (2 m temp, etc) + precipitation (rain + snow) -> then calculate fluxes and evaporation

It doesn't make much difference where the snowfall sits, as long as both the flux and atmosphere state codes can modify it.

But can I do that when it's a ``ProtectedArray``? A simple

```
    registerProtectedArray(ProtectedArray::SNOW, &snowfall);
```

in the header and

```
    snowfall[i] = 1.;
```

doesn't work.

timspainNERSC:
> But can I do that when it's a ProtectedArray? A simple
> 
> `    registerProtectedArray(ProtectedArray::SNOW, &snowfall);`
> 
> in the header and
> 
> `    snowfall[i] = 1.;`
> 
> doesn't work.

In what way doesn't work? A ProtectedArray is only protected against anything that accesses it as a ModelArrayRef from writing to the array. In the ModelComponent where it is defined as a ModelArray, there is no concept of protected or shared or write-protected.

What does need to be the case is that the ModelArray is persistent, such as being a class member variable in a class that is still alive. It won't work on variables only defined within a function. (Assuming that's what is going on.)

einola:
I was trying to modify a protected array - which probably should not be allowed anyway :)

This all becomes a bit annoying and contorted because there are a few layers where information may be coming into the model. What we've done so far is

```
AtmosphereState <= 2 m temp, etc + precipitation from file or coupler
         ||
         \/
FluxCalculation
         ||
         \/
IceThermodynamics <= Fluxes (total radiative and turbulent and evaporation) + precipitation (rain + snow) from file or coupler
```

[...]

PR #170, opened 2022-09-28
Issue154 atmosphere and ocean boundary conditions

timspainNERSC:
Changes the data layout and flow around the external boundaries of the model, removing the AtmosphereState and OceanState classes, replacing them with the `IAtmosphereBoundary` and `IOceanBoundary` interfaces.

Closes issue #154


einola (approved):
Looks fine to me. But you've deleted ``IceGrowth_test.cpp``. Was that on purpose?

timspainNERSC:
I think I made enough changes to `IceGrowth_test.cpp` that the diff was rolled up, which also happens to a deleted file. At least that's how the 'Files changed' tab displays to me.

Issue #168, opened 2022-09-28
FreezingPoint in ThermoIce0


timspainNERSC:
The function `ThermoIce0::calculateElement()` uses the function `FreezingPointModule::getImplementation()`. Does it need to? I would be able to remove the Module source file from IceGrowthTest if it does not.

timspainNERSC:
No, the temperature i question should be the (fixed) melting point of the sea ice, not the freezing point of the sea surface.

timspainNERSC:
This needs to wait until after #170 is merged.

PR #173, opened 2022-09-29
Issue168 ice0tf

timspainNERSC:
Use the correct temperature when all the ice melts away. Or at least the same one as the `finiteelement.cpp` implementation.

Closes issue #168


timspainNERSC:
As noted in the issue notes, reopening now that #170 has been merged.

einola (approved):

Issue #126, opened 2022-08-22
Pull request description into change log


einola:
Ok, I think we'll just stick with the three numbers we have now, but they're not exactly semantic versioning then. So for x.y.z we have

 - increment of z when there's a hotfix (bugfix) pushed to main
 - increment of y when develop is merged into main
 - increment of x when we've been working on something extraordinary, when there's a major change in interface (configs or outputs) ... or when y is becoming too big (26 is industry standard I believe).

We need to

 - review the read-me and make sure that this is what it says
 - set up a system where each pull request has a headline and description and that these go into the change log, together with the version number.

_Originally posted by @einola in https://github.com/nextsimdg/nextsimdg/issues/121#issuecomment-1222109945_

TomMelt:
I think this has been closed by #177. @auraoupa & @einola are you ok to close this issue now?

einola:
Yes, closing this is fine.

PR #177, opened 2022-10-03
Implement changelog generation

### LINKED ISSUE #125: review the read-me and make sure that the description of version numbering is ok
Ok, I think we'll just stick with the three numbers we have now, but they're not exactly semantic versioning then. So for x.y.z we have

 - increment of z when there's a hotfix (bugfix) pushed to main
 - increment of y when develop is merged into main
 - increment of x when we've been working on something extraordinary, when there's a major change in interface (configs or outputs) ... or when y is becoming too big (26 is industry standard I believe).

We need to

 - review the read-me and make sure that this is what it says
 - set up a system where each pull request has a headline and description and that these go into the change log, together with the version number.

_Originally posted by @einola in https://github.com/nextsimdg/nextsimdg/issues/121#issuecomment-1222109945_

auraoupa:
I've added instructions on how to generate a changelog in docs/changelog/README.md and some rules about tags and releases in the general README.md

This will close #125 and #126


docs/changelog/README.md
```diff
@@ -0,0 +1,11 @@
+# Changelog generation
+
+A description of all the changes made to the nextsimdg code can automatically be generated anytime and is visible [here]()
```
einola, line 3: Missing link

einola (changes requested):
Two minor fixes are needed.

It's also not clear when the change log should be generated and when ``docs/changelog/CHANGELOG.md`` should be updated. Can you add a line on this in the ``README.md``?

Thanks :)

auraoupa:
I've added a few lines in the general README specifying when to generate the changelog file, but you can modify it of course if you think more frequent generation is needed.

einola (approved):

PR #181, opened 2022-10-20
Issue179 Abstract Dimensions

### LINKED ISSUE #179: Abstracting dimensions and sizes within ModelArray
Currently, different `ModelArray::Type`s can have different length arrays, even if they are logically similar within the model. For instance, should the horizontal dimensions of `H` and `Z` arrays not be equal?

Abstracting the dimensions to be their own thing also allows consistent naming in I/O operations and allows a field to know that the relevant dimensions are defined in the file. It also allows a modular writing of the dimension data to the file.

The system should also be compile-time flexible enough to permit implementation of the one dimensional Lagrangian arrays. That is, don't assume that `Type::H` is two dimensional and so on.

Make changes to the ModelArray size definitions to allow these above points.

timspainNERSC:
Allows more flexibility in the names and types of defined model arrays, but fixes the dimensionality of the arrays at compile time.
Closes #179


core/test/testmodelarraydetails/include/ModelArrayDetails.hpp
```diff
@@ -0,0 +1,31 @@
...
+        Y,
+        Z,
+        U,
+        COUNT
+    };
+
+    enum class Type {
+        ONED,
+        TWOD,
+        DOSD,
```
einola, line 22: Why ``DOSD``?

einola (approved):
Just one question on naming. Otherwise fine (so I don't want to bother with "request changes").

core/test/testmodelarraydetails/include/ModelArrayDetails.hpp
timspainNERSC: I needed a second 'two d'. So I went with the Spanish for two, as I didn't want to call the dimension Tod.

PR #190, opened 2022-11-14
IO classes for issue 151

timspainNERSC:
IO classes allowing communication between the dynamics and the rest of the model.

As part the integration between the dynamics code and the rest of the model, some communication between the `DGVector`, `CGVector` &c. of the dynamics and the `ModelArray` class in the rest of the model is necessary. This is implemented in the classes `DGModelArray` and `DGModelArray`, with corresponding tests in `DGModelArray_test.cpp` and `CGModelArray_test.cpp`.

The next goal is to allow the `ModelArray` class to carry DG components as well as mean/DG0 values. Using the dimension abstraction framework, new definitions of dimensions and components were included in the files in `src/core/discontinuousgalerkin`. These changes also define `ModelArray::Type`s for DG fields, the higher order DG fields used to hold the stress data and CG fields for velocities.

The `DGVector` definition required more grid information than was carried in the `DevGrid` class. The new `ParametricGrid` class is introduced to hold this information and it is intended to be the `IStructure` implementaion that gets used and added to for the ongoing development of the model (which DevGrid and RectGrid were never intended to be).

As with the other grid implementations, there is a separate class handling the NetCDF IO with the new grid structure, `ParaGridIO`. Not only does this include the new grid parameters, but also differentiates between restart files and diagnostic output files, with the latter now able to accumulate time series data. The behaviour of the IO and the grid class are somewhat tested in `ParaGrid_test.cpp`.

A header file with definitive names of fields was also introduced (`gridnames.hpp`).

Finally, to provide insight as to how all this might tie together in the final model, on of the dynamics examples was reworked to use the new code. This can be found in `NextsimMevp.cpp`. Connected to this is a small program to write a suitable initialization file, `createNextsimMeshData.cpp`. The output from this is require to initialize the data of `NextsimMevp.cpp`.

Also, I noticed I left a debugging function in the `TimePoint` class.


core/src/include/ModelArray.hpp
```diff
@@ -416,20 +424,50 @@ class ModelArray {
         return const_cast<double&>(std::as_const(*this)(args...));
     }
 
+    /*!
+     * @brief Sets the number of components for DG & CG array types.
+     *
+     * @param cMap a map from ModelArray::Type to the number of components.
+     */
+    static void setNComponents(std::map<Type, size_t> cMap);
```
a-smith-github, line 432: Note for me: This one in the cpp file

core/test/CMakeLists.txt
```diff
@@ -115,14 +126,26 @@ target_include_directories(testRectGrid PUBLIC "${SRC_DIR}" "${CoreModulesDir}"
 target_link_directories(testRectGrid PUBLIC "${netCDF_LIB_DIR}")
 target_link_libraries(testRectGrid LINK_PUBLIC Boost::program_options Catch2::Catch2 "${NSDG_NetCDF_Library}" Eigen3::Eigen)
 
-add_executable(testModelArray
-    "ModelArray_test.cpp"
-    "${CoreSrc}/ModelArray.cpp"
-    "${ModelArrayDetails}/ModelArrayDetails.cpp"
+add_executable(testParaGrid
+    "ParaGrid_test.cpp"
+    "${CoreModulesDir}/ParametricGrid.cpp"
```
a-smith-github, line 131: Issue: Perhaps this change is large enough and this can be done as a follow-up but these test dependencies could probably do with their own cmake variable.

core/test/CMakeLists.txt
```diff
@@ -115,14 +126,26 @@ target_include_directories(testRectGrid PUBLIC "${SRC_DIR}" "${CoreModulesDir}"
...
+    "${SRC_DIR}/MissingData.cpp"
+    "${SRC_DIR}/ModelArray.cpp"
+    "${SRC_DIR}/ModelMetadata.cpp"
+    "${SRC_DIR}/Time.cpp"
+    "${CoreSrc}/${ModelArrayStructure}/ModelArrayDetails.cpp"
     )
 
-target_include_directories(testModelArray PRIVATE "${CoreSrc}" "${ModelArrayDetails}")
-target_link_libraries(testModelArray PRIVATE Catch2::Catch2 Eigen3::Eigen)
+target_include_directories(testParaGrid PUBLIC "${SRC_DIR}" "${CoreModulesDir}" "${PhysicsDir}" "${PhysicsModulesDir}" "${netCDF_INCLUDE_DIR}" "${CoreSrc}/${ModelArrayStructure}")
```
a-smith-github, line 146: Issue: Same with these include directories

core/test/ParaGrid_test.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    REQUIRE(ticeRef(12, 14, 1) == tice(12, 14, 1));
+
+    ModelArray& coordRef = ms.data.at(coordsName);
+    REQUIRE(coordRef.nDimensions() == 2);
+    REQUIRE(coordRef.nComponents() == 2);
+    REQUIRE(coordRef.components({ 12, 13 })[0] - coordRef.components({ 11, 13 })[0] == scale);
+    REQUIRE(coordRef.components({ 12, 13 })[1] - coordRef.components({ 12, 12 })[1] == scale);
+
+    std::filesystem::remove(filename);
```
a-smith-github, line 181: Comment: Perhaps worth doing this teardown as part of catch2 destructor (using sections and writing this below all sections). In event of an error we're left with junk not deleted. If you've got time I'd do it here but could be worth assessing and re-jigging some tests as a separate issue. Might also remove some code duplication by sharing setup (although my inspection here hasn't been that thorough, my guess is it would be some work to get from here to a shared test setup in this file).

dynamics/applications/CMakeLists.txt
```diff
@@ -3,8 +3,51 @@ set(DynamicsSrc "../src")
 set(SRC_DIR "${DynamicsSrc}")
 set(INCLUDE_DIR "${DynamicsSrc}/include")
 
+set(CoreDir "../../core/src/")
+
 include_directories(${INCLUDE_DIR})
 
+add_executable(createNextsimMeshData createNextsimMeshData.cpp
+    "${CoreDir}/modules/ParametricGrid.cpp"
```
a-smith-github, line 11: Comment: As before with test dependencies. Might be worth including more than we need for simplicity (I know, I know but would be cool to just include core, physics, dynamics as needed --- I've been looking at this recently which is the only reason I use the word 'cool'). Obvs not every test will want the bulked out set but could save a lot of code & time & lines. 

We also currently make everything, including .o files, for each test in it's own sub directory but in principle could we share them if build time is a concern (currently the entirety of the testing time, less a second).

dynamics/tests/DGModelArray_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+    CHECK(source(targetPoint, 4) == 4 + my * (12 + mx *(14)));
+
+    ModelArray::setDimensions(ModelArray::Type::DG, { nx, ny });
+    ModelArray::setNComponents(ModelArray::Type::DG, DG);
+
+    ModelArray dest(ModelArray::Type::DG);
+    dest.resize();
+    DGModelArray::dg2ma(source, dest);
+
+    // Did it work?
```
a-smith-github, line 104: Comment: General point (not picking on this line specifically). We could perhaps do with expanded comments above each test case to define a test objective and talk about which files/classes/methods are being targeted. But comments above each require (like in the test below this) should be encouraged). Should improve debug speed, increase our bus number, make it easier to deprecate/enhance features, etc.

a-smith-github (commented):
Thanks Tim, 

Some cool stuff here (I'm not as grumpy as my comments make me sound). I've only looked at core/ so far but its COB and I thought I'd submit before I forget what's what. 

Some concerns over duplication/maintainability, some of which are easier to modify than others. Perhaps some test & cmake simplifications and/or introducing setup/teardown and comment format which could be made at some point. Not tried to run the program with a simple config but good to see tests passing.

If I had to pick, I think refactoring the if conditions in ParaGridIO.cpp is my 'must have', other issues could be a misunderstanding or a 'nice to have' or a follow-up let's reduce future engineering effort. 

Leaving as general comment but I think any recommendations should improve future maintainability (that's the goal anyway...). Drop issues as you feel appropriate. Happy to discuss anything - could be worthwhile anyway to bridge understanding. Appreciate there's been prioritisation, etc., where I've made comments on decisions outside of this change. 

Alex

dynamics/applications/createNextsimMeshData.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    grid.dumpModelState(state, metadata, filename, true);
+
+    return 0;
+}
+}
+
+int main()
+{
+    std::string filename = "nextsimMEVPstart.nc";
```
a-smith-github, line 92: Issue: Is there a future plan to optionally add a command line argument here? Can we comment/track issue?

a-smith-github (commented):
Thanks Tim, only a cursory look (I didn't inspect sources not listed in this pull) but this covers the dynamics part.

In addition to the comments, I didn't see the dynamics tests getting added to the github actions workflow. Do the tests run locally for you and could/should they be added?

Thanks,
Alex

CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ set(CodeComponents
 
 # The exact selection of dimensions and Types available to ModelArray
 if(True)
```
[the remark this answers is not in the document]
timspainNERSC: This conditional exists for some future time when the build configuration might switch between "finiteelement" and "discontinuousgalerkin". We can take it out if you would prefer, but I would rather leave it in, possibly with an explanatory comment.

core/src/ModelArray.cpp
```diff
@@ -179,22 +177,29 @@ void ModelArray::setData(double value)
 void ModelArray::setData(const double* pData)
```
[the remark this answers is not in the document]
timspainNERSC: The class member `m_data` is defined to use doubles, and it should be possible to feed a (correctly sized) double buffer to the array. This is tested in ModelArray_test.cpp line 71. It's possible that this could be refactored, but that should be a new issue.

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+    { "xvertexyvertexncoords", ModelArray::Type::VERTEX },
+};
+
+// Which dimensions are DG dimension, which could be legitimately missing
+const std::map<ModelArray::Dimension, bool> ParaGridIO::isDG = {
+    { ModelArray::Dimension::X, false }, { ModelArray::Dimension::Y, false },
+    { ModelArray::Dimension::Z, false }, { ModelArray::Dimension::XCG, true },
+    { ModelArray::Dimension::YCG, true }, { ModelArray::Dimension::DG, true },
+    { ModelArray::Dimension::DGSTRESS, true },
+    { ModelArray::Dimension::NCOORDS,
```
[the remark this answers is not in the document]
timspainNERSC: Clang formatting, sigh. That is how these definitions were laid out, until `clang-format` blindly applied its rules and messed it all up. I'll turn off clang-format and make it look nicer.

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+ModelState ParaGridIO::getModelState(const std::string& filePath)
+{
+    netCDF::NcFile ncFile(filePath, netCDF::NcFile::read);
+    netCDF::NcGroup metaGroup(ncFile.getGroup(IStructure::metadataNodeName()));
+    netCDF::NcGroup dataGroup(ncFile.getGroup(IStructure::dataNodeName()));
+
+    // Dimensions and DG components
+    std::multimap<std::string, netCDF::NcDim> dimMap = dataGroup.getDims();
+    for (auto entry : ModelArray::definedDimensions) {
+        if (dimCompMap.count(entry.first) > 0)
```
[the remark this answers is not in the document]
timspainNERSC: It's due to looking at the same piece of code at two separate times. Corrected.

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+        if (!dimensionKeys.count(dimKey)) {
+            throw std::out_of_range(
+                std::string("No ModelArray::Type corresponds to the dimensional key ") + dimKey);
+        }
+        ModelArray::Type newType = dimensionKeys.at(dimKey);
+        state.data[varName] = ModelArray(newType);
+        ModelArray& data = state.data.at(varName);
+        data.resize();
+
+        var.getVar(&data[0]);
```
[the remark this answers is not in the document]
timspainNERSC: What is 'this' in context?
The last error check was lines 95—98, checking that the read array has a corresponding compiled-in type. The only other error checking is whatever exceptions the NetCDF library throws (which are unhandled...).

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+    ncFile.close();
+    return state;
+}
+
```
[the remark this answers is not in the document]
timspainNERSC: At `ParaGridIO.hpp`:40—56.

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+            }
+            dimMap[type] = ncDims;
+        }
+        // Everything that has components needs that dimension, too
+        for (auto entry : dimCompMap) {
+            dimMap.at(entry.second).push_back(ncFromMAMap.at(entry.first));
+        }
+
+        std::set<std::string> restartFields
+            = { hiceName, ciceName, hsnowName, ticeName, maskName, coordsName }; // TODO and others
```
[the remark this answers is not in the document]
timspainNERSC: Issue #206

core/src/ParaGridIO.cpp
```diff
@@ -0,0 +1,325 @@
...
+            // The index arrays always start from zero, except in the first/time axis
+            std::map<ModelArray::Type, std::vector<size_t>> indexArrays;
+            std::map<ModelArray::Type, std::vector<size_t>> extentArrays;
+            for (auto entry : ModelArray::typeDimensions) {
+                ModelArray::Type type = entry.first;
+                std::vector<netCDF::NcDim> ncDims;
+                // VERTEX arrays do not need time axes.
+                std::vector<size_t> indexArray;
+                std::vector<size_t> extentArray;
+                if (type != ModelArray::Type::VERTEX) {
```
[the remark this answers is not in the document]
timspainNERSC: Yes, it does.

core/src/discontinuousgalerkin/ModelArrayDetails.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+// A source file to detail the ModelArray dimensions and types for
+// Discontinuous Galerkin models, as well as the relationships between them.
+
+// Should be grouped with a consistent ModelArrayDetails.cpp and
+// ModelArrayTypedefs.hpp
+
+namespace Nextsim {
+std::map<ModelArray::Dimension, ModelArray::DimensionSpec> ModelArray::definedDimensions = {
+    { ModelArray::Dimension::X, { "x", 0 } }, { ModelArray::Dimension::Y, { "y", 0 } },
```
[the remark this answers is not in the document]
timspainNERSC: Again, clang-format seems to disagree….

core/test/CMakeLists.txt
timspainNERSC: A CMake tidy-up would be a good idea after we have the dynamics and physics tied into one model. Since this isn't currently a problem, I suggest making it a future issue.

timspainNERSC: Refactor, that's the software engineery word.

core/test/ParaGrid_test.cpp
timspainNERSC: Yes, I need to review Catch2 setup and teardown, and actually use those facilities.

dynamics/applications/CMakeLists.txt
timspainNERSC: My general attitude with CMake, like with catch2, is to get something to work, and refactor once that is happening. Also, my CMake experience is currently a bit thin, so it would require a bit of learn & play time for me to do anything much more advanced.

dynamics/applications/NextsimMevp.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+    Nextsim::ParametricGrid gridIn;
+    Nextsim::ParaGridIO* readIO = new Nextsim::ParaGridIO(gridIn);
+    gridIn.setIO(readIO);
+
+    Module::Module<Nextsim::IStructure>::setImplementation("ParametricGrid");
+
+    Nextsim::ModelState state = gridIn.getModelState("nextsimMEVPstart.nc");
+    size_t NX = Nextsim::ModelArray::definedDimensions.at(Nextsim::ModelArray::Dimension::X).length;
+    // Let's ASSUME it's square
```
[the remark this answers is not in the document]
timspainNERSC: Agreed. Witness thought processes fossilized in action.

dynamics/applications/NextsimMevp.cpp
```diff
@@ -0,0 +1,357 @@
...
+    //! Define the spatial mesh
+    Nextsim::ParametricMesh smeshOG;
+    smeshOG.readmesh(meshfile);
+
+    // Assert some facts about the vertex positions
+    assert(smesh.nx == NX);
+    assert(smesh.ny == NX);
+    assert(smesh.nelements == NX * NX);
+    assert(smesh.nnodes == (NX + 1) * (NX + 1));
+    assert(smesh.vertices(0, 0) == smesh.vertices(1, 0));
```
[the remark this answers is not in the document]
timspainNERSC: Done

dynamics/applications/NextsimMevp.cpp
```diff
@@ -0,0 +1,357 @@
...
+    //! LOG message
+    constexpr double T_log = 10.0 * 60.0; // every 30 minute
+    constexpr size_t NT_log = T_log / dt_adv + 1.e-4;
+
+    // netCDF output
+    std::string diagFile = "mevp.nc";
+    size_t ncOutputPeriod = 30; // Every hour
+
+    ////////////////////////////////////////////////// Forcing
+    Nextsim::Interpolations::Function2CG(smesh, momentum.GetOceanx(), OceanX());
```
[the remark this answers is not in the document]
timspainNERSC: Aside from hiceName, these is all taken directly from the copy/paste source. I'm disinclined to make arbitrary changes solely in this file.
And it might help to think of hice as a word in and of itself. Like the plural of house!

dynamics/applications/createNextsimMeshData.cpp
timspainNERSC: No, the data files created by this program will only ever be used by `NextsimMevp.cpp`, which is itself best thought of as a prototype bridging from the `benchmark_box_sasip_mevp.cpp` application towards the likely usage in the full model. The data in the created files will be present in a future version of the restart file, which will have its own creation script (likely python), which will of course include arguments and/or configuration for the output file name.

dynamics/tests/DGModelArray_test.cpp
timspainNERSC: (Also, I'm being very careful when crossing the road.)

CMakeLists.txt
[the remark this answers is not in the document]
a-smith-github: No, just would be good to note in the code or capture as an issue (or both --- put issue number in comment)

core/src/ModelArray.cpp
```diff
@@ -179,22 +177,29 @@ void ModelArray::setData(double value)
...
 {
     resize();
-    auto out = std::copy(pData, pData + m_sz.at(type), m_data.data());
+    auto out = std::copy(pData, pData + m_sz.at(type) * nComponents(), m_data.data());
 }
 
-void ModelArray::setData(const DataType& from) { setData(from.data()); }
+void ModelArray::setData(const DataType& from) { m_data = from; } // setData(from.data()); }
```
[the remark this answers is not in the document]
a-smith-github: No, apologies, I think I thought this was a header file when writing this one --- the actual header has a lot of complexity and I've not quite got to grips with when the header files do and don't have implementations.

core/test/ParaGrid_test.cpp
a-smith-github: From what I can tell it's a pain and you don't get much for free. I've tried in #207 to do setup and teardown in test_section and have the tests running inside sections each time doing a clean setup and tear down (and looping over sections through the dynamic section, but ignore that for the purpose of this comment). Can't seem to get much better than that though. If you find something that takes a snapshot of the environment and restores it automatically, let me know..... (otherwise maybe we might want to make one --- and undo what I did in 207).

core/src/discontinuousgalerkin/ModelArrayDetails.cpp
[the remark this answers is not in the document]
a-smith-github: As discussed we can defer to clang or modify our .clang-format, just noting that it seems odd. Can't find it here but specially the line with the comment on the end in ParaGrid(?) such that only one { class::method::thing, {thing, thing} } fit on a line when the above and below had two means clang probably wants a talking to.

CMakeLists.txt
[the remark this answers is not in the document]
timspainNERSC: Done

core/test/CMakeLists.txt
timspainNERSC: I'm not sure a single comment would work when the whole file needs a good hard refactoring. Issue #210 in any case.

core/test/ParaGrid_test.cpp
timspainNERSC: Realistically, how much of an issue is this? If the tests fail in the test environment, then the file disappears when the environment does. On a dev's machine the file will finally be deleted when they fix the test and get it running to completion. It's an incentive!

a-smith-github: Depends on the user and how big the project gets but this is really about future-proofing. For us, in CI or dev, the main impact it could have is on subsequent tests --- right now, I don't think we have many tests which would be impacted by this but if we need to validate the existence and content of output files then we start running the risk that we are actually assessing the output of previous tests rather than the test we are running. If two tests produce files and they were to share a name (or if we wanted to swap this test to dynamic sections or just split into sections) which may not be an issue here but as the project gets bigger we lose control over that / have to begin to check that for each test we write. Similarly if part 1 of a test succeeds and part 2 fails and we make changes to source to try and fix it, we could find that we've stopped producing the file or create it in a different location by accident but be under the impression our change has had no effect, while a new failure in part 1 would make it very very obvious we have broken something. In CI this is less of a big deal but obvs our dev environment persists.

For the end-user, they'll probably want to run tests to validate on their machine before doing any simulations. In such a case, it would be easier not to clutter folders with test artefacts, particularly if end-users can add functionality (and tests). If someone (including us) were to rewrite the github action workflow and run tests from the top level directory (which is attractive to simplify the file, but not necessarily what we'd want to do), this would become more critical. In general though, this test can be run from any directory and that means a user could and somewhere probably would end up doing that. Additionally in any of these scenarios, if the test later depended on another file (perhaps just to compare numerical results), it could fail as the path to file might be wrong and we'd have left the file behind.

And for any future development, the requirements/limitations of any feature are really captured (even enforced) by tests. If you wanted to see how a new feature might impact other features, you'd probably want to make a dummy change and run some tests and build a list of components/methods impacted which you'd need to modify. This would definitely be the case if we were removing support for anything - you would intentionally be triggering failures and don't want any accidental passes as a result of that. You want to be able to 100% trust that the test results reflect interdependencies in the code and that the test ecosystem has no impact. 

Dream scenario is to make a temporary dir at the start of each test, copy inputs/configs (if applicable) into that folder and run the tests there. On success or failure, we'd then remove the dir and contents and return to original directory. Obvs this is a task for later because it's not a necessary thing but it makes life easier as the project evolves (and, like all things, is easier to do earlier than later)

Before doing anything it'd be worth checking if Catch2 already does this. I couldn't see it but it would make the most sense if there was an environment setup macro (or something) it ran before re-doing the setup (top of TEST_CASE) before running the next SECTION (and then a cleanup macro after the end of the test case after exiting the section)

a-smith-github (approved):
Approving merge, some outstanding notes on both actioned and dropped issues would be helpful but not expecting further commits on this PR. Can confirm all were discussed and outcomes agreed/understood, possible all actioned were committed. Passes testing.

PR #200, opened 2022-11-25
Issue199 Seconds Duration

### LINKED ISSUE #199: Nextsim::Duration should probably have a ctor from seconds
It's easily convertible to seconds. The other direction should be easy as well.

timspainNERSC:
Close #199

### DIFF
--- a/core/src/include/Time.hpp
+++ b/core/src/include/Time.hpp
@@ -39,8 +39,7 @@ public:
         : m_d()
     {
     }
-    Duration(const std::string& str) { this->parse(str); }
-
+    Duration(const std::string& str);
     Duration(double seconds);
 
     TimePoint operator+(const TimePoint& t) const;
--- a/core/src/include/Time.hpp
+++ b/core/src/include/Time.hpp
@@ -41,6 +41,8 @@ public:
     }
     Duration(const std::string& str) { this->parse(str); }
 
+    Duration(double seconds);
+
     TimePoint operator+(const TimePoint& t) const;
 
     Duration& operator+=(const Duration& a)
--- a/core/src/include/Time.hpp
+++ b/core/src/include/Time.hpp
@@ -98,6 +98,7 @@ private:
     {
     }
     Basis m_d;
+    void setDurationSeconds(double);
 };
 
 /*!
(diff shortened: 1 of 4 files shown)


core/test/Time_test.cpp
```diff
@@ -202,6 +202,27 @@ TEST_CASE("Durations", "[Duration]")
 
     Duration maybeADay = tt_day - tt;
     REQUIRE(maybeADay.seconds() == 1 * days);
+
+    Duration tenSeconds(10.);
+    REQUIRE(tenSeconds.seconds() == 10);
+
+    Duration hundredSeconds(100.);
```
a-smith-github, line 209: Issue (optional): We probably don't get much extra with some of these new tests --- unless you're aware of some behaviour that occurs at these milestones (please add comments above each one if so). Maybe using [0, 10, 1e4] is sufficient to test the new constructor, even [0, 10]. Could add implicit casting (*if* we *need* to lock that down) otherwise fine

core/test/Time_test.cpp
```diff
@@ -202,6 +202,27 @@ TEST_CASE("Durations", "[Duration]")
...
+    REQUIRE(hundredSeconds.seconds() == 100);
+
+    Duration thousandSeconds(1000.);
```
a-smith-github, line 212: Comment: testTimeClasses doesn't fit the naming convention for text executables. I see why the choice was made (to avoid that one letter difference) but.....

a-smith-github (changes requested):
It's fine (comment and drop as needed) but some comments

core/src/include/Time.hpp
```diff
@@ -41,6 +41,8 @@ class Duration {
     }
     Duration(const std::string& str) { this->parse(str); }
```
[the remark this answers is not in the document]
timspainNERSC: Moved to the cpp file.

core/test/Time_test.cpp
timspainNERSC: I have changed these to test parsing different forms of the string representation of doubles. I think their existence is now justified.

core/test/Time_test.cpp
timspainNERSC: Convention is a strong word, but I belatedly agree.

a-smith-github (approved):
Thanks Tim

PR #204, opened 2022-12-08
Freedrift tests

einola:
Merge the free drift tests and modifications to the solver into develop.

Can't merge automatically, but @pminak has to fix it :)


pminak (approved):
This merge introduces the second scheme by Hunke and Dukowicz. The main difference lies in the treatment of Coriolis term. Now we can choose  Coriolis force treated explicitly with MEBStep and implicitly with MEBStepHD (like in the paper by  Hunke and Dukowicz).

Issue #207, opened 2022-12-12
Custom output directory


a-smith-github:
Summary: Enable users to (optionally) add a line to the configuration file to change output directory

Requirements:
- Must be an optional argument
- Must move write location of the log files, restart file, diagnostic files

Test Requirements:
- Create tests with config files that contain and do not contain the output_dir line
- Test files are created in location that matches the stored output_dir
- Test that the output_dir is read and stored correctly

PR #208, opened 2022-12-12
Issue207 --- (Optional) Config Key: Custom Output Directory

a-smith-github:
Summary:
Adds an optional argument to the configuration file 'output_dir' which can be used to specify an output directory for files generated by nextsim-dg. Users may specify the custom output by adding "output_dir = <folder>" to any cfg file passed to nextsim using --config-file(s). <folder> does not need to contain a file separator. The line does not need to be included in any cfg for nextsim-dg to run (defaults to cwd --- previous behaviour). 

---

Notes for reviewers: 

- Targets issue #207 (follow link for requirements)
- Creates new test with dependencies on core/, dynamics/, physics/, so I've placed a new test folder in root directory and modified the cmake structure to take this into account. This folder could be used for wider integration tests which require a model.run() to fully test functionality. Created a dependency list based upon nextsim-dg's full list less main.cpp which may (or may not) be a suitable list for integration/system tests. 
- Let me know if there are breaches of the design paradigm or if you would prefer information to be shared beyond Model by mechanisms other than by adding an argument to methods ---- would you prefer more setting private member variables e.g. class.setOutputDir rather than modifying method where usage takes place method(std::filesystem::path& outdir)? A blend has been used here and this may not be preferred.

For anyone:
- Let me know if you spot a mistake or can see opportunity for better naming / style

Functional:
- User adds text input line to cfg file using keyword 'output_dir' e.g. "output_dir = output/"
- System file separator is optional

Architectural / Implementation:
- Modify model config map so configurator can capture the additional line in cfg files and store
- During model configuration store output in model
- Pass std::filesystem::path to relevant classes (e.g. DevStep) and store in related classes as member variables or prepend to filename before passing to output methods (e.g. StructureFactory)
- Use std::filesystem operator '/' to handle combining dir + filename using system preferred path

Integration/test:
- No downstream dependencies
- Upstream dependencies: Configurator
- Modifies/Impacts: Model, DevStep
- Tested on a system level using model.run and example config files

---

Testing performed:

Local run of github action workflow for ubuntu on ubuntu (after change to workflow file) -> all pass
Manual 'clang-format --dry-run -Werror' on each dir containing changes -> empty output


einola:
It's not really what we're after here.

PR #213, opened 2022-12-21
Issue156 bbm crash

pminak:
We resolve issues with a BBM crash. This commit is a base for the results in the modelling paper. 
We separate the two models by implementing BBM.hpp from MEB.hpp.
Two time-stepping schemes are included MEBStep (fully explicit) and BBMStep (Hunke-Dukowicz).


winzerle (approved):
all fine

PR #214, opened 2023-01-16
Issue 209 dummy dynamics

timspainNERSC:
The easiest other type of dynamics to implement is no dynamics. This PR introduces a module for the dynamics and provides a dummy implementation of the dynamics class.


einola:
Will be merged with issue #220

PR #225, opened 2023-01-25
Calculate the ice-ocean heat flux.

### LINKED ISSUE #224: None of the ocean boundary implementations calculate Qio
And they really should!

timspainNERSC:
This will close #224 on the develop branch.

### DIFF
--- a/physics/src/modules/ConstantOceanBoundary.cpp
+++ b/physics/src/modules/ConstantOceanBoundary.cpp
@@ -6,7 +6,8 @@
  */
 
 #include "include/ConstantOceanBoundary.hpp"
-
+#include "include/IIceOceanHeatFlux.hpp"
+#include "include/Module.hpp"
 #include "include/constants.hpp"
 
 namespace Nextsim {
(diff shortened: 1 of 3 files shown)


a-smith-github (changes requested):
I'm going to need more info here and/or with issue and if this is really a bug I'd like to see a test added

What was the symptom (e.g. the error message / silent wrong answer) and diagnosis? And if really a bug there should be a test in this PR to ensure that we don't introduce a regression in future changes. This is particularly important in the case of silent wrong answers.

timspainNERSC:
The problem was that the Qio field was zero for the fixed value ice ocean boundary (the configured ocean boundary was in fact doing the calculation). And it should not be for any of the implementations of the `IIceOceanHeatFlux`, since that is one of the main functions of the module. This was prompted by me having missed adding it to the TOPAZ ocean boundary on the relevant branch, and checking whether I had in fact included the calculation in the implementations already present in the develop branch.

In any case, I have added a test.

a-smith-github (approved):

PR #229, opened 2023-01-25
Obtain SST before it is needed by the atmosphere flux calculation.

### LINKED ISSUE #227: The atmosphere boundary depends on the ocean boundary
The calculation of the fluxes in the atmosphere boundary implementations depends on the sea surface temperature to calculate  the latent heat flux over open water. Therefore the call to IOceanBoundary::updateBefore() should come before the call to IAtmosphereBoundary::update() in PrognosticData::update().

The clue is in the name, really….

timspainNERSC:
Will close #227 on develop.


a-smith-github (changes requested):
Probably fine but as with the other bug, can we add a test to lock down the desired behaviour or indicate which test (that was previously failing) now passes due to this bug fix?

timspainNERSC:
It took some doing, but I managed to write a test that failed on the old code and passes after this change is made.

core/test/PrognosticData_test.cpp
```diff
@@ -1,12 +1,105 @@
...
+    config << "snow = 0" << std::endl;
+    config << "rainfall = 0" << std::endl;
+    config << "wind_speed = 5" << std::endl;
+
+    std::unique_ptr<std::istream> pcstream(new std::stringstream(config.str()));
+    Configurator::addStream(std::move(pcstream));
+
+    ConfiguredModule::parseConfigurator();
+
+    class OceanData : public IOceanBoundary {
```
a-smith-github, line 50: Hi Tim. Noticed this declaration exists in this form in multiple locations including ConfiguredAtmosphere_test, FiniteElementFluxes_test and BasicIceOceanFlux_test (sometimes multiple times in one file). Could we either make a class in the source (presuming we will need this) or make a common class in the physics/test directory that we can import? I presume the duplication is because this is implemented in a change that is being worked on on another branch but we know is coming soon?

Let's keep maintenance engineering effort as low as possible

a-smith-github (changes requested):
Thanks Tim. Test looks legit thanks for that, just one thing... Either here or as a follow-up issue but only if that follow-up is in the not too distant future (pretty much top of the pile).

Alex

core/test/PrognosticData_test.cpp
timspainNERSC: No, the duplication is because it is supposed to be an ad-hoc class that implements a given `ModelComponent` (here `IOceanBoundary`) and provides whatever data is required by the particular test. And because it is ad-hoc it gets shoved in wherever I thought it was needed.

I would make the argument that breaking the (admittedly) similar code out into a separate class would not necessarily reduce maintenance overhead as all the classes would have to be kept reconciled. Could we at least create a new issue to discuss it. Since it impacts several tests, it should be its own issue, spec and pull request, in my opinion.

a-smith-github: Yes we can cover this as a follow-up/tidy-up issue. Please create and tag and link here and I'll approve this PR. Nudge me if needed.

timspainNERSC: A relevant issue has been created at #256

a-smith-github (approved):

PR #241, opened 2023-02-09
Issue240 debug MAR

### LINKED ISSUE #240: Catch missing ModelArrayRefs during test development
When creating tests, sometime real data arrays are not connected to ModelArrayRefs that the classes under test require. Having a way to debug this problem without adding ModelArray after ModelArray until the SIGSEGVs go away would be helpful.

Two possible solutions occur:

1. Add code that checks for accesses within ModelArrayRef, but is only compiled under certain conditions (using a preprocessor `#ifdef` or for all Debug builds)
2. Create a copy of the ModelArrayRef class that includes the checking code, but is only ever used by developers trying to work out the above problem and replaced by the current version once the test has all the data it needs.

I will implement number 2 until someone tells me otherwise.

timspainNERSC:
Implements and tests a version of ModelArrayRef that throws if an array is not registered.

Close #240


timspainNERSC:
Easier to re-implement than I thought it was going to be!

timspainNERSC:
Oh so very outdated.

PR #257, opened 2023-03-27
Issue194 topaz era into dynamics

timspainNERSC:


timspainNERSC:
Merged in my local & pushed

PR #203, opened 2022-12-01
Issue194 ERA5 and TOPAZ forcings

timspainNERSC:
In order to run models that are reasonably representative of the real world, real-world forcings need to be used.
For a certain grid (Arctic only, fixed resolution) implement forcing ingestion functions that work only with forcing data that has been preprocessed to match the spatial extent and resolution of the model. Time interpolation is still required.

To test the functionality, one of the added changes is a year-long run of the model using forcings from ERA5 and TOPAZ. Running this is as simple as running the script `run/run_topaz128x128.sh` – if you have the necessary forcings files. Since these are ~7 GB, they are not included in the repository and need to be generated. The generation script is `run/era5_topaz4_maker.py`. This requires the ERA5 and TOPAZ4 data files, which can be found on the SASIP data repository hosted at `univ-grenoble-alpes.fr`. The necessary files are:

- `ERA5_*_y2010.nc` in `/summer/sasip/model-forcings/atmo_forcing/ERA5_Arctic`
- `TP4DAILY_2010*.nc` in `/summer/sasip/model-forcings/ocean_forcing/TOPAZ4_Arctic`

The generation script is then run in a directory containing these files with start and stop time parameters. Since the forcing files should cover all of 2010, the full command is
```
./era5_topaz4_maker.py --file init_topaz128x128.nc --start 2010-01-01 --stop 2011-01-01
```

This will generate the files `ERA5_2010-01-01_2011-01-01.nc` and `TOPAZ4_2010-01-01_2011-01-01.nc`. These file, along with the restart file `init_topaz128x128.nc`, which is included in the repository, are all that is necessary to run the example model.

To determine if the model functioned correctly, open the output file `diagnostic.nc`. This will contain output data from the model sampled once per day. The data in the field `hice_true` should expand during the first part of the year, as the thermodynamics causes the sea ice to thicken. Then over the summer, the sea ice should retreat before reforming as the end of the year approaches.

Closes #194

PR #269, opened 2023-04-17
Issue220 dynamics module

einola:
The bulk of the dynamics merge


dorchard:
This PR merge depends on first solving https://github.com/nextsimdg/nextsimdg/issues/275

timspainNERSC:
Cleaning up comment out code 1/n.

With all its executive statements, `dynamics/applications/CMakeLists.txt` does nothing. Since all the source code in `applications` is for creating executables, the entire subdirectory is redundant.

So I deleted it and removed the reference to it from the `dynamics` CMakeLists.txt.

einola:
This pends on #203 and conflict resolution

timspainNERSC:
Conflicts keep popping up like bubbles under wallpaper!

einola:
I'm not a reviewer, but I say it's time to merge.

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -0,0 +1,334 @@
...
+
+    /*
+        Temrorary Ocen and Wind Field
+    */
+
+    class OceanX : public Nextsim::Interpolations::Function {
+    public:
+        double operator()(double x, double y) const
+        {
+            return  0.01 * (2.0 * y / ( 121. * 25000.) - 1.0);
```
dorchard, line 100: There are some magic numbers here which are probably better extracted as constants, particular 121/154 and 25000 (here and line 107, 121,122,142,143)

run/era5_topaz4_maker.py
```diff
@@ -65,37 +65,47 @@ def topaz4_source_file_name(field, unix_time):
...
 
+    iwrap = (i + 1) % data.shape[1]
+
     dataplier = data != missing # False is zero
 
     weighted_sum = ((1 - fj) * (1 - fi) * data[j, i] * dataplier[j, i] +
-        (1 - fj) * (fi) * data[j, i + 1] * dataplier[j, i] +
-        (fj) * (1 - fi) * data[j + 1, i] * dataplier[j, i] +
-        (fj) * (fi) * data[j + 1, i + 1] * dataplier[j, i])
+        (1 - fj) * (fi) * data[j, iwrap] * dataplier[j, iwrap] +
```
dorchard, line 101: Are we sure the second `iwrap` here is right? The code it replaces was `i` but this is now `i + 1` modded.

run/era5_topaz4_maker.py
```diff
@@ -65,37 +65,47 @@ def topaz4_source_file_name(field, unix_time):
...
+        (fj) * (1 - fi) * data[j + 1, i] * dataplier[j + 1, i] +
```
dorchard, line 102: Is the second `j+1` right here? The previous code was `j`.

run/era5_topaz4_maker.py
```diff
@@ -65,37 +65,47 @@ def topaz4_source_file_name(field, unix_time):
...
+        (fj) * (fi) * data[j + 1, iwrap] * dataplier[j + 1, iwrap])
```
dorchard, line 103: Is the second `iwrap` correct here? (not `i`? as in the previous code)?

run/era5_topaz4_maker.py
dorchard: Or do we need ` i % data.shape[1]`?

dorchard (commented):
I've done a light touch look through things and spotted a few potential problems. They could be nothing though, but thought best to flag.

core/src/modules/include/Dynamics.hpp
```diff
@@ -0,0 +1,33 @@
...
+class Dynamics : public IDynamics {
+public:
+    Dynamics();
+
+    std::string getName() const override { return "Dynamics"; }
+    void update(const TimestepTime& tst) override;
+
+    void setData(const ModelState::DataMap&) override;
+private:
+    // TODO: How to get the template parameters here?
```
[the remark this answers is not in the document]
timspainNERSC: #360

run/era5_topaz4_maker.py
timspainNERSC: This is now correct. The indices of `data` and `dataplier` should match. That `dataplier` had [j, i] for all the terms was a copy-paste error. The terms for basic bilinear interpolation should be [j, i], [j, i+1], [j+1, i] and [j+1, i+1]. Here i+1 is replaced by iwrap to account for the wrap-around in longitude.

run/era5_topaz4_maker.py
timspainNERSC: See comment on line 101.

run/era5_topaz4_maker.py
timspainNERSC: See comment on line 101.

dynamics/src/DGTransport.cpp
```diff
@@ -183,11 +182,31 @@ void DGTransport<DG>::reinitnormalvelocity()
...
+    // the normal is set only once, from the inside. These edges must be scaled with 2.0
+    
+    for (size_t seg = 0 ; seg<4; ++seg) // run over the 4 segments (bot, right, top, left)
+      {
+#pragma omp parallel for
+	for (size_t i = 0; i<smesh.dirichlet[seg].size();++i)
+	  {
+	    const size_t eid = smesh.dirichlet[seg][i];  //! The i of the boundary element 
+	    const size_t ix = eid % smesh.nx;            //! x & y indices of the element
+           const size_t iy = eid / smesh.nx;
```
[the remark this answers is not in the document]
timspainNERSC: No, using nx in both lines is correct. This is the usual way of getting 2-dimensional indices from a 1-dimensional index.

dynamics/src/include/DynamicsKernel.hpp
timspainNERSC: Very much so!
However all of this code is removed [here](https://github.com/nextsimdg/nextsimdg/pull/327/files#diff-906d264d24ee76696155c8cabfaeccccf0fbe08736c467b65e7ebb705f534015) as part of the development that gets the main model and the dynamics talking to each other correctly. After merging this PR, the dynamics still has no effect on the model, so these awful hard coded constants can be ignored for now.

dynamics/src/DGTransport.cpp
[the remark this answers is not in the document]
dorchard: Thanks

dynamics/src/include/DynamicsKernel.hpp
dorchard: Gotcha.

dorchard (approved):

PR #270, opened 2023-04-19
Issue267 dummythermo

### LINKED ISSUE #267: Dummy thermodynamics
As part of the dynamics integration and testing, we might want to run dynamics-only models, where the thermodynamics do nothing to the ongoing state of the ice.

Modules should be available so that the thermodynamics step does nothing to the state of the model, except for any ancillary data fields that might be necessary.

timspainNERSC:
Add thermodynamics modules that do nothing.

This PR closes #267 

> As part of the dynamics integration and testing, we might want to run dynamics-only models, where the thermodynamics do nothing to the ongoing state of the ice.
> 
> Modules should be available so that the thermodynamics step does nothing to the state of the model, except for any ancillary data fields that might be necessary.


timspainNERSC:
> As part of the dynamics integration and testing, we might want to run dynamics-only models, where the thermodynamics do nothing to the ongoing state of the ice.
> 
> Modules should be available so that the thermodynamics step does nothing to the state of the model, except for any ancillary data fields that might be necessary.

einola (approved):

einola:
It looks fine ... except for the conflict, of course.

PR #284, opened 2023-04-26
Issuexxx thermoice0 test

timspainNERSC:
Add the ThermoIce0 test from the TOPAZ ERA branch. Merge after #278 and #282


timspainNERSC:
Contains all changes from #278 and #282 

Introduces a test using the `ThermoIce0` thermodynamics implementation to test that the ice minima are being respected.

physics/src/modules/ThermoIce0.cpp
```diff
@@ -133,7 +131,7 @@ void ThermoIce0::calculateElement(size_t i, const TimestepTime& tst)
     }
 
     // Melt all ice if it is below minimum threshold
-    if (hice[i] < MinimumIce::thickness()) {
+    if (0. < hice[i] && hice[i] < IceMinima::h()) {
```
a-smith-github, line 134:
```suggestion
    if (hice[i] > 0 && hice[i] < IceMinima::h()) {
```

physics/src/modules/ThermoIce0.cpp
```diff
@@ -150,9 +148,10 @@ void ThermoIce0::calculateElement(size_t i, const TimestepTime& tst)
         qio[i] += hice[i] * bulkLHFusionIce / tst.step + hsnow[i] * bulkLHFusionSnow / tst.step;
 
         // No ice, no snow and the surface temperature is the melting point of ice
+        cice[i] = 0.;
         hice[i] = 0.;
         hsnow[i] = 0.;
-        tice.zIndexAndLayer(i, 0) = Ice::Tm;
+        tice.zIndexAndLayer(i, 0) = 0.;
```
a-smith-github, line 154: I don't understand the significance of cice[i] = 0 and changing Ice::Tm ---> 0. To me L154 looks like an error given the comment on L150. Please satisfy yourself that this is correct / update or expand on the comment. 

Would also help if 'c' and 'h' and 't' were converted to useful prefixes and/or we used camel case as is required by the webkit style / nextsim-dg coding conventions

physics/src/IceGrowth.cpp
```diff
@@ -181,24 +201,25 @@ void IceGrowth::lateralIceSpread(size_t i, const TimestepTime& tstep)
...
-            updateThickness(hsnow[i], cice[i], deltaC, 0);
+            updateThickness(hsnow[i], cice[i], deltaCIce[i], 0);
         }
     }
 }
 
 void IceGrowth::applyLimits(size_t i, const TimestepTime& tstep)
 {
-    if (cice[i] < MinimumIce::concentration() || hice[i] < MinimumIce::thickness()) {
+    if ((0. < cice[i] && cice[i] < IceMinima::cMin)
```
a-smith-github, line 221: Let's keep one format and saying cice is better than zero (a.k.a. it exists) is simpler.

Obvs this logic is slightly different so re-jig as needed. 

```suggestion
    if ((cice[i] > 0 && cice[i] < IceMinima::cMin)
```

physics/src/IceGrowth.cpp
```diff
@@ -181,24 +201,25 @@ void IceGrowth::lateralIceSpread(size_t i, const TimestepTime& tstep)
...
+        || (0. < hice[i] && hice[i] < IceMinima::hMin)) {
```
a-smith-github, line 222:
```suggestion
        || (hice[i] > 0 && hice[i] < IceMinima::hMin)) {
```

physics/test/ThermoIce0_test.cpp
a-smith-github, line 1: I don't know what we're testing or why we need two custom classes to do so. If this is a common workflow then let's extract the classes out of this test as that is much more maintainable both within this test and outside it. If providing a full implementation is problematic without running the full code then now is the time to develop system tests that run using config files and a test version of the nextsim main().

Let's provide a comment at the top of the file/test describing the test objective/what part of the code is being covered or what special conditions/edge cases this test covers.

physics/test/ThermoIce0_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    TimestepTime tst = { TimePoint("2000-01-01T00:00:00"), Duration(600) };
+    ThermoIce0 ti0t;
+    ti0t.configure();
+    ti0t.setData(ModelState::DataMap());
+    ti0t.update(tst);
+
+    ModelArrayRef<ModelComponent::SharedArray::H_ICE, MARBackingStore> hice(ModelComponent::getSharedArray());
+    // So little ice should be reduced to zero
+    REQUIRE(hice[0] == 0.);
```
a-smith-github, line 136: e.g. my guess is this test is trying to establish that the height(?) of the ice is being reset or being constantly pinned to zero or matches a golden value (it all melts) - however it is impossible to be sure without a comment

I have no way of knowing if this is likely to fail sporadically (due to randomness) without inspecting the rest of the code. This kind of information would be useful in a comment at the top of the test

physics/test/ThermoIce0_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+    ModelArrayRef<ModelComponent::SharedArray::C_ICE, MARBackingStore> cice(ModelComponent::getSharedArray());
+    REQUIRE(cice[0] == 0.);
```
a-smith-github, line 138: ditto

physics/test/ThermoIce0_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+        HField mlbhc; // Mixed layer bulk heat capacity
+        HField tice0;
+        HField qio;
+
+        ModelState getState() const override { return ModelState(); }
+        ModelState getState(const OutputLevel&) const override { return getState(); }
+    } atmoState;
+    atmoState.setData(ModelState::DataMap());
+
+    // Supply the atmospheric boundary arrays, without an entire
```
a-smith-github, line 82: Why? Use this comment to explain --- or move it above the relevant class. A similar comment would be appropriate for the flux class

physics/test/ThermoIce0_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+        HField hice;
+        HField hice0;
+        HField cice;
+        HField hsnow;
+        HField sst;
+        HField sss;
+        HField tf;
+        HField snow;
+        HField mlbhc; // Mixed layer bulk heat capacity
```
a-smith-github, line 73: 1. This is a copy and paste
2. The naming convention clearly isn't sufficient, please re-name

a-smith-github (changes requested):
I've not looked at files that exist in the other PRs (see PR description for the relevant numbers for slab ocean and ice minima) except where necessary. However there are some suggestions which were not made in those PRs which would be relevant. 

Not using good naming conventions is coming back to bite. We should also be using camel case as required in the webkit style.

Test need comments before it can fully reviewed and the classes should be extracted. If that is an issue we should reconsider how we go about running tests - imo it's getting to the point where it's becoming difficult to avoid not creating a system test running off of example config files. 

Possible we need more tests but it is unclear at this time

physics/src/modules/ThermoIce0.cpp
timspainNERSC: Actually, this makes sense. I have changed it the value of Ice::Tm in Celsius. Which is still zero, but is now more explicit in the code.

physics/test/ThermoIce0_test.cpp
timspainNERSC: The physical data (initial ice thickness, concentration, temperature, air temperature &c.) need to be provided from somewhere. Deriving a custom subclass of ModelComponent allows for using the functions and values defined in that class. The values could be defined directly in the function body of the test, but that would result in almost the same code with many `ModelComponent::` prefixes to functions and values.

The second custom class is an implementation of `IFluxCalculation`. The implementation of this interface is called by the class under test, but we certainly don't want the full flux calculation in this test. So the implementation 'calculates' zero values for all of the returned fluxes.

Added comments to this effect.

physics/test/ThermoIce0_test.cpp
timspainNERSC: Resolved by adding an explanatory comment.

physics/test/ThermoIce0_test.cpp
timspainNERSC: Added an explanatory comment.

PR #287, opened 2023-04-26
Correct type of componentMap.

### LINKED ISSUE #286: ModelArrayDetails map should be a std::map
Fixes a typo in the finiteelement `ModelArrayDetails.cpp` file.

timspainNERSC:
Closes #286 

> Fixes a typo in the finiteelement ModelArrayDetails.cpp file.

### DIFF
--- a/core/src/finitevolume/ModelArrayDetails.cpp
+++ b/core/src/finitevolume/ModelArrayDetails.cpp
@@ -73,6 +73,6 @@ ModelArray::DimensionMap::DimensionMap()
 {
 }
 
-const std::vector<ModelArray::Type, ModelArray::Dimension> ModelArray::componentMap = {};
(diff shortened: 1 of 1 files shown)


a-smith-github (approved):
What was the behaviour here, was there a failure or was it handled? If the former, can we have a test? Doesn't hit a changed line in PR 279. 

If the latter and purely the wrong type with little impact outside of this PR, then merge. 

Approving in advance for convenience

timspainNERSC:
Because the file in question has to be included in the model compilation and wasn't , it was not causing any specific problems. I think it is still useful to supply the finiteelement version of ModelArrayDetails, even if it is not going to be used in the main model. However, if it is being kept, then it should at least be correct.

PR #294, opened 2023-04-28
Issue239 restart

### LINKED ISSUE #239: Output restart files at points other than the end of the run
Being able to restart from the end of a run is the bare minimum functionality. The user should be able to specify a repeat time for the creation of restart files. Also, it might be necessary for the model to be able to receive a signal that it should output a restart file at some point, this being my current understand of how it will work coupled to NorESM.

timspainNERSC:
Add the configuration option `model.restart_period` to output restart files with that period after the start. The period can be defined as a ISO P duration string or a number of seconds (as with any other `Duration`).

Closes #239


timspainNERSC:
The slab ocean represents a buffer between the external data on the sea surface conditions supplied either from the coupled ocean model or the external forcings. This allows the ice to remain stable even if the external model is supplying a sea surface temperature above 0˚C.

The slab ocean includes new fields to differentiate between the external sea surface temperature and salinity, included in the `ModelArray::protectedArray` enumeration.

The main part of the slab ocean is the `SlabOcean::update()` function. This calculates updated sea surface temperature and salinity for the slab ocean by using the fluxes calculated in the rest of the model. An additional flux is then calculated to relax the slab values towards the external value with a certain relaxation time. 

The slab ocean is included in the relevant ocean boundary modules as part of the `updateAfter()` step.

Test was written to check that the slab ocean update for a given set of parameters provides the correct heat and salinity flux.

dorchard:
Is there a specific test to check that the restart files are being generated correctly? (sorry I was digging but couldn't find, also sorry I commented on the issue rather than the PR before!)

dorchard:
It would useful to have a test for this, i.e., are the restart files actually getting written and with the right data at the right time. I note that this is on a later milestone, so perhaps not a priority right now.

einola:
I've created a new issue to cover @dorchard's comments (#423). Unfortunately, we must resolve some conflicts that are probably best left to @timspainNERSC. Otherwise, we can merge.

timspainNERSC:
Conflicts resolved, code compiles, tests pass.

core/src/Model.cpp
```diff
@@ -22,23 +22,31 @@
 
 namespace Nextsim {
 
-const std::string Model::restartOptionName = "model.init_file";
+template <typename Key, typename Value>
```
MarionBWeinzierl, line 25: Could some more comments be added to this file, so that it is easier to understand what it is doing?

core/src/include/ModelConfig.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+class Model;
+
+class ModelConfig {
+public:
+    const static std::map<int, std::string> keyMap;
+    static ConfigMap getConfig();
+
+    enum {
+        DUMMY_KEY,
```
MarionBWeinzierl, line 27: What is the dummy key for...?

timspainNERSC: Notice that the configuration keys in `Model.hpp` are assigned values from this class. This is a rather unsatisfactory method, as adding the MPI partition key breaks this system. I have added some comments to explain this, but also moved the MPI configuration key.

core/src/Model.cpp
timspainNERSC: Done

MarionBWeinzierl (approved):

PR #298, opened 2023-05-05
Issue296 zindexanslayer test

### LINKED ISSUE #296: ModelIndex::zIndexAndLayer is not tested
It seemed like such a simple function, but when messing around with array index ordering, it becomes a headache.

Write a test that ensures that `zIndexAndLayer` accesses the correct point in a three-dimensional `ModelArray`.

timspainNERSC:
Add a test for the ModelArray::zIndexAndLayer function.

Will close #296 
 
> It seemed like such a simple function, but when messing around with array index ordering, it becomes a headache.
> 
> Write a test that ensures that zIndexAndLayer accesses the correct point in a three-dimensional ModelArray.

### DIFF
--- a/core/test/ModelArray_test.cpp
+++ b/core/test/ModelArray_test.cpp
@@ -139,7 +139,7 @@ TEST_CASE("Moving data")
 
 TEST_CASE("Instance setDimensions sets instance dimensions")
 {
-    DosDField uu = ModelArray::DosDField();
+    ZUField uu = ModelArray::ZUField();
     ModelArray::MultiDim udim = {5, 5};
     uu.setDimensions(udim);
     REQUIRE(uu.size() == udim[0] * udim[1]);
@@ -256,7 +256,8 @@ TEST_CASE("Location from index")
     REQUIRE(loc[2] == z);
 }
 
-// Test the zIndexAndLayer function
+// Test the zIndexAndLayer function to ensure that it accesses the correct
+// point in a three-dimensional ModelArray.
 TEST_CASE("zIndexAndLayer")
 {
     const size_t nx = 29;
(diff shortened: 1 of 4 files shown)


core/test/ModelArray_test.cpp
```diff
@@ -255,6 +255,36 @@ TEST_CASE("Location from index")
     REQUIRE(loc[1] == y);
     REQUIRE(loc[2] == z);
 }
+
+// Test the zIndexAndLayer function
```
TomMelt, line 259: I think you should expand the description a little bit. At least something similar to issue #296

core/test/testmodelarraydetails/ModelArrayDetails.cpp
```diff
@@ -43,6 +49,7 @@ const std::map<ModelArray::Type, std::string> ModelArray::typeNames = {
     { ModelArray::Type::ONED, "OneDField" },
     { ModelArray::Type::TWOD, "TwoDField" },
     { ModelArray::Type::DOSD, "TwoDField" },
+    { ModelArray::Type::DOSD, "ThreeDField" },
```
TomMelt, line 52: I think this should be `ModelArray::Type::THREED`

As a side note, do we explain anywhere what `DOSD` is? how does it differ from TWOD. I am guessing `DOS` is the number 2 in Spanish?

TomMelt (changes requested):

core/test/ModelArray_test.cpp
timspainNERSC: Copied from the #296 description (with minor editing).

core/test/testmodelarraydetails/ModelArrayDetails.cpp
timspainNERSC: > I think this should be ModelArray::Type::THREED

Indeed it should.

DOSD has now been renamed to the more descriptive ZUFIELD.

TomMelt (approved):
Looks good. Thanks for making the changes.

PR #306, opened 2023-05-10
Issue275 arrayordering

### LINKED ISSUE #275: Array ordering AGAIN
While finally trying to reconcile the dynamics and the thermodynamics, I am now convinced that we need to discuss array ordering.

The multi-indexing in `ModelArray` follows the C/C++ convention that right-hand indices vary fastest. This arises from the form of a naive n-dimensional array, which would be initialised as
```
int matrix[2][3] = {{1, 4, 2}, {3, 6, 8}};
```
The `()` multi-dimensional indexing operator of the `ModelArray` class follows this convention, such that `array(23, 13)` and `array(23, 14)` are consecutive in memory, and would be accessed sequentially when iterating over the raw indices. This results in being able to write a for loop as
```
for (size_t i = 0; i < nx; ++i)
    for (size_t j = 0; j < ny; ++j)
        array(i, j) = something;
```
which access memory locations sequentially and orders indices in reading order from left to right.

In the dynamics, on the other hand, the data is considered to vary with the _x_ coordinate varying fastest. This would result the data points at (_x_, _y_) = (13, 23) and (14, 23) being adjacent in memory.

There are three immediately obvious solutions.

## 1) Swap the coordinates in the dynamics
By swapping _x_ and _y_ in the dynamics, the two arrays are consistent with each other and the C convention for indexing multi-dimensional arrays and nested `for` loops are in reading-order. This is probably a non-starter as so much of the dynamics code is programmed under the x-varies fastest paradigm.

## 2) Swap the coordinates in ModelArray
Again, this will bring the dynamics and ModelArray into consistency. ModelArray would then more resemble a Fortran multi-dimensional array, where dimensions in nested `for`(`DO`) loops are in reverse-reading order
```
DO i = 1, SIZE(a, 2)
  DO j = 1, SIZE(a, 1)
    a(j, i) = something
  END DO
END DO

for (size_t i = 0; i < nx; ++i)
    for (size_t j = 0; j < ny; ++j)
        array(j, i) = something;
```

## 3) Deal with it
Accept that the two conventions are different and work with them as they are. This is a potential source of bugs for any code that interacts with both the dynamics and with `ModelArray`.

timspainNERSC:
I think the array index swapping for the current state of `develop` is complete.

The dynamics is coded such that the x-dimension varies the fastest. This was contradictory to how ModelArray (and C arrays generally) order their indices, where later indices vary faster. This PR ensures that the core/thermodynamics data arrays have the same data ordering as the thermodynamics.

Will close #275 (q.v.).


einola:
@timspainNERSC: Can you fix the conflict? Looks like a simple one.

timspainNERSC:
> @timspainNERSC: Can you fix the conflict? Looks like a simple one.

It really isn't. But fixing this is already on today's To Do list.

timspainNERSC:
This PR has been superseded by #373 which fixes the conflict in ParaGridIO without giving me a headache.

PR #308, opened 2023-05-10
Issue307 confighelp

### LINKED ISSUE #307: Some ConfigHelp text never gets to the terminal
The config help from the freezing point and ice-ocean heat flux modules does not get printed as part of the `--help-config all` text.

timspainNERSC:
Make sure that all possible config help is output to when `--help-config all` is called (or the specific module help is requested).

Will close #307


a-smith-github (commented):
I think the docs need to be updated to say that --help-config is an option and I think, just looking at this change, it isn't quite clear how it ties in (and was a little bit of work to figure out how to test this functions correctly). Imo a comment and (ideally) a test to verify this works would go a long way and would act as dev documentation going forwards. 

Manual verification as follows: 

The output from "./nextsim --help-config-all" is: 

neXtSIM_DG command line options::
  -h [ --help ]         print help message
  --help-config arg     arg = [avail | moduleName | all]
                        print help associated with one or more modules.
                        avail    list all available module names.
                        moduleName    print the help message associated with
                        the given module.
                        all    print all help messages associated with all
                        available modules.
  --config-file arg     specify a configuration file
  --config-files arg    specify a list of configuration files

When using --help-config avail  IIceOceanHeatFlux and IFreezingPoint are listed. 

The output from "./nextsim --help-config IIceOceanHeatFlux" is:


Modules

Modules.Nextsim::IIceOceanHeatFlux
module [Nextsim::BasicIceOceanHeatFlux]
The module for calculating the ice-ocean heat fluxes.

timspainNERSC:
> I think the docs need to be updated to say that --help-config is an option and I think, just looking at this change, it isn't quite clear how it ties in (and was a little bit of work to figure out how to test this functions correctly). Imo a comment and (ideally) a test to verify this works would go a long way and would act as dev documentation going forwards.

While this all true, I don't see that this small PR needs complicating to that degree. I have opened issue #349 to capture this defect and the corresponding PR #350.

einola:
Asking @TomMelt to review this, since Alex has left.

TomMelt (approved):
Looks good to me

PR #316, opened 2023-05-24
Two bug fixes for the Winton thermodynamics

einola:
The first bug fix moves .resize() calls from the constructor to the setData method and adds a call to IIceThermodynamcis::setData to the ThermoWinton::setData method. This ensures that all data are properly initialised, as calling .resize() in the constructor is too early.

The second bug fix is a Celsius vs Kelvin error in the melting point temperature.


timspainNERSC:
We should continue to target `develop` for bug fixes like this and only do big PRs from develop to main. Hopefully the next one will not be to far in the future (TOPAZ/ERA?).

einola:
Ah, yes, but I meant to merge to develop! Too quick on the quick fix.

timspainNERSC (approved):

PR #319, opened 2023-05-30
A demo case for single column thermodynamics

einola:
This PR contains the code needed to run a single-column demonstration of the model thermodynamics, following the forcing from Maykut and Untersteiner (1971). It also includes some minor fixes in the two thermodynamics routines and tests thereof (ThermoWinton and ThermoIce0).

TODO (and will be put into separate issues):
 - The tests show issues with the Winton thermodynamics (also present in Lagrangian neXtSIM) - but this should be addressed in a separate PR.

 - We now set qia as the non-solar ice-atmosphere heat flux, but qow is still the total heat flux. For consistency's sake, this should also be the non-solar flux (this should also help with the coupling).

To run the test use: `./nextsim --config-file config_column.cfg` 

To view the results use: ``python column_pp.py``. The figures should be comparable to [figure 2](https://github.com/nextsimhub/nextsimdg/files/13279071/Winton_figure2.pdf) of Winton (2000), but some work is still required to reproduce exactly his figures 2 and 3. This should be a separate issue.


einola:
This depends now on issue #320 and PR #322

einola:
This PR/branch is now a total mess. It seems that everything of note is already in the june23_demo branch now, but there's things here that are not yet in develop.

timspainNERSC:
Removing myself as a reviewer, as I authored several of the changes in this PR.

timspainNERSC:
To be merged after #389 better dynamics

timspainNERSC:
> How do I run the test case?

`./nextsim --config-file config_column.cfg`

The June '23 demo and the netCDF files that drive it are certainly something we need to discuss, but not directly relevant to this PR.

core/src/include/MonthlyCubicBSpline.hpp
```diff
@@ -0,0 +1,86 @@
...
+    explicit monthlyCubicBSpline(const std::vector<double>& f)
+    {
+        // Create a cyclic B-spline using a fold with a "ghost border"
+        /* We need k+1 ghost points, plus one for each of the derivatives at the end, where k is the
+         * order of the spline
+         * (here k=3). So, with STL's vector.insert() we need ghostWidth = k + 1 + 2 = 6 */
+        const int ghostWidth = 6;
+        std::vector<double> y = f;
+        y.insert(y.begin(), f.end() - ghostWidth, f.end());
+        y.insert(y.end(), f.begin(), f.begin() + ghostWidth);
```
TomMelt, line 52: Nice comments/description. Thanks :+1:

run/column_pp.py
```diff
@@ -0,0 +1,51 @@
...
+        time = np.array(np_time)
+        hice = np.array(root.groups["data"].variables["hice"][0].data)
+        hsnow = np.array(root.groups["data"].variables["hsnow"][0].data)
+        tice = np.array(root.groups["data"].variables["tice"][0].data)
+        first = False
+    else:
+        time = np.append(time, np_time)
+        hice = np.append(hice, root.groups["data"].variables["hice"][0].data)
+        hsnow = np.append(hsnow, root.groups["data"].variables["hsnow"][0].data)
+        tice = np.vstack((tice, root.groups["data"].variables["tice"][0].data))
```
TomMelt, line 25: This could potentially be tidied up by using the `xarray` module. `xarray` works well in python and with netcdf files. Could make reading and plotting of datasets slightly easier (see [here](https://docs.xarray.dev/en/stable/user-guide/io.html#reading-and-writing-files))

run/make_init25kmNH.py
```diff
@@ -160,6 +160,7 @@ def topaz4_interpolate(target_lon_deg, target_lat_deg, data, lat_array):
     mask = datagrp.createVariable("mask", "f8", field_dims)
     sst_data = topaz4_interpolate(element_lon, element_lat, source_file["temperature"][0, :, :].squeeze(), lat_array)
     mask[:, :] = 1 - ma.getmask(sst_data)
+    mask[100, 95] = 0.0 # Ad-hoc fix for the westernmost point of the New Siberian Islands
```
TomMelt, line 163: is this safe to leave in?

run/make_init_column.py
```diff
@@ -0,0 +1,99 @@
...
+zfield_dims = ("z", "y", "x")
+
+# Array coordinates
+node_lon = np.zeros((nfirst + 1, nsecond + 1))
+node_lat = np.zeros((nfirst + 1, nsecond + 1))
+
+node_lon[0, 0] = 0
+node_lon[0, 1] = 270
+node_lon[1, 1] = 180
+node_lon[1, 0] = 90
```
TomMelt, line 55: Are `node_lon` and `node_lat` expected to vary often and if so, would these manual assignments get harder to implement?

run/make_init_column.py
```diff
@@ -0,0 +1,99 @@
...
+
+lat0 = 89.84 # Gives a box around the pole 25 km a side
+
+node_lat[0, 0] = lat0
+node_lat[0, 1] = lat0
+node_lat[1, 1] = lat0
+node_lat[1, 0] = lat0
```
TomMelt, line 62: e.g., this could be simplified to `node_lat = lat0`.

TomMelt (changes requested):
overall looks good. Just a couple of points to address. Happy to have a chat about `xarray` for python, but it's not necessary to implement now. I just thought it might be useful for you in the future.

run/make_init25kmNH.py
timspainNERSC: Yes, but perhaps it should not be in this PR.

For a reason I cannot fathom, the (apparently) same code working on the same data consistently produces a difference in the land-sea mask between the forcing files and this initial restart file. The difference is whether this single point is land or sea. Adding this change makes the two datasets consistent.

Of course, it would be better if the two python files produced the same data, but I have not been able to get to that point.

Since it does not actually concern the single-column thermodynamics tests, this change shouldn't really be in this PR.

run/init_25km_NH.nc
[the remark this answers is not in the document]
timspainNERSC: We desperately need a solution to this. We should organize a meeting to discuss it (or dedicate one of the regular meetings).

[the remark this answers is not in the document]
einola: You can discuss this on Monday. I'll be away, but that's fine :)

einola:
Unfortunately, this test doesn't produce diagnostic output anymore! @timspainNERSC and I will have to figure out why.

timspainNERSC:
> Unfortunately, this test doesn't produce diagnostic output anymore! @timspainNERSC and I will have to figure out why.

To avoid cases of double output at the start of a run, I changed the logic of `ConfigOutput` to only output when the timestep time and output time exactly match. This means that `ConfigOutput::start` must occur on a valid TS. Since this is all integer logic it shouldn't be a problem.

core/test/MonthlyCubicBSpline_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+//
+
+#include "include/MonthlyCubicBSpline.hpp"
+
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+namespace Nextsim {
+
+TEST_SUITE_BEGIN("MonthlyCubicBSpline");
```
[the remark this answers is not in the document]
einola: Yes, good point. I've added a description that should give the purpose of the tests (plus some other comments and minor changes).

physics/src/modules/MU71Albedo.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "include/MU71Albedo.hpp"
+
+namespace Nextsim {
+
+MU71Albedo::MU71Albedo()
+    : snowAlbedo(albedoTable)
+{
+}
+
+std::tuple<double, double> MU71Albedo::albedo(double temperature, double snowThickness, double i0)
```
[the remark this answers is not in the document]
einola: You're right that ``albedo`` is not a descriptive enough name, but it's there because returning ``penSW``, as well as ``albedo``, is an afterthought. Something like ``surfaceSWBalance`` or ``surfaceShortWaveBalance`` would be more correct because the ``doubles`` ``albedo`` and ``penSW`` are the fraction of the incoming shortwave radiation that is reflected into space and transmitted into the interior of the ice, respectively.

The problem is that both names are confusing; ``albedo`` returns more than albedo, which is confusing, and nobody talks about the "surface short-wave balance", so that's also confusing. On the whole, though, I think something like ``surfaceShortWaveBalance`` is probably better, but I'll have a bit of a think to see if I can come up with a better name.

run/column_pp.py
```diff
@@ -0,0 +1,51 @@
...
+
+end = len(time) - 1
```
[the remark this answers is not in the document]
einola: Thanks for the tip!

run/column_pp.py
einola: I've rewritten this post-processing script for Tim's new outputs. It's much simpler now, and I think it works ok without ``xarray``. But I'll keep ``xarray`` in mind for the future.

I also realised that my test description is incomplete, so I've added a bit about the post-processing script.

einola: I've opened issue #431 to describe the improvements still needed for a trivial comparison of the test results with figures from the literature.

run/make_init_column.py
einola: @timspainNERSC: Can you have a look at this one?

run/make_init_column.py
einola: @timspainNERSC: and this one?

Once these are addressed, we can ask @TomMelt to re-review.

run/make_init_column.py
timspainNERSC: For this script, they will never change. In other scripts for creating restart files, they will be different and not hard-coded as they are here. See for example the equivalents in `make_init25kmNH.py`, currently lines 121–124.

run/make_init_column.py
timspainNERSC: Reduced to a single assignment.

core/test/MonthlyCubicBSpline_test.cpp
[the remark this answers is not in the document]
TomMelt: thanks :+1:

run/make_init25kmNH.py
TomMelt: I think I misunderstood your comment:

> Since it does not actually concern the single-column thermodynamics tests, this change shouldn't really be in this PR.

Do this mean leave it in this PR or take it out of this PR?

run/init_25km_NH.nc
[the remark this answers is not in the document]
TomMelt: Looks like we will move towards having an FTP server to store `.nc` files going forward.

TomMelt (changes requested):
All looks good to me. Just waiting to hear back if we are leaving in the line

```python
mask[100, 95] = 0.0 # Ad-hoc fix for the westernmost point of the New Siberian Islands
```

or taking it out of this PR. Otherwise I am happy to merge

run/make_init25kmNH.py
timspainNERSC: I'll create another PR such that is becomes redundant in this PR.

timspainNERSC: See #441 

We should merge #441 before this PR. Do we need to rebase this PR for one line in a python script?

timspainNERSC: PR #441 has been merged.

TomMelt (approved):

PR #329, opened 2023-06-07
Issue328 winton bugs

### LINKED ISSUE #328: Winton errors with no ice
1. The Winton thermodynamics as implemented carries over the code that limits the ice if it drops below the minimum thickness threshold. This does not function correctly with the code in the `IceGrowth` class that updates the concentration.
2. The Winton thermodynamics try to solve the temperature balance even when there is zero ice. This goes poorly (NaN results).

timspainNERSC:
Add tests for the erroneous condition that arise with Winton thermodynamics and no ice and fix those tests.

Closes #328

### DIFF
--- a/physics/src/IceGrowth.cpp
+++ b/physics/src/IceGrowth.cpp
@@ -202,7 +202,7 @@ void IceGrowth::lateralIceSpread(size_t i, const TimestepTime& tstep)
         iLateral->melt(tstep, hice0[i], hsnow[i], deltaHi[i], cice[i], qow[i], deltaCMelt[i]);
     }
     deltaCIce[i] = deltaCFreeze[i] + deltaCMelt[i];
-    cice[i] += (hice[i] > 0) ? deltaCIce[i] : -cice[i];
+    cice[i] = (hice[i] > 0) ? cice[i] + deltaCIce[i] : 0;
     if (cice[i] >= IceMinima::c()) {
         // The updated ice thickness must conserve volume
         updateThickness(hice[i], cice[i], deltaCIce[i], newice[i]);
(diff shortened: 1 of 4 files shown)


draenog (commented):
The commit replaces the existing test with  a new one. Could you explain why the old test is not kept (ideally it should go to a commit message).

draenog (commented):
Why do you set cice to itts negative value instead of setting it to zero?  I am worried what happens if the function is called twice.

timspainNERSC:
> The commit replaces the existing test with a new one. Could you explain why the old test is not kept (ideally it should go to a commit message).

Copy & paste became cut & paste, it seems.

timspainNERSC:
> Why do you set cice to itts negative value instead of setting it to zero? I am worried what happens if the function is called twice.

If the function `lateralIceSpread()` is called twice then on the second call `cice[i]` will already be 0, so `cice[i] -= cice[i]` becomes 0 -= 0, which is still zero.

draenog:
> > Why do you set cice to itts negative value instead of setting it to zero? I am worried what happens if the function is called twice.
> 
> If the function `lateralIceSpread()` is called twice then on the second call `cice[i]` will already be 0, so `cice[i] -= cice[i]` becomes 0 -= 0, which is still zero.

You are right. What confused me is the += operator. But maybe it is an indication that the code is not clear. What about changing this line to:
` cice[i] = (hice[i] > 0) ? cice[i] + deltaCIce[i] : 0.;`

draenog:
I would myself probably clean git history in this repository by squashing 4a118f8, 33cc782 and b06111d. But I am also fine leaving as it is.

timspainNERSC:
I'm currently in a [splitter](https://en.wikipedia.org/wiki/Lumpers_and_splitters) phase when it comes to git commits.

PR #330, opened 2023-06-12
Issue264 netcdf strings

### LINKED ISSUE #264: Using NetCDF4 attributes instead of string variables
NetCDF4 represents variables of type NC_STRING in HDF5 as strings of variable lenght. However there is currently no support in HDF5 of writing such variables in parallel mode. There are two possible solutions:

1. Write string variables as NetCDF attributes not variables.
2. Open the file in serial mode; write all string variables; close the file and reopen it in parallel mode

Currently we use NC_STRING variables only in run metadata so I think it would make sense to represent it as attributes.  For consistency we should probable to the same for all other members of ModelMetadata written to NetCFF4 files.

draenog:
## Change description
The PR replaces NetCDF4 string variables with attributes as writing such variables is not supported in parallel mode. For consistency it also writes all other members of configuration group as attributes.

- [x] formatted time is written as attribute to time variable
- [x] write all members of configuration group as attributes for consistency

### DIFF
--- a/core/src/CommonRestartMetadata.cpp
+++ b/core/src/CommonRestartMetadata.cpp
@@ -31,18 +31,17 @@ netCDF::NcGroup& CommonRestartMetadata::writeRestartMetadata(
 
     // Current time
     netCDF::NcGroup timeGroup = metaGroup.addGroup(timeNodeName());
-    // As a formatted string
-    netCDF::NcVar formVar = timeGroup.addVar(formattedName(), netCDF::ncString);
-    const std::string fTime = metadata.m_time.format();
-    const char* timeCopy = fTime.c_str();
-    formVar.putVar(&timeCopy);
-    formVar.putAtt(std::string("format"), TimePoint::ymdhmsFormat);
     // As Unix time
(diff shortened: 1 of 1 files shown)


core/src/CommonRestartMetadata.cpp
```diff
@@ -31,44 +31,39 @@ netCDF::NcGroup& CommonRestartMetadata::writeRestartMetadata(
...
+
     // All other configuration data
     netCDF::NcGroup configGroup = metaGroup.addGroup(configurationNode());
 
     for (auto entry : metadata.m_config) {
         switch (entry.second.index()) {
         case (CONFIGMAP_DOUBLE): {
-            netCDF::NcVar dblVar = configGroup.addVar(entry.first, netCDF::ncDouble);
-            dblVar.putVar(std::get_if<double>(&entry.second));
+            double value = 1.5;
```
timspainNERSC, line 52: I don't think this is doing anything? Unless it is really subtle.

timspainNERSC (changes requested):
Looks good apart from the possible unused double value.

timspainNERSC (approved):

Issue #120, opened 2022-08-09
MPI parallelization of Dynamics


athelaf:

mondus:
A decision on this work is required. Either

1. Merge now and then add the Dynamics (Milestone) parallelisation at a later date
2. Wait until Dynamics merge, add dynamics parallelisation and merge later.

Preference is perhaps for 1 and then a new PR for the dynamics parallelisation later.

draenog:
Requirements:

1. MPI parallelization should be based on domain decomposition
2. Decomposition should use rectangular boxes
3. Decomposition should take into account land mask for load balancing
4. Decomposition will not change during the run

draenog:
- [ ] Write a preprocessing tool to generate domain decomposition
- [ ] MPI parallelization of thermodynamics where all operations, exept for I/O, are local to an MPI rank
- [ ] MPI parallelization of dynamics

MarionBWeinzierl:
Info from @einola and @timspainNERSC about size of problems to be computed:

> Hi! This is to follow up the discussion we had just now. The question was what the problem size is. So I went and did a bit of 
> checking. The standard development test that I use has about 500k elements. Our coupled run has 300k - but we run it for much longer. A fully coupled climate run at the lowest resolution would have about 100k. We want to be able to run regional setups with much higher resolutions - 10 million would be nice!

> So:
> - Standard coupled: 100k, maybe 1 million
> - Regional: 500k to 2 million
>  - A dream scenario for regional: 10 million

> 
> For reference, the 25kmNH test run has 18k elements, and I don't think we would want the production version of the model to be much slower than that is when running single threaded. So 10k elements per core might be a good rule of thumb.

PR #265, opened 2023-04-06
WIP: XIOS Integration (MVP) 

a-smith-github:
Status: Draft

neXtSIM-DG XIOS Integration - Step 1/MVP

- [ ] CMake Configure to include XIOS, tests
- [ ] XIOS Handler Class with Server Init
    - [ ] Remove temporary and leverage new MPI (Branch???: https://github.com/nextsimdg/nextsimdg/tree/feat_mpi)
- [ ] Add XIOS Calendar Sync
- [ ] Add XIOS Axis Init & Grid Data Sync
- [ ] Add XIOS Output of Restart File

Summary: Provide (optional) file IO via XIOS. The goal of this change is to be able to write a restart file via XIOS using data stored on an XIOS grid after synchronising data and calendar using current data read infrastructure. 

See design document for more information on XIOS Integration: https://github.com/nextsimdg/nextsimdg/wiki/XIOS-Integration

PR #331, opened 2023-06-12
MPI parallelization of thermodynamics

draenog:
## Background 
This PR is part of #120, where the basis of strategy for MPI parallelisation is described.

## Change description
As all thermodynamics operations are local to a grid cell all required MPI communication, is handled by NetCDF4 library. Therefore the only required steps are:

- [x] Initialize and finalize MPI stack
- [x] Read decomposition metadata on each rank
- [x] Read and write the necessary part of the grid on each rank (depends on #330)
- [x] Tests for parallel I/O 

**NOTE** PR #432 should be merged into this branch before it is merged into develop 

To run `run_simple_example.sh` you will need to generate the following netcdf file

```
$ ncdump partition.nc 
netcdf partition {
dimensions:
        P = 1 ;
        L = 1 ;
        globalX = 30 ;
        globalY = 30 ;

group: bounding_boxes {
  variables:
        int global_x(P) ;
        int global_y(P) ;
        int local_extent_x(P) ;
        int local_extent_y(P) ;
  data:

   global_x = 0 ;

   global_y = 0 ;

   local_extent_x = 30 ;

   local_extent_y = 30 ;
  } // group bounding_boxes
}

```

### DIFF
--- a/core/src/RectGridIO.cpp
+++ b/core/src/RectGridIO.cpp
@@ -161,6 +161,8 @@ void RectGridIO::readPartitionData(const std::string& partitionFile, ModelMetada
             + std::to_string(nBoxes) + "\n";
         throw std::runtime_error(errorMsg);
     }
+    metadata.globalExtentX = ncFile.getDim("globalX").getSize();
+    metadata.globalExtentY = ncFile.getDim("globalY").getSize();
     netCDF::NcGroup bboxGroup(ncFile.getGroup(bboxName));
     std::vector<size_t> index(1, metadata.mpiMyRank);
     bboxGroup.getVar("global_x").getVar(index, &metadata.localCornerX);
--- a/core/src/RectGridIO.cpp
+++ b/core/src/RectGridIO.cpp
@@ -3,6 +3,7 @@
  *
  * @date Feb 8, 2022
  * @author Tim Spain <timothy.spain@nersc.no>
+ * @author Kacper Kornet <kk562@cam.ac.uk>
  */
 
 #include "include/RectGridIO.hpp"
@@ -46,7 +47,11 @@ void dimensionSetter(
     ModelArray::setDimensions(type, dims);
 }
 
+#ifdef USE_MPI
+ModelState RectGridIO::getModelState(const std::string& filePath, const std::string& partitionFile)
+#else
 ModelState RectGridIO::getModelState(const std::string& filePath)
+#endif
 {
     ModelState state;
     netCDF::NcFile ncFile(filePath, netCDF::NcFile::read);
--- a/core/src/RectGridIO.cpp
+++ b/core/src/RectGridIO.cpp
@@ -48,7 +48,8 @@ void dimensionSetter(
 }
 
 #ifdef USE_MPI
-ModelState RectGridIO::getModelState(const std::string& filePath, const std::string& partitionFile)
+ModelState RectGridIO::getModelState(
+    const std::string& filePath, const std::string& partitionFile, ModelMetadata& metadata)
 #else
 ModelState RectGridIO::getModelState(const std::string& filePath)
 #endif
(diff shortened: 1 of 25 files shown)


dorchard:
`develop` branch now in a position for this to be turned into a non-draft PR, once the conflicts are resolved, which @draenog will do next week.

TomMelt:
Based on discussion with Tim and Einar, this should be rebased onto `feat_mar3`. I will then look to rebase #265 onto this (`mpi_local`) branch.

einola:
We need to be careful not to use MPP_WORLD as the communicator because the model should receive the communicator from OASIS in a coupled mode.

draenog:
> We need to be careful not to use MPP_WORLD as the communicator because the model should receive the communicator from OASIS in a coupled mode.

Yes, it is a good point. I think it is currently quite well encapsulated. Constructor of  Nextsim:Model takes MPI communicator as an argument which is then saved in ModelMetadata.

TomMelt:
I have rebased this branch to `develop` to that it is ready to be merged in. First however, we should merge #432 into this branch. For some reason github is showing the incorrect diff. I think maybe the GUI is broken. The actual diff is quite small:

```
$ git ds mpi_local
 core/src/RectGridIO.cpp               |  42 +++++++--------
 core/test/CMakeLists.txt              |  30 +++++++++++
 core/test/MPI_example_tests.cpp       |  46 ++++++++++++++++
 core/test/MainMPI.cpp                 |  19 +++++++
 core/test/RectGrid_test.cpp           |  24 +++++++++
 lib/doctest/doctest.h                 |  33 ++++++++----
 lib/doctest/extensions/doctest_mpi.h  | 169 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lib/doctest/extensions/doctest_util.h |  37 +++++++++++++
 lib/doctest/extensions/mpi_reporter.h | 271 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lib/doctest/extensions/mpi_sub_comm.h |  84 +++++++++++++++++++++++++++++
 10 files changed, 724 insertions(+), 31 deletions(-)

```

core/src/RectGridIO.cpp
```diff
@@ -22,13 +23,31 @@
...
+    ModelArray::MultiDim dims;
+    dims.resize(nDims);
+    dims[0] = metadata.localExtentX;
+    dims[1] = metadata.localExtentY;
+    for (size_t d = 2; d < nDims; ++d) {
+        dims[d] = dataGroup.getVar(fieldName).getDim(d).getSize();
+    }
+    ModelArray::setDimensions(type, dims);
+}
+#else
```
timspainNERSC, line 50: Is this still correct in a post-dimension ordering swap world?

core/src/StructureFactory.cpp
```diff
@@ -41,25 +42,42 @@ std::string structureNameFromFile(const std::string& filePath)
...
     // TODO There must be a better way
     if (DevGrid::structureName == structureName) {
         Module::setImplementation<IStructure>("DevGrid");
         DevGrid gridIn;
         gridIn.setIO(new DevGridIO(gridIn));
+#ifdef USE_MPI
+        return gridIn.getModelState(filePath, partitionFile, metadata);
+#else
         return gridIn.getModelState(filePath);
+#endif
```
timspainNERSC, line 62: DevGrid has been removed. Is there a need to rebase? Or will the merge/PR sort this all out?

core/src/modules/include/DevGrid.hpp
timspainNERSC, line 1: DevGrid no longer exists in develop.

timspainNERSC (changes requested):
* Need to check that the changes to DevGrid go through cleanly now that it doesn't exist in `develop`.
* Need to check DimensionSetter() is still correct in the MPI version now that the ordering of the array dimensions has been swapped.

Otherwise, this looks like a god set of MPI changes.

core/src/StructureFactory.cpp
MarionBWeinzierl: I think `DevGrid` here can be replaced by `RectGrid` and it should (hopefully) be fine.

core/src/RectGridIO.cpp
timspainNERSC: I see this is dealt with in #432

timspainNERSC (approved):

Issue #338, opened 2023-07-03
Making the physics truly do nothing


timspainNERSC:
Currently, the ice thermodynamics can be turned off by using `DummyIceThermodynamics` and the ice spread disabled by using `DummyIceSpread` as the implementations for `IIceThermodynamics` and `LateralSpread`, respectively.

To entirely disable any changes to the ice by the physics calculations, the ice growth due to the freezing of sea water also need to be prevented. This can be done by implementing a new derived class of `IFluxCalculation` which returns 0 for the open water heat flux. The question is then whether this implementation should also zero all other calculated fluxes. Also, this would only be called by atmospheric boundary modules that actually call the flux calculation routine. Other boundary modules, such as `FluxConfiguredOcean` may result in a situation where all of the 'turn off the physics' implementations have been configured to be used, but the ocean freezing still occurs.

Another alternative would be to wrap the top-level call to the physics in an if statement with an additional model configuration option selecting to turn off the thermodynamics. Does this already exist in neXtSIM-LG?

timspainNERSC:
Yes, `thermo.use_thermo_forcing`, `nextsim/options.cpp`:L371

timspainNERSC:
Fix merged as PR #340.

PR #339, opened 2023-07-03
Issue338 Option to turn off ice physics

timspainNERSC:
## Fixes \#338

Add an option to turn off the ice physics (metonymically: thermodynamics). Prevent any updates to the ice concentration, ice thickness or show thickness fields. However, do continue to calculate the ice-averaged values of those fields. The option name is ported from the equivalent option in Lagrangian neXtSIM, becoming `nextsim_thermo.use_thermo_forcing`. This is a boolean configuration option which defaults to true.

---
# Test

Add an additional test case to the existing `IceGrowth` class test suite. This re-runs one of the previous tests, except with the configuration option set to `false`.

---
# Documentation Impact

Actually documented the options in the 'nextsim_thermo' configuration section.
Added the description of the new `use_thermo_forcing` option to that section.


timspainNERSC:
This was clearly a bad choice of branch to start from.

PR #340, opened 2023-07-03
Issue338 no thermo

timspainNERSC:
## Fixes \#338

Add an option to turn off the ice physics (metonymically: thermodynamics). Prevent any updates to the ice concentration, ice thickness or show thickness fields. However, do continue to calculate the ice-averaged values of those fields. The option name is ported from the equivalent option in Lagrangian neXtSIM, becoming `nextsim_thermo.use_thermo_forcing`. This is a boolean configuration option which defaults to true.

---
# Test

Add an additional test case to the existing `IceGrowth` class test suite. This re-runs one of the previous tests, except with the configuration option set to `false`.

---
# Documentation Impact

Actually documented the options in the 'nextsim_thermo' configuration section.
Added the description of the new `use_thermo_forcing` option to that section.

### DIFF
--- a/physics/src/IceGrowth.cpp
+++ b/physics/src/IceGrowth.cpp
@@ -143,9 +143,6 @@ void IceGrowth::update(const TimestepTime& tsTime)
                      std::placeholders::_2),
         tsTime);
 
-    // reset the new ice volume array
-    newice = 0;
-
     if (doThermo) {
         iVertical->update(tsTime);
         // new ice formation
@@ -166,6 +163,9 @@ void IceGrowth::initializeThicknesses(size_t i, const TimestepTime&)
         hice[i] = hice0[i] = 0.;
         hsnow[i] = hsnow0[i] = 0.;
     }
+
+    // reset the new ice volume array
+    newice = 0;
 }
 
 void IceGrowth::newIceFormation(size_t i, const TimestepTime& tst)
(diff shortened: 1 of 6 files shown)


physics/src/IceGrowth.cpp
```diff
@@ -132,18 +137,22 @@ ConfigMap IceGrowth::getConfiguration() const
...
                      std::placeholders::_2),
         tsTime);
 
-    iVertical->update(tsTime);
-    // new ice formation
-    overElements(
-        std::bind(&IceGrowth::updateWrapper, this, std::placeholders::_1, std::placeholders::_2),
-        tsTime);
+    // reset the new ice volume array
+    newice = 0;
```
einola, line 147: I don't understand why ``newice`` is set to zero here. Isn't it enough to do it in line 195? And if we still need to initialise it, shouldn't that be done in ``initializeThicknesses``?

einola (approved):
It's minor revisions ;) - i.e. one comment you can respond to as you see fit and then merge.

physics/src/IceGrowth.cpp
timspainNERSC: The tricky thing was this at line 195 
```
       newice[0] = 0;
```
This has now been corrected with the commit that fixes it in other branches.

Setting `newice`to zero there ensures that it is zero even if `doThermo` is false. In that case, the code at line 195 is never approaches, as the `lateralIceSpread` function is never called.

In any case, the zeroing has been moved to `initializeIceThicknesses` it does make sense there.

einola: Ah, and I was looking at that line and didn't see the bug! :D Nice fix!

Issue #352, opened 2023-08-09
Homebrew no longer extracts casks in the test environments


timspainNERSC:
After an update from v4.0.28 to v4.1.3 the macOS package manager `homebrew` began failing to extract packages in the test workflow environment, causing the tests to fail. Extracting, rather than installing, is an unusual thing to do and since this is now causing problems, should be eliminated.

One of the extractions is to obtain the most recent v2.xx of catch2, before the change to the multi-header code layout. Since catch2 is no longer used, this can be removed, meaning (hopefully) that the Mac compile and test workflow will complete.

The other is to get version v13.0.1 of clang-format. Why is this?

PR #353, opened 2023-08-09
Remove homebrew extractions from the test workflows

timspainNERSC:
# Remove homebrew extractions from the test workflows
## Fixes \#352

---

# Change Description

After an update from v4.0.28 to v4.1.3 the macOS package manager homebrew began failing to extract packages in the test workflow environment, causing the tests to fail. Extracting, rather than installing, is an unusual thing to do and since this is now causing problems, should be eliminated.

One of the extractions is to obtain the most recent v2.xx of catch2, before the change to the multi-header code layout. Since catch2 is no longer used, this can be removed, meaning (hopefully) that the Mac compile and test workflow will complete.

The other is to get version v13.0.1 of clang-format.

---
# Test Description

If the workflows on github successfully complete, then it is fixed. If they do not, it is not fixed.

### DIFF
--- a/.github/workflows/clang-compile-tests.yml
+++ b/.github/workflows/clang-compile-tests.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: clang
       run: |
-        brew extract --version=13.0.1 clang-format homebrew/cask
+        brew extract --version=13.0.1 clang-format homebrew/cask-versions
         brew install clang-format@13.0.1
         cd core/src
         clang-format --dry-run -Werror *cpp include/*hpp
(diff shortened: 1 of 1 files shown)


timspainNERSC:
Solution thanks to stackexchange.com user Jay Zhou in a comment on [this answer](https://stackoverflow.com/a/66458452):

> it can not extract to homebrew/cask and homebrew/core now, but it can still extract to homebrew/cask-versions – 
[Jay Zhou](https://stackoverflow.com/users/4619189/jay-zhou)
[Jul 27 at 9:23](https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula/66458452#comment135356534_66458452)

Issue #347, opened 2023-08-03
improve cmake build efficiency


TomMelt:
Currently running the cmake build process (see below) takes around 30 mins using 4 processors on my `Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz`.

```sh
# build process
cmake ..
make -j 4
```

This is due to duplicated recompilation of the core source code. 
* each individual test currently recompiles core source files in order to build the test binary
* This makes the build time (`t_total`) roughly proportional to `t_total = N_tests * t_core` where `N_tests` is the number of tests and `t_core` is the time taken to compile the core source files.

![image](https://github.com/nextsimdg/nextsimdg/assets/25719944/90226f65-0d1c-473e-9e2a-1445c531efba)

My first attempt to fix this issue can be seen in branch [issue347-improve-cmake-build-efficiency](https://github.com/nextsimdg/nextsimdg/compare/develop...issue347-improve-cmake-build-efficiency)
* I have compiled the majority of source code all into one library `nextsimlib`
* this is then used to build the main executable and other tests e.g.,
   - https://github.com/nextsimdg/nextsimdg/blob/66bd28fd157162915b0b73550699afc116c097bc/CMakeLists.txt#L171
   - https://github.com/nextsimdg/nextsimdg/blob/66bd28fd157162915b0b73550699afc116c097bc/core/test/CMakeLists.txt#L16
* this dramatically speeds up compilation time

So far this is a proof of concept. It **does not currently work** for all tests (see example of errors below):

```
/usr/bin/ld: CMakeFiles/testRectGrid.dir/RectGrid_test.cpp.o: warning: relocation against `_ZN7Nextsim10ModelArray6m_dimsE' in read-only section `.text._ZN7Nextsim10ModelArray10dimensionsENS0_4TypeE[_ZN7Nextsim10ModelArray10dimensionsENS0_4TypeE]'
/usr/bin/ld: CMakeFiles/testRectGrid.dir/RectGrid_test.cpp.o: in function `Nextsim::DOCTEST_ANON_FUNC_14()':                                                 
RectGrid_test.cpp:(.text+0x159f7): undefined reference to `Nextsim::NZLevels::set(unsigned long)'                                                            
/usr/bin/ld: RectGrid_test.cpp:(.text+0x15a82): undefined reference to `Nextsim::ModelArray::setDimensions(Nextsim::ModelArray::Type, std::vector<unsigned long, std::allocator<unsigned long> > const&)'
```

[...]

Issue #210, opened 2022-12-16
Refactor CMake


timspainNERSC:
The CMake system works, but isn't necessarily the best. It probably needs refactoring once the functionality to built the complete model exists.

TomMelt:
closed by  #449

PR #366, opened 2023-08-17
Issue347 improve cmake build efficiency - Work in progress

TomMelt:
# Pull Request Title

---
# Change Description

* Previous build of code base took ~45 mins (e.g. see [github actions](https://github.com/nextsimdg/nextsimdg/actions/runs/5855514762)
* I have modified the following `CMakeLists.txt` files:
  * `core/test/CMakeLists.txt`
  * `dynamics/applications/CMakeLists.txt`
  * `dynamics/tests/CMakeLists.txt`
  * `physics/test/CMakeLists.txt`
* They should be a lot simpler to read and should be faster to compile.
* I have removed build duplication by linking against a compiled lib rather than rebuilding every cpp file for each executable (there are many tests, each are separate exe's)
* In my first attempt I have created two libraries:
  * the main library containing the majority of nextsimdg code (`nextsimlib`)
  * a smaller library containing code for dynamics (`dynamicslib`)
* These libraries can be merged or split further depending on use-case in future

* **My primary goal was to improve build time not necessarily thinking about how best to structure the code into libraries. This can be discussed at a later stage.**

---
# Documentation Impact

* shouldn't need any changes to docs but I might update the build instructions anyway as they are a bit out of date e.g., they still refer to `Catch2` which is no longer used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,13 +87,85 @@ foreach(compo ${CodeComponents})
     add_subdirectory("${compo}")
 endforeach()
 
-# Link the model
-add_executable(nextsim "${NextsimSources}")
+set(sources
+"./core/src/Logged.cpp"
+"./core/src/Timer.cpp"
+"./core/src/Model.cpp"
+"./core/src/Iterator.cpp"
+"./core/src/Configurator.cpp"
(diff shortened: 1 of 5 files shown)


TomMelt:
This PR has been replaced by #369 .

PR #369, opened 2023-08-22
add libs to CMakeLists.txt to remove build duplication

TomMelt:
# Pull Request Title

---
# Change Description

* Previous build of code base took ~45 mins (e.g. see [github actions](https://github.com/nextsimdg/nextsimdg/actions/runs/5855514762))
* I have modified the `CMakeLists.txt` files
* They should be a lot simpler to read and faster to compile.
* I have removed build duplication by linking against a compiled lib rather than rebuilding every cpp file for each executable (there are many tests, each are separate exe's)

* In my first attempt I have created one main library containing the majority of nextsimdg code (`nextsimlib`)
* This library could be split into modular libraries depending on use-case in future

* **My primary goal was to improve build time not necessarily thinking about how best to structure the code into libraries. This can be discussed at a later stage.**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,13 +88,20 @@ foreach(compo ${CodeComponents})
     add_subdirectory("${compo}")
 endforeach()
 
-# Link the model
-add_executable(nextsim "${NextsimSources}")
+add_library(nextsimlib SHARED ${NextsimSources})
+target_include_directories(nextsimlib
+    PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}"
+    "${NextsimIncludeDirs}"
+    PUBLIC
+    "${netCDF_INCLUDE_DIR}"
+    )
+target_link_directories(nextsimlib PUBLIC "${netCDF_LIB_DIR}")
+target_link_libraries(nextsimlib PUBLIC Boost::program_options Boost::log "${NSDG_NetCDF_Library}" Eigen3::Eigen)
 
+add_executable(nextsim ./core/src/main.cpp)
 target_include_directories(nextsim PRIVATE
     "${CMAKE_CURRENT_SOURCE_DIR}"
     "${NextsimIncludeDirs}"
-    "${netCDF_INCLUDE_DIR}"
     )
-target_link_directories(nextsim PUBLIC "${netCDF_LIB_DIR}")
-target_link_libraries(nextsim LINK_PUBLIC Boost::program_options Boost::log "${NSDG_NetCDF_Library}" Eigen3::Eigen)
+target_link_libraries(nextsim PUBLIC nextsimlib)
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -1,7 +1,6 @@
 # Sources for the main neXtSIM model
 
 set(BaseSources
-    "main.cpp"
     "Logged.cpp"
     "Timer.cpp"
     "Model.cpp"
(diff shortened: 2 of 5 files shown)


TomMelt:
> The other simple solution would be to just build these tests directly from the sources. Neither uses many source files (2 and 3 source files, respectively).

I think this approach makes the most sense. I will build against source :+1:

TomMelt:
@timspainNERSC it looks like the mac build `build-and-tests-on-mac` is failing. I don't have a mac. Would you be able to find out why the test fails?

Also, it's worth noting. Unfortunately the CI workflow does not show the test name -- so I cannot even see what test has failed. I have created a PR #372 to fix this and tagged you as a reviewer.

timspainNERSC:
> Would you be able to find out why the test fails?

Apparently not! Building the library fails, due to being unable to find the `<Eigen/Dense>` header:
```
/Users/timpai/nextsim/dg/nextsimdg/dynamics/applications/../src/include/ParametricMesh.hpp:10:10: fatal error: 'Eigen/Dense' file not found
#include <Eigen/Dense>
         ^~~~~~~~~~~~~
1 error generated.
```
Yet the previous build system was able to find this file without any problem. And I cannot see any changes in CMakeLists that would cause to occur. Very puzzling.

TomMelt:
> > Would you be able to find out why the test fails?
> 
> Apparently not! Building the library fails, due to being unable to find the `<Eigen/Dense>` header:
> 
> ```
> /Users/timpai/nextsim/dg/nextsimdg/dynamics/applications/../src/include/ParametricMesh.hpp:10:10: fatal error: 'Eigen/Dense' file not found
> #include <Eigen/Dense>
>          ^~~~~~~~~~~~~
> 1 error generated.
> ```
> 
> Yet the previous build system was able to find this file without any problem. And I cannot see any changes in CMakeLists that would cause to occur.
> 
> Also, the CI manages to build successfully, and it more or less the same set up as on my Mac. Very puzzling.

Can you check the version of Eigen you have on your mac and maybe check it is loaded/ in your PATHs?

timspainNERSC:
> Can you check the version of Eigen you have on your mac and maybe check it is loaded/ in your PATHs?

Which paths? Aren't PATHs exactly what CMake sets up with `find_module` and so on?

I don't think it is an installation issue: if I checkout `develop` from this branch without changing anything else, then that builds fine.

TomMelt:
@timspainNERSC  and I have fixed the issue of building on mac by cleaning the build environment. I suspect that `make` was picking up the old `nextsimlib` and trying to build against that.

`./testModelArrayRefDebugging` was failing on the CI mac build. This was due to an error in `core/test/CMakeLists.txt`. I have removed erroneous link to `nextsimlib` and added a link to `eigen` to fix it.

einola:
I moved this to milestone two because it's not milestone one. What's the status? We've had some more important issues coming up in the last weeks, but we can't forget about this because of those.

TomMelt:
It was ready to go in before the dynamics merge. I think @timspainNERSC didn't want to merge it yet though. I think he still has reservations about this approach. Happy to discuss options when Tim is back.

Current state is that it needs rebasing off of `feat_mar3` if it is to go into the next milestone. 

I can do this if everyone agrees but we would need to merge it back in before any other big changes. The problem is that the `CMakeLists.txt` files are central to the repo structure and if we have any big merges this will generally cause quite a few merge conflicts.

It shouldn't take me more than a couple of hours to rebase when we need it, so please just let me know and we can look at merging it back in.

TomMelt:
@einola , I have rebased against `feat_mar3` so that it is ready to go when @timspainNERSC  is back

einola:
Thanks @TomMelt! We'll get @timspainNERSC's feedback when he's back.

timspainNERSC:
I certainly think this is in a state where it can be converted to a non-draft PR. Then I will take a more thorough look at it as part of the review.

timspainNERSC (approved):
Am I getting around to reviewing this because I realised I needed the code: maybe?
In any case, this looks good and more importantly, works.

TomMelt:
superseded by #449

Issue #379, opened 2023-09-04
Remove netCDF test and init files


timspainNERSC:
The netCDF files that are currently included in the repo take a lot of the limited storage space. Many of them are generated from python script and could be recreated as part of the build process, thereby saving the storage space.

timspainNERSC:
All of the .nc files are generated from included python scripts and simple data sources, with the exception of `init_25km_NH.nc`, which requires the sizeable TOPAZ4 data files to be generated.

PR #407, opened 2023-09-21
WIP: Issue379 netCDF generator

timspainNERSC:
# netCDF file generation
## Fixes \#379

---
# Change Description

To avoid cluttering the repo with binary netCDF files (as far as possible), generate those files that can be generated from python scripts. This involves adding custom targets and commands to the `CMakeLists.txt` files to run python on the necessary scripts.

---
# Test Description

Since the ERA5 and TOPAZ tests depend on some of these netCDF files, if these tests both continue to pass without file not found errors, then the changes relating to this issue can be considered to pass testing.


timspainNERSC:
Almost all of the netCDF files have been removed from the repository, so this PR has been fulfilled elsewhere.

PR #408, opened 2023-09-21
fix CMakeLists.txt to build nextsim

TomMelt:
nextsim build fails due to incorrect order of libraries

-lxios needs to come before -lnetcdf

also

for gnu compiler we need -lgfortran and for intel compiler we need -lifcore

TODO:

Findxios in conjuction with the CMakeLists.txt needs fixing still so that it can work when xios is not installed.

@a-smith-github , we should have a think about how best to generalize this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,11 @@ endif()
 
 find_package(MPI REQUIRED)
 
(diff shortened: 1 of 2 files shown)


a-smith-github (approved):
Thanks Tom. Manually verified in group call. Accepting.

PR #424, opened 2023-10-16
Melt/feature218 xios io/testmpi

TomMelt:
This allows the XIOS test to use mpi-aware doctest.


TomMelt:
This PR is not required any more.

Issue #425, opened 2023-10-23
The example runs no longer function correctly


timspainNERSC:
As of commit `1b1dda8` in the `develop` branch the example runs of the model do not function correctly. On my Mac the following results are observed:

- `run_simple_example.sh`
   - output is produced, but has dimension 1x30, instead of the expected 30x30
- `run_topaz128x128.sh`
   - crashes with a netCDF exception

Both of these examples should run for all of their short run length and produce valid data.

timspainNERSC:
The `simple_example` runs to completion, but we may need some discussion on the exact values of the fluxes used, mainly Qia.

timspainNERSC:
The `topaz128x28` example runs almost to completion, hitting a segfault at the very end of the run. This is caused by the netCDF diagnostic file not being correctly closed. The `FileCallbackCloser` class solves this and will be included in a near future PR. I'd rather not have to disentangle it (again) for this issue.

timspainNERSC:
PR #426 was merged.

PR #383, opened 2023-09-08
Issue215 configoutput

timspainNERSC:
# ConfigOutput enhancements
## Fixes \#215

---

# Change Description

* Correct the output times of ConfigOutput.
* Reuse files for multiple output instances, rather than creating a new file each time.

---

# Test Description

The test ConfigOutput_test tests this class.

This test shall remain a draft until PR #373 is merged into develop.

PR #426, opened 2023-10-26
Issue425 Example runs should run

timspainNERSC:
# Example runs should run
## Fixes \#425

# Change Description

Fixes the example runs launched from `run_simple_example.sh` and `run_topaz128x128.sh` so that they run and do not crash, at least not until the end of the run.

---
# Test Description

The runs now execute mostly successfully.


MarionBWeinzierl (commented):
That ran for me with test_simple_example.sh and did not fail.

What do you mean with "runs mostly successful", though? The failing checks?

timspainNERSC:
> That ran for me with test_simple_example.sh and did not fail.
> 
> What do you mean with "runs mostly successful", though? The failing checks?

There was also a failing unit test, now fixed. But the other example `topaz128x128` will complete its calculations and then crash with a segfault on exit as the netCDF files are not properly closed. The code to fix this is in PR #383, which is one of the higher priority of the waiting PRs. It is tied into the rest of the code in that PR in such a way that I'd rather not do the work to disentangle just the section that would prevent this crash.

MarionBWeinzierl (approved):
LGTM, issue mentioned above will be fixed in other PR

PR #432, opened 2023-11-09
Mpi local tests

TomMelt:
# Add Tests for MPI version of code

### Task List
- [ ] add tests
- [ ] rebase to develop
- [ ] fix array ordering

---
# Change Description

I rebased @draenog's `mpi_local` and `mpi_local_tests` branches. This introduced a bug due to array reordering that occurred as part of #426. I have modified the MPI parts of `./core/src/RectGridIO.cpp` to reorder array so that they now match the order in the serial version of the code.

---
# Test Description

To run `testRectGrid_MPI`


MarionBWeinzierl:
cmake has to be run with `-DENABLE_MPI`, otherwise this option
https://github.com/nextsimhub/nextsimdg/blob/8a0a730dae65be09b27055dd0c1bdc7ff89962fe/CMakeLists.txt#L53
is off. If this is off cmake does not run through and I get this error:
```
CMake Error at core/test/CMakeLists.txt:151 (target_link_libraries):
  Target "testRectGrid_MPI" links to:

    MPI::MPI_CXX

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

Also, `netcdf_par.h` is included in`core/src/ParallelNetcdfFile.cpp:3:10` and needs to be installed (through netcdf4-parallel?).

TomMelt:
> cmake has to be run with `-DENABLE_MPI`, otherwise this option
> 
> https://github.com/nextsimhub/nextsimdg/blob/8a0a730dae65be09b27055dd0c1bdc7ff89962fe/CMakeLists.txt#L53
> 
> is off. If this is off cmake does not run through and I get this error:
> 
> ```
> CMake Error at core/test/CMakeLists.txt:151 (target_link_libraries):
>   Target "testRectGrid_MPI" links to:
> 
>     MPI::MPI_CXX
> 

Thanks @MarionBWeinzierl , I have pushed 2 commits:
* The first (b0b4f29) fixes this issue. I had forgotten to surround the mpi test binary with `if(ENABLE_MPI)`
* The second (3d2504f) removes unnecessary test binary I no longer need (`testMPI`)

TomMelt:
This PR is ready for review. I will work on adding information to the README and docs, but this will not change the content.

core/test/RectGrid_test.cpp
```diff
@@ -18,10 +23,16 @@
 #include <fstream>
 
 const std::string filename = "RectGrid_test.nc";
+const std::string partition_filename = "partition_metadata_1.nc";
```
timspainNERSC, line 26: Add a TODO (or FIXME?) to ensure that this netCDF file gets removed and generated when the issue #379 changes are merged.

timspainNERSC (changes requested):
Just one change: add a comment about the ParaGridIO_input_test.nc file to ensure no-one (me) forgets about it when the rest of the netCDF files are removed.

Otherwise, most of the changes are to doctest files, so I assume these are correct.

core/test/RectGrid_test.cpp
TomMelt: I will address this as part of #455 .

Issue #271, opened 2023-04-19
Implement the dynamics benchmark winds & currents


timspainNERSC:
The dynamics benchmark uses an analytic set of wind and current forcings, coded into the benchmark applications themselves. These should be ported to implementations of `IAtmosphereBoundary` and `IOceanBoundary` to allow the benchmark to run wholly within nextSIM-DG.

Since the benchmark will need the dummy thermodynamics, I guess this issue is coupled to issue #267.

timspainNERSC:
Moving to Milestone 2: Functional Tests, since solving this issue is not essential for the merge itself.

PR #435, opened 2023-11-10
Issue 271 Analytic forcing

timspainNERSC:
# Analytic forcing
## Fixes \#271

# Change Description

Running the dynamics benchmark requires specific forcing velocities from the atmosphere and ocean. This change implements them and some necessary code changes to ensure the benchmark test runs correctly.

---
# Test Description

A new example run configuration is included, `config_dynamics_benchmark.cfg`. Running the model with this configuration should result in a `diagnostic.nc` file being produced with 15 time samples, showing a cyclonic storm moving the ice of the `init_25kmNH.nc`restart file. When the integration test framework is introduced, this may well be included therein.


physics/src/modules/BenchmarkAtmosphere.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    // distance from the centre of the cyclone
+    const ModelArray& xPrime = BenchmarkCoordinates::x() - x0;
+    const ModelArray& yPrime = BenchmarkCoordinates::y() - y0;
+
+    // Perform the rest of the calculation per element
+    for (size_t j = 0; j < BenchmarkCoordinates::ny(); ++j) {
+        for (size_t i = 0; i < BenchmarkCoordinates::nx(); ++i) {
+            // Expression taken from the original implementation:
+            // double scale = exp(1.0) / 100.0 * exp(-0.01e-3 * sqrt(SQR(x - cMx) + SQR(y - cMy))) * 1.e-3;
```
MarionBWeinzierl, line 69: Is this equivalent to the below, or why has it changed (and why is the original kept)?

MarionBWeinzierl (approved):
It looks good to me, as far as I can tell

PR #437, opened 2023-11-14
Remove devgrid files and usage

### LINKED ISSUE #433: Fail for devgrid in the rect10x10 test case
When running `run_rect10x10.sh` I get the following error:

```
 terminate called after throwing an instance of 'netCDF::exceptions::NcHdfErr'
   what():  NetCDF: HDF error
 file: ncVar.cpp  line:795
 Aborted (core dumped)
```

This error does not occur, for example, for `run_simple_example.sh` or `run_rect20x30.sh`. Inspecting the respective `.nc` files shows that the 10x10 case uses the group attribute type `devgrid`, while the others use `metagrp.type = "simple_rectangular"` or `parametric_rectangular`. 

I changed the type from `devgrid` to `simple_rectangular` for 10x10 in #331 . However, it needs further investigation into the `devgrid` implementation to find out where the problem lies and whether it can be fixed, and/or `devgrid` can be abandoned and removed from the codebase, falling to `simple_rectangular ` for the cases where we would have used it.

This decision will impact #292 .

MarionBWeinzierl:
# Remove DevGrid and associated tests
### Task List
- [x] Remove files for DevGrid, DevGrid IO and associated tests
- [x] Adapt code and remove any usage or mention of DevGrid
- [x] Remove the `rect10x10` test
- [x] Make sure I haven't broken anything and all the tests run through

---
# Change Description
The DevGrid functionality is no longer needed, as it can be fullfilled by the `RectangularGrid` or `ParametricGrid`. Therefore it was removed from the code base, and the tests and cmake lists adapted accordingly.


MarionBWeinzierl:
@timspainNERSC @einola What about DevStep, what is that used for, and should that go as well?

core/src/modules/IStructureModule.cpp
```diff
@@ -7,27 +7,24 @@
...
 
 template <>
 Module<Nextsim::IStructure>::map Module<Nextsim::IStructure>::functionMap = {
-    { DEV_GRID, newImpl<Nextsim::IStructure, Nextsim::DevGrid> },
     { RECT_GRID, newImpl<Nextsim::IStructure, Nextsim::RectangularGrid> },
     { PARAMETRICGRID, newImpl<Nextsim::IStructure, Nextsim::ParametricGrid> },
 };
 template <>
-Module<Nextsim::IStructure>::fn Module<Nextsim::IStructure>::spf = functionMap.at(DEV_GRID);
+Module<Nextsim::IStructure>::fn Module<Nextsim::IStructure>::spf = functionMap.at(RECT_GRID);
```
MarionBWeinzierl, line 24: @einola @timspainNERSC I am not sure about this part -- can it go, or should DEV_GRID be replaced by RECT_GRID? Is it needed somewhere?

core/src/modules/IStructureModule.cpp
```diff
@@ -7,27 +7,24 @@
...
 template <>
 std::unique_ptr<Nextsim::IStructure> Module<Nextsim::IStructure>::staticInstance
-    = std::move(newImpl<Nextsim::IStructure, Nextsim::DevGrid>());
+    = std::move(newImpl<Nextsim::IStructure, Nextsim::RectangularGrid>());
```
MarionBWeinzierl, line 27: Same here as above

timspainNERSC:
> @timspainNERSC @einola What about DevStep, what is that used for, and should that go as well?

`DevStep` was never superseded. It is used for iterating the model state. Perhaps it should be renamed.

core/src/modules/IStructureModule.cpp
timspainNERSC: It should be replaced by `RECT_GRID`. This line designates the default implementation for this Module.

core/src/modules/IStructureModule.cpp
timspainNERSC: Indeed. Once again replace with DevGrid.

timspainNERSC (approved):
The removal of DevGrid looks good, and thanks for picking up the whitespace and comment errors.

Issue #440, opened 2023-11-17
Discrepancy in land masks


timspainNERSC:
Despite having the same data source, the restart file generation script and forcing generation script have a consistent, one pixel difference at the grid point [100, 95], the western tip of the New Siberian Islands.

PR #441, opened 2023-11-17
Explicitly correct the restart file land mask.

timspainNERSC:
# Explicitly correct the restart file land mask
## Fixes \#440

Despite having the same data source, the restart file generation script and forcing generation script have a consistent, one pixel difference at the grid point [100, 95], the western tip of the New Siberian Islands.

After much work trying to reconcile the forcing generation and restart file generation script, I do not seem able to get the two to agree, despite using the same data source files and, as best as I can see, the same code to interpolate between grids. Hence, I am explicitly patching the land mask in the restart file generation script. This removes the discrepancy and my headache.

Even if it is a bit of a hack.


MarionBWeinzierl (approved):
Should do what it is meant to do, but maybe slightly extend the comment to say why this fix is needed.

PR #447, opened 2023-11-23
Generate the data independent restart files at config time.

timspainNERSC:
# Generate the data independent restart files at config time.
## Fixes \#379

Add the CMake commands necessary to generate the restart files that do not depend on external data (`init_rect20x30.nc` and `init_rect30x30.nc`). These are generated at config time, not compile time, so CMake will have to be invoked whenever the generating scripts are changed. Since they are independent of the many code base, with no dependencies on it, config time seems a reasonable point at which to do this. The files can still be regenerated manually by running the appropriate script.


MarionBWeinzierl (commented):
Looks sensible to me

MarionBWeinzierl (approved):

TomMelt (approved):
Thanks for sorting this out Tim. Looks like this branch just needs rebasing to fix the conflict before it is merged.

timspainNERSC:
Needed to add the single column thermodynamics initial restart file.

PR #449, opened 2023-11-24
Issue347 nextsimlib

timspainNERSC:
# Pull Request Title

---
# Change Description

* Previous build of code base took ~45 mins (e.g. see [github actions](https://github.com/nextsimdg/nextsimdg/actions/runs/5855514762))
* @TomMelt  has modified the `CMakeLists.txt` files
* They should be a lot simpler to read and faster to compile.
* @TomMelt has removed build duplication by linking against a compiled lib rather than rebuilding every cpp file for each executable (there are many tests, each are separate exe's)

* In his first attempt @TomMelt has created one main library containing the majority of nextsimdg code (`nextsimlib`)
* This library could be split into modular libraries depending on use-case in future

* **@TomMelt's primary goal was to improve build time not necessarily thinking about how best to structure the code into libraries. This can be discussed at a later stage.**


TomMelt:
@MarionBWeinzierl this PR is based on https://github.com/nextsimhub/nextsimdg/pull/369. I basically modify the CMakeLists.txt to build nextsim as a library and then all the test binaries link against it, instead of recompiling. It would be best if we merge this before you start your dev branch. Let me know what you think and we can merge it in.

@timspainNERSC  also can you confirm that you are happy for it to be merged in now :)

Once I hear back from both of you, I will merge.

timspainNERSC:
A minor point:

> all the test binaries link against it

Not *all* the test binaries.

During rebasing, I resolved some of the conflicts by *not* linking to nextsimlib for a couple of the tests that only incorporate the test file and the source file from the code. Specifically, these are `testMonthlyCubicBSpline` and `testFileCallbackCloser`.

timspainNERSC:
Yes, I am happy for this to be merged. Particularly so, as it will then be in place when I talk to the folks in Grenoble next week.

MarionBWeinzierl (approved):
Works for me.

TomMelt:
merged. I will now start rebasing the mpi branches :+1:

Issue #439, opened 2023-11-14
Easier to use Modules


timspainNERSC:
After discussion with colleagues working on SASIP WP4, @einola and @timspainNERSC discussed making the process of adding a new implementation of a `Module` more intuitive and streamlined than it currently is.

The current process, once the user/dev has identified the `Module` they are targetting is:
* Write the new implementation source code, tests, &c.
  * This will remain the same in all cases: the code needs to be written and compiled.
* The source files are saved in `something/modules` and the header files in `something/modules/include`.
* Add the header to the `Module.cpp` source file as an `#include` directive.
* Optionally and preferably add a `const std::string` variable for the name of the new implementation.
* Add the mapping between the implementation name string and an instantiation of `newImpl<ModuleClass, ImplementationClass>
* Optionally and preferably add the implementation name to the list of available implementations in the map of help text.
* Add the Module source file name to the correct `set(ModuleSources` list in the relevant `CMakeLists.txt` file.
* Use the `Module` by the include directive `#include "include/ModuleNameModule.hpp`.
* Re-run CMake.
* Recompile.

The intended process will be more along the lines of:
* Write the new implementation source code, tests, &c.
* The source and header files are saved in `something/modules/ModuleName/`.
* Add the name of the implementation to the list `module.implementations` in the file `something/modules/ModuleName/module.cfg`.
* Add a new section, titled with the implementation name, to the file `something/modules/ModuleName/module.cfg`, listing
  * the file prefix,
  * an optional description,
  * whether the implementation class has help text supplied via a `getHelpRecursive(HelpMap& map, bool getAll)` function.
* Use the `Module` by the include directive `#include "include/ModuleNameModule.hpp`.
* Re-run CMake.
* Recompile.

The key difference is that the re-run CMake step will do some amount of parsing and code generation to create the Module file based on the values in the config file. The minimal addition to the config file is just the name in two places and a prefix for the file names of the implementation header and source files.

I believe this will be a less conceptually complex process than currently, and less prone to mistakes. The complexity is hidden in the python code generation script, which already substantially exists as `module_maker.py`.

The branch `feat_flexiblemodules` will be the location of developments on this issue.

PR #456, opened 2023-12-05
Flexible modules

timspainNERSC:
# Flexible Modules
## Fixes \#439

To make the modules easier to use, simplify the process of adding a new implementation. This also involves organizing the source and header files rather more neatly.

---
# Test

If all the tests that use modules (which is the majority) still pass, then all is working.

---
# Documentation Impact

Added `modules.rst` do `docs`.


timspainNERSC:
Replaced by #458

PR #458, opened 2023-12-12
Flexible modules

timspainNERSC:
# Flexible Modules
## Fixes \#439

To make the modules easier to use, simplify the process of adding a new implementation. This also involves organizing the source and header files rather more neatly.

---
# Test

If all the tests that use modules (which is the majority) still pass, then all is working.

---
# Documentation Impact

Added `modules.rst` to `docs`.


core/src/modules/DiagnosticOutputModule.cpp.bak
MarionBWeinzierl, line 1: Do we need this backup file or can it be deleted? Should be in the git history if it turns out it is still needed

timspainNERSC: It is definitely not needed and has been deleted.

MarionBWeinzierl (approved):
LGTM

Issue #463, opened 2023-12-19
Spherical coordinates for the dynamics


timspainNERSC:
Currently the DG dynamics in nextSIM assume Cartesian coordinates, and any spherical coordinates will be replaced with a 25 km Cartesian grid.

- [x] Check the operation of the dynamics with spherical coordinates.
- [x] Determine within nextSIM-DG if spherical or Cartesian coordinates should be used.

timspainNERSC:
> Check the operation of the dynamics with spherical coordinates.
They're bad. Really bad.

winzerle:
Hi Tim :-) 

Guess we need to discuss this in some detail. I am sure that it will require some work to make it really run in one of your test cases. There are many things to settle such as the system for angles (radians, degrees, etc.). But there will be much more. There are many decisions to be made when going to spherical coordinates and I choice between various options :-) We should zoom :-) 

Also regarding boundary. I am on it and there I have a good idea on how to integrate it.

Thomas


* --------------------------------------------
| Prof. Dr. Thomas Richter
| Numerische Mathematik in den Anwendungen
| Otto-von-Guericke Universität Magdeburg
| numerics.ovgu.de
| +49 391 67-57162  +49 176 615 612 26 (mobil)

PR #464, opened 2023-12-19
Remove the need for smesh files

timspainNERSC:
# Remove the need for smesh files
### Task List
The parameters of `ParametricMesh` can now be initialized from the data contained in a nextSIM-DG restart file. This removes the need of providing a `.smesh` file and indeed the ability to use one (the call to `readmesh(filename)` having been removed).

Note that until the spherical coordinates are fully implemented and checked (#463), a Cartesian grid is enforced.

---
# Test Description

The test `testParametricMesh` tests the conversion of data held in `ModelArray` form into `ParametricMesh` versus an equivalent `.smesh` file.  

---
# Documentation Impact

Any mention of needing a `.smesh` file has been removed. Description of the necessary fields for setting up the `ParametricMesh` information has been added.


dynamics/src/ParametricMesh.cpp
```diff
@@ -189,6 +189,80 @@ void ParametricMesh::readmesh(std::string fname)
...
+    landmask.resize(nelements);
+    for (size_t idx = 0; idx < mask.trueSize(); ++idx)
+    {
+        landmask[idx] = (mask[idx] == 1.);
+    }
+}
+
+void ParametricMesh::dirichletFromMask()
+{
+    // BOTTOM, RIGHT, TOP, LEFT
```
MarionBWeinzierl, line 217: Sorry, I am confused here: should it be bottom left, bottom right, lop left, top right?

dynamics/src/ParametricMesh.cpp
```diff
@@ -189,6 +189,80 @@ void ParametricMesh::readmesh(std::string fname)
     }
 }
```
MarionBWeinzierl, line 190: It would be useful to add some docstrings/function documentation.

dynamics/tests/FakeSmeshData.cpp
```diff
@@ -0,0 +1,192 @@
+/*!
+ * @file FakeSmeshData.cpp
+ *
+ * @date Dec 14, 2023
+ * @author Tim Spain <timothy.spain@nersc.no>
+ */
```
MarionBWeinzierl, line 6: What is this for? Documentation would be useful

dynamics/src/ParametricMesh.cpp
timspainNERSC: This part deals with edges, not vertices. Here I am noting Thomas' order, as much for my benefit as anyone else's.

I have expanded the comment.

dynamics/src/ParametricMesh.cpp
timspainNERSC: Copied the Doxygen documentation from the header file.

dynamics/tests/FakeSmeshData.cpp
timspainNERSC: A testing data source. I have added a brief description to each of the relevant files.

dynamics/src/ParametricMesh.cpp
MarionBWeinzierl: Aaah, thanks

dynamics/tests/FakeSmeshData.cpp
MarionBWeinzierl: Where?

MarionBWeinzierl: Ah, found it.

MarionBWeinzierl (approved):
I am ok with this as is, but maybe @TomMelt can have another look nevertheless

PR #471, opened 2024-01-04
A fix to the crash reported in issue 470

einola:
# A fix to the crash reported in issue 470
### Task List

---
# Change Description

The function Logged::configure() is trying to access keyMap, but this is not accessible. I propose fixing this by adding keyMap to Logged.hpp in the protected clause. I'm not sure if this is the right place or if it should be public or private. But the model runs now (at least the thermo test).

---
# Test Description

Run the thermodynamics integration test (demo), i.e.
```nextsim --config-file config_column.cfg```

Before the fix the model crashed with a sigsev message `(Process finished with exit code 139 (interrupted by signal 11: SIGSEGV))`. After the fix, the model runs through and produces the expected output: a netCDF file called `diagnostic.nc`, which produces the right plots when `column_pp.py` is run.

---
# Documentation Impact

N/A

---
# Other Details

N/A

---
### Pre-Request Checklist

### DIFF
--- a/core/src/Logged.cpp
+++ b/core/src/Logged.cpp
@@ -42,7 +42,7 @@ const std::map<std::string, Logged::level> Logged::levelNames = {
     { "NONE", level::NONE },
 };
 
-const std::map<int, std::string> keyMap = {
+const std::map<int, std::string> Logged::keyMap = {
     { Logged::MINIMUM_LOG_LEVEL_KEY, "Logged.minimum_log_level" },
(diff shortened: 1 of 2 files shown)


timspainNERSC:
The changes here look good, but I think the `Logged` class should inherit from `Configured`. Let me check if that also solves the problem. If it does, then we should use that solution instead.

timspainNERSC:
Also, the changes *look* good, but don't fix the SEGFAULT for me.

timspainNERSC:
Superseded by PR #472

Issue #473, opened 2024-01-04
Incorrect duration


einola:
The thermodynamics test crashes with the message

```libc++abi: terminating due to uncaught exception of type std::invalid_argument: The duration should be an ISO 8601 duration (P…) or a number of seconds. Got: ```

This seems to be related to PR #294, as rolling the model back to 55548256fa2f9bd1d4410e69c80f5d8be8865314 (before that PR) from 1eed3eee62025f190a4994286452245188e01e2d (current develop) the model runs fine.

The debugger was not very helpful, but I'm still investigating.

einola:
I think there's something wrong with how `string` and `istream` are handled in `Time.cpp`. I am still figuring out exactly how to fix it.

timspainNERSC:
The problem is with parsing an empty string, which is what the (absent) configuration defaults to.

PR #474, opened 2024-01-04
Duration defaults

timspainNERSC:
# Duration defaults
## Fixes \#473

With the introduction of the intermediate restart file period (issue #239), the default value from the configuration file was an empty string. This is not parsed correctly by `Duration::parse()`.

Change the configuration default to "0", a duration of 0 seconds. This will then result in a restart file being written every timestep. So if the Duration between restart files is set to zero, instead set it to 10 000 years. I think it is safe to assume that if someone is doing a multi-10 000 year run, then they will want restart files with a shorter period than that.

---
# Test Description

Added a test condition to ensure that Duration::parse("") throws.
The single column thermodynamics configuration is intended as a test, and that is once again working, so tests this.

---
# Documentation Impact

Added text about the default behaviour to the online help system for the relevant option.

### DIFF
--- a/core/src/DevStep.cpp
+++ b/core/src/DevStep.cpp
@@ -49,7 +49,6 @@ void DevStep::setRestartDetails(const Duration& restartPeriod, const std::string
 {
     /*
      * A restart period of zero means zero intermediate restart files.
-     * Use 10 000 years as the period.
      */
     m_restartPeriod = restartPeriod;
     m_restartFileName = fileName;
--- a/core/src/DevStep.cpp
+++ b/core/src/DevStep.cpp
@@ -8,16 +8,14 @@
 #include "include/DevStep.hpp"
 
 #include "include/ConfiguredModule.hpp"
-#include "include/DiagnosticOutputModule.hpp"
-
+#include "include/IDiagnosticOutput.hpp"
+#include "include/Module.hpp"
 namespace Nextsim {
-
-const Duration DevStep::myriadJulianYears = Duration(3.15576e11);
 
 DevStep::DevStep()
     : pData(nullptr)
     , mData(nullptr)
-    , m_restartPeriod(3.1536e11) // 10 000 years of 365 days
+    , m_restartPeriod(0)
 {
 }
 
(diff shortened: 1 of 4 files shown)


core/src/DevStep.cpp
```diff
@@ -12,6 +12,8 @@
 
 namespace Nextsim {
 
+const Duration DevStep::myriadJulianYears = Duration(3.15576e11);
```
einola, line 15: This seems  a bit arbitrary solution to me, even though it will work. Can we consider one of two alternatives?

1. Change line 39 of the modified DevStep.cpp to read ` if ((m_restartPeriod > 0) || (mData->time() >= lastOutput + m_restartPeriod)) {` and then a RESTARTPERIOD_KEY with a value 0 gives no restart files (and use this as default).
2. Change line 39 of the modified DevStep.cpp to read ` if ((m_restartPeriod >= 0) || (mData->time() >= lastOutput + m_restartPeriod)) {` and then a RESTARTPERIOD_KEY with a negative value gives no restart files (and use this as default).

einola (commented):
There is one minor discussion point; otherwise, all is good.

core/src/DevStep.cpp
timspainNERSC: Choice 1, except with AND `&&` rather than OR `||`

core/src/DevStep.cpp
```diff
@@ -47,6 +47,10 @@ void DevStep::iterate(const TimestepTime& tst)
 
 void DevStep::setRestartDetails(const Duration& restartPeriod, const std::string& fileName)
 {
+    /*
+     * A restart period of zero means zero intermediate restart files.
+     * Use 10 000 years as the period.
```
einola, line 52:
```suggestion
     * A restart period of zero means zero intermediate restart files.
```

core/src/Model.cpp
```diff
@@ -164,9 +164,10 @@ Model::HelpMap& Model::getHelpText(HelpMap& map, bool getAll)
             "Model physics timestep, formatted as an ISO8601 duration (P prefix). " },
         { keyMap.at(RESTARTFILE_KEY), ConfigType::STRING, {}, "", "",
             "The file path to the restart file to use for the run." },
-        { keyMap.at(RESTARTPERIOD_KEY), ConfigType::STRING, {}, "", "",
+        { keyMap.at(RESTARTPERIOD_KEY), ConfigType::STRING, {}, "0", "",
             "The period between restart file outputs, formatted as an ISO8601 duration (P prefix) "
-            "or number of seconds." },
+            "or number of seconds. A value of zero defaults to 10 000 years to ensure no "
+            "intermediate restart files are written." },
```
einola, line 170:
```suggestion
            "or number of seconds. A value of zero ensures no "
            "intermediate restart files are written." },
```

einola (approved):
Nice :) I've left suggestions to fix an out-of-date comment and an out-of-date output message. Other than that it's approved.

core/src/Model.cpp
timspainNERSC: Oops, yes of course. Done.

Issue #476, opened 2024-01-05
Crash at the end


einola:
I get a crash at the end (segfault) if I run cmake with build-type "Release", not when I compile with the debug build type. The model crashes both on the thermodynamics test case and on the june23 test. All expected outputs are produced, but the model doesn't print out the last timestep information to the console. Otherwise, everything seems to be fine, so it's probably in one destructor or another.

Since I can only reproduce the crash using "Release", then the debugger is not very helpful. It does, however, report bad access in `std::tree` as called by `Nextsim::Timer::~Timer()`. So that's a place to start.

einola:
Changing the destructor from

```
    virtual ~Timer() = default;
```

to 

```
    ~Timer() { reset(); }
```

does the trick. But I'm not sure if there are unwanted side-effects of doing this.

einola:
As a side note, I was wrong about the last timestep output on the console. I just made a mistake; everything's printed the same in debug and release mode.

einola:
A somewhat mysterious solution of removing `virtual` works. So we'll use that, even if we're not sure why it works.

PR #478, opened 2024-01-08
Fix the crash at the end (issue #476)

einola:
# Fix a crash at the end
### Task List

---
# Change Description

I fix the crash at the end by replacing the destructor of Timer (in Timer.hpp). This seems to work, but I wonder if there are side effects because there are two constructors, Timer() and Timer(const Key& rootKey), and because the destructor is no longer a virtual function.

---
# Test Description

Run cmake with CMAKE_BUILD_TYPE=Release, and compile the model from scratch. Run either the thermodynamics or june23 integration tests. These should run to the end, but end with a segmentation fault (signal 11).

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/core/src/include/Timer.hpp
+++ b/core/src/include/Timer.hpp
@@ -39,7 +39,7 @@ public:
      * @param rootKey Name of the root node.
      */
     Timer(const Key& rootKey);
-    virtual ~Timer() = default;
+    ~Timer() { reset(); }
 
     /*!
      * @brief Starts a named timer.
--- a/core/src/include/Timer.hpp
+++ b/core/src/include/Timer.hpp
@@ -39,7 +39,7 @@ public:
      * @param rootKey Name of the root node.
      */
     Timer(const Key& rootKey);
-    ~Timer() { reset(); }
+    ~Timer() = default;
 
     /*!
      * @brief Starts a named timer.


timspainNERSC:
Is it the lack of `virtual` or the `reset()` call that fixes the crash? Or are both changes necessary?

einola:
Good point! I thought it was the `reset()` call that fixed things, but it's in fact removing `virtual` that does it. I.e. 

```
    ~Timer() = default;
```

works, and so does

```
    ~Timer() { reset(); }
```

but both

```
    virtual ~Timer() { reset(); }
```

and

```
    virtual ~Timer() { reset(); }
```

give a crash at the end. Which also means that I don't understand what's going on. I thought that calling `reset()` would clear the memory so that we could have a clean exit.

timspainNERSC:
> Good point! I thought it was the `reset()` call that fixed things, but it's in fact removing `virtual` that does it.

That does suggest a deeper problem with how I am using the class. I'll take a look and see if I can see anything obviously wrong. If I can't find anything, then we can just push the non-`virtual` version.

core/src/include/Timer.hpp
```diff
@@ -39,7 +39,7 @@ class Timer {
      * @param rootKey Name of the root node.
      */
     Timer(const Key& rootKey);
-    virtual ~Timer() = default;
+    ~Timer() { reset(); }
```
einola, line 42:
```suggestion
    ~Timer() = default;
```

timspainNERSC:
> > Good point! I thought it was the `reset()` call that fixed things, but it's in fact removing `virtual` that does it.
> 
> That does suggest a deeper problem with how I am using the class. I'll take a look and see if I can see anything obviously wrong. If I can't find anything, then we can just push the non-`virtual` version.

Not a flippin´scooby.

timspainNERSC (approved):
Let's get this merged.

Issue #480, opened 2024-01-17
NcOpen exceptions do not include the file name


timspainNERSC:
If an ncOpen exception is thrown it is difficult to know which file was not found, as the only file name is the netCDF source file from which the exception was thrown.

timspainNERSC:
PR #481 was merged

PR #481, opened 2024-01-17
Wrap the ncOpen calls to append the file name to the exception message.

timspainNERSC:
# Wrap the ncOpen calls to append the file name to the exception message.
## Fixes \#480

---

Really only does what it says in the title.

### DIFF
--- a/core/src/ParaGridIO.cpp
+++ b/core/src/ParaGridIO.cpp
@@ -14,12 +14,14 @@
 #include "include/gridNames.hpp"
 
 #include <ncDim.h>
+#include <ncException.h>
 #include <ncFile.h>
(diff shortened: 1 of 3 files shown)


MarionBWeinzierl (approved):
Looks good to me. Pedantically, I could ask for the catch part to be put into a function to avoid code duplication, but realistically, that would probably require more code and just save two duplicated lines...

timspainNERSC:
> Pedantically, I could ask for the catch part to be put into a function to avoid code duplication, but realistically, that would probably require more code and just save two duplicated lines...

I had exactly that thought and came to the same conclusion. Possibly with a frisson more laziness.

PR #484, opened 2024-01-22
add support for virtual environments to CMake

TomMelt:
Currently `cmake` will just search for the system `python`.

It does not find python inside of a virtual environment (`conda` or `venv`)

I use `conda` and therefore my `python` distribution is in a virtual environment.

These modifications to the `CMakeLists.txt` should make the process more general, whilst still working with the system `python`.


MarionBWeinzierl (approved):
Looks good to me. Still runs with system Python. Couldn't test the Mac version, though

TomMelt:
@timspainNERSC  if you can confirm that it runs on your mac then I will merge this in. I just rebased to develop and it works with the `run_simple_example.sh`.

timspainNERSC:
> @timspainNERSC if you can confirm that it runs on your mac then I will merge this in. I just rebased to develop and it works with the `run_simple_example.sh`.

Works for me!

Issue #485, opened 2024-01-22
run_simple_example is failing


TomMelt:
When trying to run the simple example from the latest develop branch I get the following error:

```
$ ./run_simple_example.sh 
cleared m_configHelp
terminate called after throwing an instance of 'netCDF::exceptions::NcBadId'
  what():  NetCDF: Not a valid ID
file: ncType.cpp  line:138
Aborted (core dumped)
```

The problem seems to be that nextsim is looking for a variable called `"longitude"` in file `init_rect30x30.nc`. But this variable does not exist in the netcdf file `init_rect30x30.nc`.

```
netcdf init_rect30x30 {

group: structure {
[...]
  } // group metadata

group: data {
  dimensions:
  	x = 30 ;
  	y = 30 ;
  	nLayers = 1 ;
  variables:
  	double mask(y, x) ;
  	double cice(y, x) ;
  		cice:missing_value = -2.03703597633449e+90 ;
  	double hice(y, x) ;
  		hice:missing_value = -2.03703597633449e+90 ;
  	double hsnow(y, x) ;
  		hsnow:missing_value = -2.03703597633449e+90 ;
  	double tice(nLayers, y, x) ;
  		tice:missing_value = -2.03703597633449e+90 ;
  data:
[...]
```

The error is triggered when it tries to read the variable from the netcdf file in file `core/src/RectGridIO.cpp`:

https://github.com/nextsimhub/nextsimdg/blob/cdd7e6b5b1631d3aaa7340692a1cbb5364dd2839/core/src/RectGridIO.cpp#L141

PR #486, opened 2024-01-23
Coordinates for the 'simple rectangular' test

timspainNERSC:
# Coordinates for the 'simple rectangular' test
## Fixes \#485
---
# Change Description

Add coordinates to the `init_rect30x30.nc` restart file so that it is a valid restart file once again. The coordinates are generated in the `make_init.py` Python script along with the rest of the file. They consist of an `x` and a `y` array for element centres and a `coords` array containing the _x_ and _y_ coordinates of the vertices. The origin is the bottom left vertex of element (0, 0) so all coordinates are non-negative. The grid spacing is 150 km.

Also remove an errant debug print statement from the command line parser.

---
# Test Description

Running the `run_simple_example.sh` script works once again.

### DIFF
--- a/core/src/CommandLineParser.cpp
+++ b/core/src/CommandLineParser.cpp
@@ -63,7 +63,6 @@ CommandLineParser::CommandLineParser(int argc, char* argv[])
     }
 
     m_configHelp.clear();
-    std::cerr << "cleared m_configHelp" << std::endl;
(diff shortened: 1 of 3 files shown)


TomMelt (approved):
Thanks @timspainNERSC . This fixes issue #485 and I tested it on my machine as well. Looks like it produces the correct output (diagnostic and restart netcdf files)

Issue #378, opened 2023-08-31
The BBM rheology exists, but is not usable


timspainNERSC:
Within the dynamics of the model, the BBM rheology exists (`dynamics/src/include/bbm.hpp`). It is not currently usable as the dynamics are hard-coded to use MEVP. Implement an additional class derived from `IDynamics` that which uses the BBM update function and provides all the other data and configuration required to enable this.

timspainNERSC:
Notes from a morning's investigation:
Three functions exist in `cgParametricMomentum.cpp`:
* `mEVPStep`
* `MEBStep`
* `BBMStep`
This suggests that the switching can take place with the `DynamicsKernel` class.

There is a lot of overlap between the three functions, and they would fir quite well into an object oriented framework.

The parameter sets, stored in the classes `VPParameters` and `MEBParameters` are constant per-model run and are also amenable to being turned into a class hierarchy (of two).

The function `prepareIteration` exists in versions with and without a damage argument. These would probably be better described as mEVP and non-mEVP versions, as the contained functionality is different, depending on whether the damage argument is passed.

timspainNERSC:
The MEB and BBM rheologies should be merged as far as possible, since they are identical with the exception of a single term, P tilde.

If the refactoring can be performed in such a way to abstract the mesh dependent operation from the rest, that would be an advantage for including other meshes. This would be anything in the code that performs a horizontal derivative. This is likely limited to the advection and the calculation of the rate of strain tensor from the horizontal velocity gradients. The rest of the rheology is performed on a per-point basis and could (likely) be the same between a structured grid and an unstructured mesh.

Is there any scope to support semi-implicit solution of the momentum equation? Currently all schemes are explicitly solved.

einola:
We abandoned the semi-implicit approach, as it's too costly for a large-scale model. Re-introducing it would be a significant development, and I don't have any good ideas on how it can be done cost-efficiently.

timspainNERSC:
Current design as of end January 2024 (potentially subject to much change):

The dynamics are called from a module that is referenced in `PrognosticData`, with interface type `IDynamics`. This will have four implementation classes, corresponding to the three current rheologies plus no dynamics:

- `DummyDynamics`
- `MEVPDynamics`
- `MEBDynamics`
- `BBMDynamics`.

[...]

PR #487, opened 2024-01-23
Rename bare class name to include MEVP

timspainNERSC:
# Rename bare class name to include MEVP
## Fixes \#378

---
# Change Description

Part of implementing the dynamics switcher is renaming the 'Dynamics' classes to include MEVP in their name. Several different issue branches related to the dynamics will touch these files, so it will be easier to have the renaming changes included in develop for basing the branches off.

---
# Test Description

N/A, only renames existing classes.


MarionBWeinzierl (approved):
Just a rename of a class, cannot spot any problem

PR #489, opened 2024-01-24
Add Dockerfiles and document usage

### LINKED ISSUE #482: Add Dockerfile

MarionBWeinzierl:
This PR fixes #482 . We are
- [x] Adding Dockerfiles for serial and parallel build.
- [x] Updating the documentation to explain their usage.
- [x] Considering the use of [Docker Hub](https://hub.docker.com/) to deploy the Docker image.

Consideration about the style and usage of the Dockerfile can be found in the discussion on #482.


MarionBWeinzierl:
I have added some documentation as well. Regarding the use of Docker Hub: I think we agree that we will not do it right now, but might consider it in future.

MarionBWeinzierl:
@einola @timspainNERSC The Dockerfiles at the moment still copy over the code and do not check the repository out. This is what I explain in the docs, too.

Would you prefer me to change this and have them checkout the code instead?

einola:
That's ok, and we can leave it like this for now. I would very much like inputs from people actually using the docker files on how they use them - but that's step 2 ;)

Dockerfiles/Dockerfile_Ubuntu
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update
+RUN apt-get install -y build-essential
+RUN apt-get install -y netcdf-bin libnetcdf-c++4-dev libboost-all-dev libeigen3-dev cmake
+RUN apt-get install -y python3-netcdf4
+
+WORKDIR /home/nextsim
+COPY . /home/nextsim
+
+RUN rm -r build
```
TomMelt, line 11: you might need to include `-f` flag here because on a clean clone of the github repo build would not already exist.

```suggestion
RUN rm -rf build
```

Dockerfiles/Dockerfile_Ubuntu
```diff
@@ -0,0 +1,20 @@
...
+RUN mkdir -p /home/nextsim/build 
+WORKDIR /home/nextsim/build 
+RUN cmake -B/home/nextsim/build  -H/home/nextsim/ >&2
```
TomMelt, line 14: why do you redirect output to `stderr` i.e., `>&2`?

Dockerfiles/Dockerfile_Ubuntu_MPI
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:23.10 
+
+RUN apt-get update
+RUN apt-get install -y build-essential
+RUN apt-get install -y netcdf-bin libnetcdf-c++4-dev libboost-all-dev libeigen3-dev cmake
+RUN apt-get install -y python3-netcdf4
+RUN apt-get install -y libnetcdf-mpi-dev 
+RUN apt-get install -y openmpi-bin libopenmpi-dev 
+
+ENV CPLUS_INCLUDE_PATH /usr/lib/x86_64-linux-gnu/netcdf/mpi/include
```
TomMelt, line 10: does this still work? I know we took it out of your `.bashrc`. Does it cause any issues in the docker instance?

Dockerfiles/Dockerfile_Ubuntu_MPI
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR /home/nextsim
+COPY . /home/nextsim
+
+RUN rm -r build
```
TomMelt, line 15:
```suggestion
RUN rm -rf build
```

TomMelt (changes requested):
looks good to me. Just a couple of questions

Dockerfiles/Dockerfile_Ubuntu
MarionBWeinzierl: Because I like to see what happens, mainly for debugging purposes/to make sure everything works as expected. I can delete that if you think it is not good style.

Dockerfiles/Dockerfile_Ubuntu_MPI
MarionBWeinzierl: I just tested it. If I remove the line I am back to `nextsim/core/src/ParallelNetcdfFile.cpp:3:10: fatal error: netcdf_par.h: No such file or directory`, with that line it works.

Dockerfiles/Dockerfile_Ubuntu
MarionBWeinzierl: ok, I removed that in both files

TomMelt (approved):
thanks for the changes. Looks good to me.

PR #492, opened 2024-01-30
Add damage and switch to BBM.

timspainNERSC:
Just want to see the changes.


timspainNERSC:
Dynamics switcher code has been merged into `develop` and does indeed contain both damage and BBM.

PR #494, opened 2024-02-05
Add spack support

TomMelt:
Add `spack.yaml` to project. This will allow `spack` users to easily install dependencies.

I have also added documentation to explain how to:
* install `spack`
* install the dependencies into a `spack` environment
* use `spack` environments

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -62,6 +62,47 @@ Install conda via anaconda or miniconda (no root privileges required)
         conda -c conda-forge boost
         conda -c anaconda cmake
 
+**Installing dependencies via spack**
+
+``spack`` is a package manager which was designed for HPC software. It is language agnostic and
+compiles binaries from source. See quote from ``spack``'s `documentation
+<https://spack.readthedocs.io/en/latest/>`_ below:
+
+   Spack is a package management tool designed to support multiple versions and configurations of
+   software on a wide variety of platforms and environments. It was designed for large supercomputing
+   centers, where many users and application teams share common installations of software on clusters
+   with exotic architectures, using libraries that do not have a standard ABI. Spack is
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -62,6 +62,47 @@ Install conda via anaconda or miniconda (no root privileges required)
...
 
+**Installing dependencies via spack**
+
+``spack`` is a package manager which was designed for HPC software. It is language agnostic and
+compiles binaries from source. See quote from ``spack``'s `documentation
+<https://spack.readthedocs.io/en/latest/>`_ below:
+
+   Spack is a package management tool designed to support multiple versions and configurations of
+   software on a wide variety of platforms and environments. It was designed for large supercomputing
+   centers, where many users and application teams share common installations of software on clusters
```
MarionBWeinzierl, line 73: Not sure whether this quote is encouraging or leads users to believe they don't need spack on their personal computer. Maybe leave out the "It was designed..." sentence would make it sound less scary?

docs/installation.rst
```diff
@@ -62,6 +62,47 @@ Install conda via anaconda or miniconda (no root privileges required)
...
+all required dependencies. The following command will create a ``spack`` environment (called
+``nextsim`` in this example) and install the dependencies automatically:
+
+.. code::
+
+   spack env create nextsim spack.yaml
+
+.. note::
+
+   This will install by building from source. This can take a significant amount of time, depending
```
MarionBWeinzierl, line 91: Running the `env create` command does not start the install, I think. You need to activate the environment, and then run `spack install` for the install to actually start.

TomMelt: Thanks for spotting this. I have updated the instructions :ok_hand:

MarionBWeinzierl (approved):
This looks good to me now.

Issue #503, opened 2024-02-22
Only fake up coordinates for spherical grids


timspainNERSC:
I tried to add a fix for spherical grids not working on the dynamics, but managed to squash the coordinates for both types of grid.

timspainNERSC:
The temporary fix only fixes what it needs to.

PR #504, opened 2024-02-22
Only override the restart coords in case of a spherical grid.

timspainNERSC:
# Pull Request Title
## Fixes \#503

Makes the coordinates work again for non-spherical grids.

Until issue #463 is complete, the dynamics do not work with spherical coordinates. In order to be able to continue using the `init_25kn_NH.nc` restart file, the coordinates are replaced in the initialization of the `DynamicsKernel` class with the Cartesian coordinates with a 25 km spacing that the corresponding `.smesh` file used to provide. However, when running from a file containing a Cartesian grid, the coordinates should not be replaced. This PR ensures that only `SPHERICAL` grids have their coordinates replaced by the 25 km grid.

### DIFF
--- a/core/src/modules/DynamicsModule/MEVPDynamics.cpp
+++ b/core/src/modules/DynamicsModule/MEVPDynamics.cpp
@@ -43,21 +43,24 @@ void MEVPDynamics::setData(const ModelState::DataMap& ms)
     }
 
     // TODO: Remove this when spherical coordinates are fully implemented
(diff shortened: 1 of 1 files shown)


MarionBWeinzierl:
I do not really understand what is happening here or why it is needed, could you please explain?

MarionBWeinzierl (approved):
Looks good to me

PR #507, opened 2024-03-08
Docker ci build

TomMelt:
## Nextsim CI \*\*now with added Docker\*\*

This PR mainly pertains to the ubuntu build of the code. 

Previously the CI workflow built `nextsim` using a GitHub VM instance that would `apt install` basic dependencies required for the build process.

However, with the addition of `xios` (#496 ) this is no longer a tractable solution. Furthermore, it would take ~15 mins just to build `xios` before then going on to build `nextsim`.

Now I have created a `Dockerfile` and associated pre-built image (currently on my Dockerhub: `tommelt/nextsimdg-dev-env:with-xios`) which contains all the prerequisites to build and test the code for our typical CI workflow.

In theory it could also be repurposed by developers as a dev environment (although I personally don't love this).

### more about the dockerfile

The `Dockerfile` is based on a `spack` environment file (`spack.yaml`) was generated originally using the following command:

```
spack containerize > Dockerfile
```

However, I needed to make some manual adjustments to the `Dockerfile` (necessary for installing `xios` and some other dependencies).

`Dockerfiles/Dockerfile` is the one that was used to generate the docker image `tommelt/nextsimdg-dev-env:with-xios`.

The image is `2.41GB` and contains all prerequisites for `MPI` and `XIOS`. When uploaded to dockerhub (`docker push tommelt/nextsimdg-dev-env:with-xios`) it is compressed to `625.97 MB`.

### benefits of Docker in CI

* It's fast
  - only takes 45s to download the image and then 150s to build `nextsim` :fire:  
  - a full CI (for ubuntu) is now less than 4 minutes -- building `XIOS` alone used to take ~15 minutes 
* It's reproducible
  - one of the problems of github CI is it's hard to debug when things don't work
  - you can now pull the docker image (`docker pull tommelt/nextsimdg-dev-env:with-xios`) and test stuff locally
* It's easy
  - `nextsim` is necessarily adding complexity, which increases the required dependencies
  - the docker image can be built with all dependencies ready-to-go

### To Do:

- [ ] move pre-built image to nextsim dockerhub repo
- [ ] **discuss**: should we keep existing docker images? or "rebase" them? or just have one with all dependencies and users can build their own version of nextsim to meet their needs?
- [ ] add documentation around new docker image


docs/installation.rst
```diff
@@ -123,38 +123,75 @@ For example, on Debian-based Linux we need to also do:
...
+
+.. code-block:: console
+
+   docker pull ghcr.io/nextsimhub/nextsimdg-dev-env:latest
+
+Production Dockerfile
+^^^^^^^^^^^^^^^^^^^^^
+
+The production image is not stored on the nextsimhub `GitHub container registry
+<https://github.com/orgs/nextsimhub/packages>`_ because it is not needed for the CI. Users of the
```
jwallwork23, line 181: But might it be worth testing the production image every once in a while? Otherwise it could break without us noticing. To reduce doing this unnecessarily often, it could be done using a scheduled GitHub Actions job, perhaps once a week or month.

[I might be overthinking this as I see that the production image is based off the development environment and doesn't change much other than configuration options. Thought it would be worth mentioning, though.]

docs/installation.rst
```diff
@@ -123,38 +123,75 @@ For example, on Debian-based Linux we need to also do:
...
-
-With the last command, you are inside your running container on the bash and can use your usual command line commands to look around and run the code.
-    
-    
-.. _NetCDF: https://www.unidata.ucar.edu/software/netcdf/
-.. _Boost: https://www.boost.org/
-.. _CMake: https://cmake.org/
-.. _Homebrew: https://brew.sh/
-.. _Docker: https://www.docker.com/
+For a full list of options, please see ``Dockerfile.production``. By default ``MPI`` and ``xios``
```
jwallwork23, line 196: It would be nice to have the ``Dockerfile.production`` here be a hyperlink to the relevant file on GitHub, especially since it says "see" this file.

.github/workflows/clang-compile-tests.yml
```diff
@@ -32,36 +32,41 @@ jobs:
...
             # run serial tests
-            for file in $(find test* -maxdepth 0 -type f | grep -vP "_MPI\d*$"); do echo $file; ./$file; done
-            # run MPI tests that require 2 ranks
-            for file in $(find test* -maxdepth 0 -type f | grep -P "_MPI2$"); do echo $file; mpirun -n 2 --oversubscribe ./$file; done
+            for file in $(find test* -maxdepth 0 -type f | grep -vP "_MPI\d*$")
+            do
+              echo $file
+              ./$file
+            done
+            # run MPI tests
```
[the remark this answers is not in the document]
TomMelt: I chose not to split because we can reuse more of the code that way. I don't have a strong preference though. I thought this was just easier to maintain. As for the download it should take about 1s.

docs/installation.rst
TomMelt: Yeah, we do not want to support this at present.

The idea of production release is more for users that want a simpler installation of nextsim for testing. The name is perhaps (currently) misleading because I would not recommend using it on HPC because it will not use the system MPI (amongst other core deps) and this could affect performance. But it might be useful for other scientists who want to try/test new features e.g., XIOS and MPI.

.github/workflows/clang-compile-tests.yml
```diff
@@ -74,49 +79,36 @@ jobs:
...
-            cd -
-        done
-    - name: make
-      run: |
-        cmake . -DPython_ROOT_DIR="/Library/Frameworks/Python.framework/Versions/Current/bin/"
-        make
+        # added Python_ROOT_DIR to help cmake find correct Python in the mac VM
+        cmake -DENABLE_MPI=OFF -DENABLE_XIOS=OFF -DPython_ROOT_DIR=/Library/Frameworks/Python.framework/Versions/Current/bin ..
+        make -j 4
     - name: run tests
```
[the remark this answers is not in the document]
TomMelt: as mentioned in the above reply I do not want to split at present because it will lead to some duplication.

.github/workflows/clang-compile-tests.yml
```diff
@@ -74,49 +79,36 @@ jobs:
...
+        pip install numpy
+        pip install netCDF4
 
-        svn checkout http://forge.ipsl.jussieu.fr/ioserver/svn/XIOS/trunk xios
-        cp .github/xios_arch/ubuntu.arch xios/arch/arch-GCC_LINUX.path
-        cp .github/xios_arch/ubuntu.env xios/arch/arch-GCC_LINUX.env
-        (cd xios && ./make_xios --arch GCC_LINUX --netcdf_lib netcdf4_seq)
+    # uncomment these lines to debug mac build
+    # - name: Setup tmate session
+    #   uses: mxschmitt/action-tmate@v3
```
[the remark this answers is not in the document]
TomMelt: commit [efe9ea5](https://github.com/nextsimhub/nextsimdg/pull/507/commits/efe9ea5918aa925f7aff2b4701dc1cc3815d7d56) fixes this.

docs/installation.rst
TomMelt: I had a thought about this. I can't think of a sensible way to do this. If it was a `README.md` I would link to the relative path. Because it's external I would need a URL. But if I do that it will have to link to a specific branch/verison/commit hash. But then it would be out of date after any commit/PR that changes the file. Am I missing something?

jwallwork23: Yeah I suppose you're right, especially if it would have to go via `develop` before `main`.

.github/workflows/clang-compile-tests.yml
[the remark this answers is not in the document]
jwallwork23: [Leaving thread unresolved because in today's meeting we decided to have separate workflows for MPI and non-MPI tests.]

timspainNERSC (approved):

PR #509, opened 2024-03-11
Require only Python interpreter

draenog:
# Require only Python interpreter
## Change Description

* Require only Python interpreter in CMakeLists.txt.


MarionBWeinzierl (approved):
Makes sense.

Issue #510, opened 2024-03-11
FluxConfiguredAtmosphere is missing a key


timspainNERSC:
Using FluxConfiguredAtmosphere will fail with a key error, as the evaporation key is not defined correctly.

timspainNERSC:
Fixed and merged.

PR #511, opened 2024-03-11
Add the evap key.

timspainNERSC:
# Add the evap key
## Fixes \#510

---
# Change Description

Add the evaporation key to FluxConfiguredAtmosphere.
---
# Documentation Impact

Added the key to the config help text of `FluxConfiguredAtmosphere`. Also corrected the units of the sublimation flux, which was (apparently) a heat flux, but should be a mass flux.


MarionBWeinzierl (approved):
Simple changes, looks sensible to me as far as I can tell.

PR #514, opened 2024-03-12
Issue513 rect grid test improvements

### LINKED ISSUE #513: Improvements to MPI RectGrid_test
Currently MPI test in RectGrid test runs on only one MPI rank. Maybe it should use the following pattern:

1. Run test file in on 1 MPI rank with known data.
2. Read the same file in parallel and compare that the data are the same.
3. Write the same file in parallel so it can be compared with the file written serially.

draenog:
# Improvements to RectGrid parallelisation
## Fixes #513 

The pull requests implements the following workload in RectGrid_test:
1. Runs test file in on 1 MPI rank with known data.
2. Reads the same file in parallel and compare that the data are the same.
3. Writes the same file in parallel so it can be compared with the file written serially.

The resulting field is still different  as currently `RectGridIO::getModelState` does not read  `sst` and `sss` components
so they are missing form the resulting file.

Additionally the pull requests:
1. Tests local dimensions in parallel case
2. Introduce tests for ModelArray::Type::Z dimensions
3. Fixes reading of 3D data in paralle mode
4. Fix setting dimensions in parallel case
5. Tests local dimensions in parallel case 
6. Introduce tests for ModelArray::Type::Z dimensions


draenog:
@timspainNERSC Could you maybe comment on this:

> The resulting field is still different as currently RectGridIO::getModelState does not read `sst` and `sss` components
so they are missing form the resulting file.

Is it on purpose or should RectGridIO::getModelState read `ssst` and `sss`data?

timspainNERSC:
> @timspainNERSC Could you maybe comment on this:
> 
> > The resulting field is still different as currently RectGridIO::getModelState does not read `sst` and `sss` components
> > so they are missing form the resulting file.
> 
> Is it on purpose or should RectGridIO::getModelState read `ssst` and `sss`data?

When I wrote RectGridIO, we hadn't looked at the slab ocean, which is what uses `sss` and `sst`. These were added to ParaGridIO when I added the slab ocean, but I never went back and added them to RectGridIO. Since RectGrid is not intended to be used in the production model (it doesn't support the dynamics at all), I'm not planning on going back and adding support. Unless told otherwise, of course.

draenog:
> > @timspainNERSC Could you maybe comment on this:
> > > The resulting field is still different as currently RectGridIO::getModelState does not read `sst` and `sss` components
> > > so they are missing form the resulting file.
> > 
> > 
> > Is it on purpose or should RectGridIO::getModelState read `ssst` and `sss`data?
> 
> When I wrote RectGridIO, we hadn't looked at the slab ocean, which is what uses `sss` and `sst`. These were added to ParaGridIO when I added the slab ocean, but I never went back and added them to RectGridIO. Since RectGrid is not intended to be used in the production model (it doesn't support the dynamics at all), I'm not planning on going back and adding support. Unless told otherwise, of course.

In this case I wonder if we could  be worth removing these two field from ModelState in `RectGrid_test.cpp`. That way is the state is written, read and written again it should produce exactly the same file which is currently not the case.

core/test/RectGrid_test.cpp
```diff
@@ -104,47 +105,95 @@ TEST_CASE("Write and read a ModelState-based RectGrid restart file")
...
     REQUIRE(ModelArray::dimensions(ModelArray::Type::H)[1] == ny);
+    REQUIRE(ModelArray::dimensions(ModelArray::Type::Z)[0] == nx);
+    REQUIRE(ModelArray::dimensions(ModelArray::Type::Z)[1] == ny);
+#endif
     REQUIRE(ms.data.at("hice")(targetX, targetY) != 0);
     REQUIRE(ms.data.at("hice")(targetX, targetY) > 1);
     REQUIRE(ms.data.at("hice")(targetX, targetY) < 2);
-    REQUIRE(ms.data.at("hice")(targetX, targetY) == 1.0703);
+#ifdef USE_MPI
+    REQUIRE(ms.data.at("hice")(targetX, targetY) == 1.0201 + metadataIn.localCornerY * 0.01 + metadataIn.localCornerX * 0.0001);
```
MarionBWeinzierl, line 167: Where do these numbers come from?

MarionBWeinzierl (approved):
This looks generally good to me. I just have some questions to help me better understand the changes.

MarionBWeinzierl (commented):
This looks good to me. But could you just briefly clarify the relationship between nDims possibly being larger then 2, and the partition file and bounding box etc only having x and y dimensions?

core/test/RectGrid_test.cpp
draenog: Good catch. They should be `xFactor` and `yFactor. I will fix it.

core/test/RectGrid_test.cpp
```diff
@@ -39,8 +40,8 @@ TEST_CASE("Write and read a ModelState-based RectGrid restart file")
     grid.setIO(new RectGridIO(grid));
 
     // Fill in the data. It is not real data.
-    size_t nx = 25;
-    size_t ny = 15;
+    size_t nx = 5;
```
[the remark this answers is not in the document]
draenog: Yes, when debugging I often look at output of `ncdump`. It is easier for me to comprehend it with smaller dimensions  and the larger sizes do not bring anything here in my opinion. Size  5x7 is small enough to be easy to look at, and large enough to provide 3 blocks with non flat topology and not equal sizes in decomposition.

core/test/RectGrid_test.cpp
```diff
@@ -104,47 +105,95 @@ TEST_CASE("Write and read a ModelState-based RectGrid restart file")
...
+// Create subcommunicator with only first rank
     metadata.setMpiMetadata(test_comm);
-    metadata.globalExtentX = nx;
-    metadata.globalExtentY = ny;
-    metadata.localCornerX = 0;
-    metadata.localCornerY = 0;
-    metadata.localExtentX = nx;
-    metadata.localExtentY = ny;
-#endif
+    int colour = MPI_UNDEFINED, key = 0;
```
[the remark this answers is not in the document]
draenog: It is a term used by MPI standard when defining `MPI_Comm_split` API:

> This function partitions the group associated with comm into disjoint subgroups, one for each value of color. Each subgroup contains all processes of the same color. 

There is a comment above that this part of code creates an MPI subcommunicator which contains only the first rank.

draenog:
> This looks good to me. But could you just briefly clarify the relationship between nDims possibly being larger then 2, and the partition file and bounding box etc only having x and y dimensions?

This comes from the design of parallelization that the domain only in x and y directions. I am not sure if the test is a good place to document it.

MarionBWeinzierl:
> > > @timspainNERSC Could you maybe comment on this:
> > > > The resulting field is still different as currently RectGridIO::getModelState does not read `sst` and `sss` components
> > > > so they are missing form the resulting file.
> > > 
> > > 
> > > Is it on purpose or should RectGridIO::getModelState read `ssst` and `sss`data?
> > 
> > 
> > When I wrote RectGridIO, we hadn't looked at the slab ocean, which is what uses `sss` and `sst`. These were added to ParaGridIO when I added the slab ocean, but I never went back and added them to RectGridIO. Since RectGrid is not intended to be used in the production model (it doesn't support the dynamics at all), I'm not planning on going back and adding support. Unless told otherwise, of course.
> 
> In this case I wonder if it would be worth removing these two field from ModelState in `RectGrid_test.cpp`. That way if the state is written, read and written again it should produce exactly the same file which is currently not the case.

@timspainNERSC , would that be ok?

timspainNERSC:
> > > > @timspainNERSC Could you maybe comment on this:
> > > > > The resulting field is still different as currently RectGridIO::getModelState does not read `sst` and `sss` components
> > > > > so they are missing form the resulting file.
> > > > 
> > > > 
> > > > Is it on purpose or should RectGridIO::getModelState read `ssst` and `sss`data?
> > > 
> > > 
> > > When I wrote RectGridIO, we hadn't looked at the slab ocean, which is what uses `sss` and `sst`. These were added to ParaGridIO when I added the slab ocean, but I never went back and added them to RectGridIO. Since RectGrid is not intended to be used in the production model (it doesn't support the dynamics at all), I'm not planning on going back and adding support. Unless told otherwise, of course.
> > 
> > 
> > In this case I wonder if it would be worth removing these two field from ModelState in `RectGrid_test.cpp`. That way if the state is written, read and written again it should produce exactly the same file which is currently not the case.
> 
> @timspainNERSC , would that be ok?

Yes, that sounds fine.

draenog:
I've just pushed  the updated branch. The changes:

1. Rebases on develop commit 1423515cd293a978de5d050dfafadc91756e50e0 to fix CI tests on Mac OS.
2. Remove sss and sst fields from ModelState as they are not read by ParaGridIO
3. Remove hardcoded offsets in test spotted by @MarionBWeinzierl

core/test/RectGrid_test.cpp
```diff
@@ -24,13 +24,14 @@
 #include <fstream>
 
 const std::string filename = "RectGrid_test.nc";
-const std::string partition_filename = "partition_metadata_1.nc";
+const std::string partition_filename = "partition_metadata_3.nc";
```
TomMelt, line 27: could you please send this partition file (`partition_metadata_3.nc`) to @timspainNERSC  so that he can upload it to the ftp server?

That way I can add it to the CI workflow.

My PR #507 `wget`s the required `.nc` files and will run the test as part of CI.

core/test/RectGrid_test.cpp
```diff
@@ -104,47 +103,95 @@ TEST_CASE("Write and read a ModelState-based RectGrid restart file")
...
-    ModelState ms = gridIn.getModelState(filename, partition_filename, metadata);
+    ModelMetadata metadataIn;
+    metadataIn.setTime(TimePoint(date_string));
+    metadataIn.setMpiMetadata(test_comm);
+    ModelState ms = gridIn.getModelState(filename, partition_filename, metadataIn);
 #else
     ModelState ms = gridIn.getModelState(filename);
 #endif
 
+#ifdef USE_MPI
```
TomMelt, line 150: could you add a couple more comments to this test file (`core/test/RectGrid_test.cpp`) please?

TomMelt (approved):
Awesome thanks for this @draenog , this really improves the test and improves are likelihood of spotting bugs sooner. 

I would like a few more comments in the test, but otherwise I am really happy with this PR.

Thanks again :+1:

core/test/RectGrid_test.cpp
draenog: Instead I've implemented generation of `partition_metadata_3.nc` file from its text dump during the building process.

core/test/RectGrid_test.cpp
draenog: I've added some more comments as requested.

PR #516, opened 2024-03-12
Issue515 domain decompose refactor

### LINKED ISSUE #515: Move logic related to partition file  to ModelMetadata
Currently reading and setting metadata related to MPI domain decomposition is done inside RectGrid classes. However the same functionality will be also needed in order to parallelise   ParaGrid . So maybe the relevant code should be moved to ModelMetadata classes so it can be reused.

draenog:
# Move domain decompose code to ModelMetadata

### DIFF
--- a/core/src/Model.cpp
+++ b/core/src/Model.cpp
@@ -119,11 +119,11 @@ void Model::configure()
 #ifdef USE_MPI
     std::string partitionFile
         = Configured::getConfiguration(keyMap.at(PARTITIONFILE_KEY), std::string("partition.nc"));
+    m_etadata.getPartitionMetadata(partitionFile);
 #endif
 
 #ifdef USE_MPI
-    ModelState initialState(
-        StructureFactory::stateFromFile(initialFileName, partitionFile, m_etadata));
+    ModelState initialState(StructureFactory::stateFromFile(initialFileName, m_etadata));
 #else
     ModelState initialState(StructureFactory::stateFromFile(initialFileName));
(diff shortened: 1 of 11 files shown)


draenog:
It should go on top of #514.

MarionBWeinzierl (approved):
This looks relatively straightforward to me: the mpi code is moved to ModelMetadata, and the partition file is removed from the function calls.

core/src/Model.cpp
```diff
@@ -119,11 +119,11 @@ void Model::configure()
 #ifdef USE_MPI
     std::string partitionFile
         = Configured::getConfiguration(keyMap.at(PARTITIONFILE_KEY), std::string("partition.nc"));
+    m_etadata.getPartitionMetadata(partitionFile);
```
TomMelt, line 122: it's really not a big issue, but I am just curious. Why is the metadata called `m_etadata`?

TomMelt (approved):
Looks good to me.

core/src/Model.cpp
draenog: This is question to @timspainNERSC.

timspainNERSC: <_< >_> 
No good reason.
Some classes, especially from early on in development, have a `m_` prefix for member variables. It slightly amused me to not repeat the 'm' in `m_metadata`.

TomMelt: fair enough. Thanks for the answer.

Issue #453, opened 2023-12-04
Free drift dynamics


timspainNERSC:
Currently the mEVP dynamics are known to work and the brittle (MEB, BBM) rheologies seem not to (for me, on my computer). In order to test the dynamics switcher, a second known-working solution to the momentum would be useful. Largely using the code that already exists, a free-drift 'solver' could be easily put together.

Related to #378

PR #522, opened 2024-03-21
Free drift dynamics: now with dynamics switcher!

timspainNERSC:
# Free drift dynamics
## Fixes \#453

---

# Change Description

Add a free drift equivalent to MEVPStep and BBMStep. Running this depends on implementing the dynamics switcher (closing #378).

---

# Test Description

Not yet. Should be tested once this form of dynamics can be tested.

---

# Documentation Impact

The free drift dynamics step needs to be documented along with the other available dynamics steps.

### DIFF
--- a/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
+++ b/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
@@ -21,7 +21,7 @@ class FreeDriftDynamics : public IDynamics {
 public:
     FreeDriftDynamics()
         : IDynamics()
-        , kernel()
+        , kernel(params)
     {
         getStore().registerArray(Protected::ICE_U, &uice, RO);
         getStore().registerArray(Protected::ICE_V, &vice, RO);
@@ -85,6 +85,7 @@ public:
 private:
     // TODO: How to get the template parameters here?
     FreeDriftDynamicsKernel<6> kernel;
+    DynamicsParameters params;
 };
 }
 
--- a/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
+++ b/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
@@ -18,8 +18,6 @@
 namespace Nextsim {
 static const std::vector<std::string> namedFields = { hiceName, ciceName, uName, vName };
 
-// Degrees to radians as a hex float
-static const double radians = 0x1.1df46a2529d39p-6;
 
 class FreeDriftDynamics : public IDynamics {
 public:
@@ -55,6 +53,9 @@ public:
 
     void setData(const ModelState::DataMap& ms) override
     {
+        // Degrees to radians as a hex float
+        static const double radians = 0x1.1df46a2529d39p-6;
+
         IDynamics::setData(ms);
 
         bool isSpherical = checkSpherical(ms);
--- a/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
+++ b/core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
@@ -1,0 +1,91 @@
+/*!
+ * @file FreeDriftDynamics.hpp
+ *
+ * @date 27 Mar 2023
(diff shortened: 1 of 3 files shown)


MarionBWeinzierl (commented):
There is quite a bit copy-paste between the different Dynamics implementation (including the same comments about spherical coordinates). I am wondering whether this could be avoided by moving them into a parent class, or otherwise a shared function?

timspainNERSC:
I have completely stomped on the code in the old branch, and cherry picked what I think is necessary to get free-drift dynamics.
@MarionBWeinzierl could you review the code, now it is free of all the repetition?
@einola could you review the maths to make sure I got all the commits that should be there?

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
```diff
@@ -0,0 +1,92 @@
...
+            double d = 25000; // 25 km in metres
+            // Fill the fake coordinate array
+            for (size_t j = 0; j < ModelArray::size(ModelArray::Dimension::YVERTEX); ++j) {
+                for (size_t i = 0; i < ModelArray::size(ModelArray::Dimension::XVERTEX); ++i) {
+                    fake25kmCoords.components({i, j})[0] = d * i;
+                    fake25kmCoords.components({i, j})[1] = d * j;
+                }
+            }
+            coords = fake25kmCoords;
+            // End of code to be removed
```
einola, line 73: Can we remove this now?

dynamics/src/include/FreeDriftDynamicsKernel.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+    const double cosOceanAngle = cos(radians * params.ocean_turning_angle);
+    const double sinOceanAngle = sin(radians * params.ocean_turning_angle);
+    const double NansenNumber = sqrt(params.F_atm / params.F_ocean);
+
+    void updateMomentum(const TimestepTime& tst) override
+    {
+#pragma omp parallel for
+        for (int i = 0; i < u.rows(); ++i) {
+            // Ice velocity is equal to ocean velocity
```
einola, line 63: That comment is not correct anymore

```suggestion
            // "Classic free drift", assuming zero atmospheric turning angle.
```

einola (changes requested):
The math looks fine, but there's still a reference to the "fake" spherical coordinates. This can be removed, right?

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
timspainNERSC: Yes, done.

dynamics/src/include/FreeDriftDynamicsKernel.hpp
timspainNERSC: Eh, close enough.

einola (approved):

core/src/modules/DynamicsModule/module.cfg
```diff
@@ -21,3 +21,8 @@ has_help = false
 file_prefix = BBMDynamics
 description = Discontinuous Galerkin dynamics with BBM rheology.
 has_help = false
+
+[Nextsim::FreeDriftDynamics]
+file_prefix = FreeDriftDynamics
+description = Free drift dynamics.
+has_help = false
```
MarionBWeinzierl, line 28: Just out of curiosity: What is the "has help" field for, what does it define/do?

MarionBWeinzierl (approved):
This looks sensible to me.

core/src/modules/DynamicsModule/module.cfg
timspainNERSC: Within a Module, there is a function that fetches the help text from any implementation that has configuration options (this is the text that is shown when running `./nextsim --help-config`). When I was writing the Module classes by hand, I had to choose which of the implementation sub-classes were interrogated to get help text. This field in the module configuration determines the same thing for the scripts that are building the `Module` classes.
There may be a better way to do this, but that should be another issue & PR.

Issue #527, opened 2024-04-04
Copying in the inner loop and other cleanup


timspainNERSC:
Currently the redesigned dynamics copies three big arrays (`DGVector<DGstress>` arrays, so the biggest we have) each sub timestep. This is inefficient and the copying must be avoided.

Futhermore, `cgParametericMomentum` _delenda est_.

PR #528, opened 2024-04-10
Dynamics references

timspainNERSC:
# Dynamics references
## Fixes \#527

Avoid doing big copies in tight loops by passing references, rather than copying the data when calling the stress update functions.

The edited files have also been `clang-format`ted.

### DIFF
--- a/dynamics/src/include/BBMStressUpdateStep.hpp
+++ b/dynamics/src/include/BBMStressUpdateStep.hpp
@@ -20,7 +20,7 @@ template <int DGadvection, int DGstress, int CG>
 class BBMStressUpdateStep : public StressUpdateStep<DGadvection, DGstress> {
 
 public:
-    typedef std::array<DGVector<DGstress>, N_TENSOR_ELEMENTS> SymmetricTensorVector;
+    typedef std::array<std::reference_wrapper<DGVector<DGstress>>, N_TENSOR_ELEMENTS> SymmetricTensorVector;
     BBMStressUpdateStep()
         : pmap (nullptr)
         , p_d (nullptr)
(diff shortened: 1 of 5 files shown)


core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
```diff
@@ -0,0 +1,91 @@
...
+    void setData(const ModelState::DataMap& ms) override
+    {
+        IDynamics::setData(ms);
+
+        bool isSpherical = checkSpherical(ms);
+
+        // TODO: Remove this when spherical coordinates are fully implemented
+        ModelArray coords;
+        if (isSpherical) {
+            std::cout << "Spherical coordinates are not yet implemented. Reverting to Cartesian." << std::endl;
```
MarionBWeinzierl, line 61: Should this be a warning rather than normal print out?

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
```diff
@@ -0,0 +1,91 @@
+/*!
+ * @file FreeDriftDynamics.hpp
```
MarionBWeinzierl, line 2: How is adding this new class related to cleaning up and avoiding the copying of the arrays? Can you add more information to the PR?

timspainNERSC: This PR needs to be merged after PR #522.

timspainNERSC: This PR is now independent of #522

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
timspainNERSC: This is now superseded code

Issue #530, opened 2024-04-15
simple example broken in recent commit


TomMelt:
commit [`e67982ae`](https://github.com/nextsimhub/nextsimdg/commit/e67982ae9b06b3a2ddb535af14f39221ef6a744c) added  `u`, `v` and `damage` to the restart file.

This commit breaks `run_simple_example.sh`

The following line in `ModelComponent.cpp` try to multiply `data` by `oceanMask()`, but `data` is set to an uninitialized pointer (/ null pointer) 

https://github.com/nextsimhub/nextsimdg/blob/bd673a8158ad568d5fe5e02bb8a54b14a2e357e4/core/src/ModelComponent.cpp#L106

`m_data = Eigen::Array<double,0,1,RowMajor> (data ptr: 0x0)}`

This comes from this part of the code. My guess is that `uice` and `vice` have not been initialized properly. @timspainNERSC, do you have any guesses what's going on?

https://github.com/nextsimhub/nextsimdg/blob/bd673a8158ad568d5fe5e02bb8a54b14a2e357e4/core/src/modules/include/IDynamics.hpp#L40-L47

timspainNERSC:
Rather annoyingly, `run_simple_example.sh` Works For Me :-/

Could you try the branch `issue530_dumbdummydynamics`? That should have fixed the obvious problem with the initialization of the u & v arrays with `DummyDynamics`.

TomMelt:
@timspainNERSC `issue530_dumbdummydynamics` fixes the issue, but I don't understand why. Could you explain?

timspainNERSC:
The u & v arrays need to be initialized. This is done in `IDynamics::setData()`. However, `DummyDynamics` overrode this with its own empty implementation of that function. By deleting this empty function, the base class implementation of `setData()` is called instead and initializes u and v arrays correctly.

PR #532, opened 2024-04-17
Initialize the u,v arrays in DummyDynamics

timspainNERSC:
# Initialize the u,v arrays in DummyDynamics
## Fixes \#530

---
# Change Description

The u & v `ModelArrays` need to be initialized. This is done in `IDynamics::setData()`. However, `DummyDynamics` overrides this with its own empty implementation of `setData()`. By deleting this empty function, the base class implementation of setData() is called instead and initializes u and v arrays correctly.

---
# Test Description

Any model run using `DummyDynamics` as its implementation of `IDynamics` now works correctly.


TomMelt (approved):
Thanks @timspainNERSC , I have tested and it fixes the issue.

PR #540, opened 2024-05-02
Improve clang format

TomMelt:
previously mac was used a base image for `clang-format`, but this is slower than using Ubuntu.


jwallwork23 (approved):
Looks good, verified that it does indeed speed things up.

PR #545, opened 2024-05-07
Pass `ModelArray` to `Xios.write`, rather than array of `double`s

### LINKED ISSUE #544: Pass `ModelArray` to `Xios.write` rather than array of `double`s
The newly introduced `Xios` class has a `write` method that accepts a C-style array of `double`s, along with its dimensions in each direction. It would be more in line with Nextsim's approach to instead pass a `ModelArray`, which already knows about both the data and dimensions.

jwallwork23:
# Pass `ModelArray` to `Xios.write`, rather than array of `double`s
### Task List

---
# Change Description

In the existing implementation of the `Xios` class, its `write` method accepts a C-style array of `double`s, along with its dimensions in each direction. The approach adopted here - better in line with Nextsim's approach - is to instead pass a `ModelArray`, which knows about both the data and dimensions. I think this change is fairly uncontroversial.

While working on this last week, @timspainNERSC said it would be preferable if runtime configuration of Nextsim-DG would not required when using Xios functionality. To address this, the changes here also make the `Xios` class independent of Nextsim-DG. Is this acceptable @TomMelt, or are there other reasons that we should keep the configuration code in that we aren't aware of?

# Test Description

@timspainNERSC also said it would be preferable if the Xios tests are built without linking to `nextsimlib`, to keep them separate. As such, the Xios test build has been reworked. In the current setup, the tests build and all pass. However, the latest commit (2e83114b1fd9efda9a0b99ca7900ce0ff8d0d8a1) is a temporary hack to get the Xios test to build correctly. @TomMelt do you have any insight on how to avoid this?

### Pre-Request Checklist

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -68,10 +68,7 @@ void Xios::finalize()
  * Configure the XIOS server if XIOS is enabled in the settings.
  *
  */
-void Xios::configure()
-{
-    configureServer();
-}
+void Xios::configure() { configureServer(); }
 
 //! Configure calendar settings
 void Xios::configureServer()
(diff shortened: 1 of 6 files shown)


jwallwork23:
Note that I had to turn off `clang-format` to get the includes in the right order in the Xios test. (88ddbceeb12fe643251e54e156acc49b0ecc5ba3)

jwallwork23:
This PR is out of date and could do with splitting into separate PRs for the different issues it addresses. Closing now.

Issue #550, opened 2024-05-14
Rename the x & y dimensions


timspainNERSC:
For Cartesian restart files, both the dimensions of the fields and the coordinate positions share the names `x` and `y`. This is confusing to humans reading these files, but causes Panoply to refuse to plot the fields at all.

Rename the dimensions to `x_dim` and `y_dim`.

PR #551, opened 2024-05-14
Dimension names

timspainNERSC:
# Dimension names
## Fixes \#550

---
# Change Description

For Cartesian restart files, both the dimensions of the fields and the coordinate positions share the names x and y. This is confusing to humans reading these files, but causes Panoply to refuse to plot the fields at all.

Rename the dimensions to x_dim and y_dim.

---
# Test Description

The `ParaGrid` and `RectGrid` tests should still run

---
# Documentation Impact

No impact: these names are held within the creation scripts.

### DIFF
--- a/core/src/ParaGridIO.cpp
+++ b/core/src/ParaGridIO.cpp
@@ -28,8 +28,8 @@
 namespace Nextsim {
 
 const std::map<std::string, ModelArray::Type> ParaGridIO::dimensionKeys = {
-    { "yx", ModelArray::Type::H },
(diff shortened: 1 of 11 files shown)


jwallwork23 (approved):
Seems reasonable to me. The only place that it's particularly difficult to read is in the `dimensionKeys` definition, but these are private anyway.

timspainNERSC:
> Seems reasonable to me. The only place that it's particularly difficult to read is in the `dimensionKeys` definition, but these are private anyway.

I'm definitely open to a better way to handle that part :)

PR #555, opened 2024-05-15
Wrap XIOS API calls part 1: getters

### LINKED ISSUE #553: Wrap getters from XIOS API
Currently, the XIOS functionality relies on axes, grids, fields, and files being defined in the user-defined `iodef.xml` file.

We should wrap the getters from the XIOS API so that neXtSIM-DG can read these.

jwallwork23:
# Wrap XIOS API calls part 1: getters
### Task List

---
# Change Description

Currently, the XIOS functionality relies on axes, grids, fields, and files being defined in the user-defined `iodef.xml` file. This PR wraps the corresponding parts of the XIOS API so these can be read by neXtSIM-DG.

- [x] Address methods in `Xios.hpp` not yet implemented.
- [x] Interface XIOS functions from `icfile.cpp` (for handling files).
- [x] Wrap axis getters from XIOS API.
- [x] Wrap domain getters from XIOS API.
- [x] Wrap grid getters from XIOS API.
- [x] Wrap field getters from XIOS API.
- [x] Wrap file getters from XIOS API.

---
# Test Description

Update `TestXiosInit_MPI2.cpp`...
- [x] To use the updated `Xios.getCurrentDate` implementation.
- [x] To test axis getters.
- [x] To test domain getters.
- [x] To test grid getters.
- [x] To test field getters.
- [x] To test file getters.
- [x] Fix write to file.

---
### Pre-Request Checklist

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -534,16 +534,103 @@ std::vector<double> Xios::getDomainLatitudeValues(std::string domainId)
 }
 
 /*!
- * Get the file associated with a given ID
+ * Get the field associated with a given ID
  *
- * @param the file ID
- * @return a pointer to the XIOS CFile object
+ * @param the field ID
+ * @return a pointer to the XIOS CField object
  */
-xios::CFile* Xios::getFile(std::string fileId)
+xios::CField* Xios::getField(std::string fieldId)
 {
-    xios::CFile* file = NULL;
-    cxios_file_handle_create(&file, fileId.c_str(), fileId.length());
-    return file;
+    xios::CField* field = NULL;
+    cxios_field_handle_create(&field, fieldId.c_str(), fieldId.length());
+    return field;
+}
+
+/*!
+ * Get the name of a field with a given ID
+ *
+ * @param the field ID
+ * @return name of the corresponding field
+ */
+std::string Xios::getFieldName(std::string fieldId)
+{
+    int size = 20;
+    char cStr[size];
+    xios::CField* field = getField(fieldId);
+    cxios_get_field_name(field, cStr, size);
+    std::string fieldName(cStr, size);
+    boost::algorithm::trim_right(fieldName);
+    return fieldName;
+}
+
+/*!
+ * Get the operation associated with a field with a given ID
+ *
+ * @param the field ID
+ * @return operation used for the corresponding field
+ */
+std::string Xios::getFieldOperation(std::string fieldId)
+{
+    int size = 20;
+    char cStr[size];
+    xios::CField* field = getField(fieldId);
+    cxios_get_field_operation(field, cStr, size);
+    std::string operation(cStr, size);
+    boost::algorithm::trim_right(operation);
+    return operation;
+}
+
+/*!
+ * Get the grid reference associated with a field with a given ID
+ *
+ * @param the field ID
+ * @return grid reference used for the corresponding field
+ */
+std::string Xios::getFieldGridRef(std::string fieldId)
+{
+    int size = 20;
+    char cStr[size];
+    xios::CField* field = getField(fieldId);
+    cxios_get_field_grid_ref(field, cStr, size);
+    std::string gridRef(cStr, size);
+    boost::algorithm::trim_right(gridRef);
+    return gridRef;
+}
+
+/*!
+ * Verify whether a name has been defined for a given field ID
+ *
+ * @param the field ID
+ * @return `true` if the name has been set, otherwise `false`
+ */
+bool Xios::isDefinedFieldName(std::string fieldId)
+{
+    xios::CField* field = getField(fieldId);
+    return cxios_is_defined_field_name(field);
+}
+
+/*!
+ * Verify whether an operation has been defined for a given field ID
+ *
+ * @param the field ID
+ * @return `true` if the operation has been set, otherwise `false`
+ */
+bool Xios::isDefinedFieldOperation(std::string fieldId)
+{
+    xios::CField* field = getField(fieldId);
+    return cxios_is_defined_field_operation(field);
+}
+
+/*!
+ * Verify whether a grid reference has been defined for a given field ID
+ *
+ * @param the field ID
+ * @return `true` if the grid reference has been set, otherwise `false`
+ */
+bool Xios::isDefinedFieldGridRef(std::string fieldId)
+{
+    xios::CField* field = getField(fieldId);
+    return cxios_is_defined_field_grid_ref(field);
 }
 
 /*!
@@ -558,16 +645,30 @@ bool Xios::validFileId(std::string fileId)
     cxios_file_valid_id(&valid, fileId.c_str(), fileId.length());
     return valid;
 }
+
 /*!
  * Verify whether an output frequency has been defined for a given file ID
  *
  * @param the file ID
  * @return `true` if the output frequency has been set, otherwise `false`
  */
-bool Xios::isDefinedOutputFreq(std::string fileId)
+bool Xios::isDefinedFileOutputFreq(std::string fileId)
 {
     xios::CFile* file = getFile(fileId);
     return cxios_is_defined_file_output_freq(file);
+}
+
+/*!
+ * Get the file associated with a given ID
+ *
+ * @param the file ID
+ * @return a pointer to the XIOS CFile object
+ */
+xios::CFile* Xios::getFile(std::string fileId)
+{
+    xios::CFile* file = NULL;
+    cxios_file_handle_create(&file, fileId.c_str(), fileId.length());
+    return file;
 }
 
 /*!
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -405,10 +405,9 @@ xios::CDomain* Xios::getDomain(std::string domainId)
  */
 std::string Xios::getDomainType(std::string domainId)
 {
-    int size = 20;
-    char cStr[size];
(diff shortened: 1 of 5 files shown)


jwallwork23:
I realised wrapping the entire API will be a significant change and that wrapping the getters will be easier than wrapping the setters. As such, this just addresses the getters, leaving the setters for a subsequent PR.

core/src/Xios.cpp
```diff
@@ -213,69 +275,517 @@ void Xios::setCalendarTimestep(cxios_duration timestep)
...
+
+/*!
+ * Get the type of a given domain
+ *
+ * @param the domain ID
+ * @return the corresponding domain type
+ */
+std::string Xios::getDomainType(std::string domainId)
+{
+    int size = 20;
```
jwallwork23, line 416: This is just the string length used in the toy code. We'd probably want to use a larger value.

core/src/Xios.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include "include/Xios.hpp"
 
+#include <boost/algorithm/string.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/format.hpp>
 #include <boost/format/group.hpp>
```
jwallwork23, line 29: I figured it's okay to use boost for trimming strings, given that it's already used for other stuff.

core/src/include/Xios.hpp
```diff
@@ -28,34 +28,91 @@ class Xios : public Configured<Xios> {
...
-    void finalize();
+    void close_context_definition();
     void context_finalize();
+    void finalize();
     bool isInitialized();
 
     void configure() override;
     void configureServer();
-    void configureCalendar();
+    void configureCalendar(); // TODO?
```
jwallwork23, line 38: @TomMelt I noticed you added this method to the class definition but didn't implement it. Did you have something in mind?

core/src/include/Xios.hpp
```diff
@@ -28,34 +28,91 @@ class Xios : public Configured<Xios> {
...
+    std::string convertXiosDatetimeToString(cxios_date datetime, bool isoFormat = true);
+    void printCXiosDate(cxios_date date);
+    void printCXiosDuration(cxios_duration duration);
 
+    /* Calendar */
     cxios_date getCalendarOrigin();
     cxios_date getCalendarStart();
-    xios::CDate getCurrentDate();
     cxios_duration getCalendarTimestep();
+    void getCalendarConfiguration(); // TODO?
```
jwallwork23, line 49: @TomMelt similarly here

core/src/Xios.cpp
```diff
@@ -49,19 +50,26 @@ Xios::Xios() { configure(); }
...
 {
     if (isEnabled) {
         cxios_context_finalize();
+    }
+}
+
+//! Close context and finialize server
+void Xios::finalize()
+{
+    if (isEnabled) {
```
jwallwork23, line 72: Note that I needed to rework these to get tests to pass because the context definition should only be closed once the domain has been set. (See f7fbc446)

core/src/include/Xios.hpp
```diff
@@ -28,34 +28,91 @@ class Xios : public Configured<Xios> {
...
+    void createDomain(std::string domainId); // TODO
+    void setDomainType(std::string domainId, std::string domainType); // TODO
+    void setDomainGlobalLongitudeSize(std::string domainId, int size); // TODO
+    void setDomainGlobalLatitudeSize(std::string domainId, int size); // TODO
+    void setDomainLongitudeSize(std::string domainId, int size);
+    void setDomainLatitudeSize(std::string domainId, int size);
+    void setDomainLongitudeStart(std::string domainId, int start);
+    void setDomainLatitudeStart(std::string domainId, int start);
+    void setDomainLongitudeValues(std::string domainId, std::vector<double> values);
+    void setDomainLatitudeValues(std::string domainId, std::vector<double> values);
```
jwallwork23, line 74: I did end up implementing a few setters (here) because they are quantities that are local to the MPI process.

core/src/Xios.cpp
TomMelt: yeah, I would say so.

The documentation suggests installing all of boost and the `spack` package I made includes `algorithms` as part of the base `spack` recipe. So, I think it makes sense to use it.

@timspainNERSC do you have any objections?

core/src/include/Xios.hpp
TomMelt: Ahh, I think I was going to split configure up further but I realised it wasn't necessary.

Feel free to delete this method.

TomMelt:
```suggestion
```

core/src/include/Xios.hpp
TomMelt: Again, there's no specific plan. Could possibly be used to group the calendar specific variables together perhaps. Like time_origin, start_date. But I think it would ruin the generality so I guess I never added it.

Again, feel free to delete it or keep it if you think it will be useful.

core/src/include/Xios.hpp
jwallwork23: Thanks. Yeah I figured I've probably wrapped some parts of the API we won't end up using but I thought it was better to do it this way and prune it later, rather than having to do more wrapping.

jwallwork23: Good point on using neXtSIM-DG's data types (although I thought it would be `ModelArray` rather than `ModelArrayRef`). Perhaps if we merge this PR first then #545 could be updated to do this for these setters as well as the `write` method(s).

core/src/include/Xios.hpp
jwallwork23: Addressed in acc3b554.

core/src/include/Xios.hpp
jwallwork23: Addressed in acc3b554.

core/src/include/Xios.hpp
```diff
@@ -28,34 +28,89 @@ class Xios : public Configured<Xios> {
...
-    std::string convertXiosDatetimeToString(cxios_date datetime, bool isoFormat = true);
 
-    void printCXiosDate(cxios_date date);
-    void printCXiosDuration(cxios_duration duration);
+    /* Axis */
+    void createAxis(std::string axisId); // TODO
+    void setAxisSize(std::string axisId, int size); // TODO
+    void setAxisValues(std::string axisId, std::vector<double> values); // TODO
+    int getAxisSize(std::string axisId);
+    std::vector<double> getAxisValues(std::string axisId);
```
jwallwork23, line 60: Oh that was another thing I wanted to comment on here. I've wrapped C-style arrays of `double`s with `std::vector` because I read the following in the neXtSIM-DG README:
> * All array operations should be done using std::vectors - not C-style arrays
> * The use of C-style pointers, new, and delete is strongly discouraged

core/src/Xios.cpp
TomMelt: we should check what happens if `domain` is larger than 20. Does `xios` throw an error?

core/src/include/Xios.hpp
TomMelt: sounds like a good idea

core/src/Xios.cpp
```diff
@@ -213,69 +275,517 @@ void Xios::setCalendarTimestep(cxios_duration timestep)
...
+
+/*!
+ * Get the grid reference associated with a field with a given ID
+ *
+ * @param the field ID
+ * @return grid reference used for the corresponding field
+ */
+std::string Xios::getFieldGridRef(std::string fieldId)
+{
+    int size = 20;
```
TomMelt, line 579: We may want to make `size` a class variable so we can change it once for all fields? Possibly it could become a user-defined input through the `Nextsim::Configurator`

core/src/Xios.cpp
```diff
@@ -213,69 +275,517 @@ void Xios::setCalendarTimestep(cxios_duration timestep)
...
+
+/*!
+ * Get the file associated with a given ID
+ *
+ * @param the file ID
+ * @return a pointer to the XIOS CFile object
+ */
+xios::CFile* Xios::getFile(std::string fileId)
+{
+    xios::CFile* file = NULL;
```
TomMelt, line 657: just a thought, not necessarily for this PR. Would it be sensible to throw an exception if any of the getter objects are still `NULL`? Ill be honest... I didn't really check what happens if `xios` getters fail. If `xios` throws the error anyway then it may not be necessary.

TomMelt (approved):
thanks @jwallwork23. This is great. 

I ran the test locally on my laptop and everything looks good :+1:

core/src/Xios.cpp
jwallwork23: Good point. I've opened #562 to address this.

core/src/Xios.cpp
jwallwork23: Good idea. Addressed in f60e88fc19cc50ff0cbde3b0438daa6798b5d6ef, just adding a public attribute for now. But yes, could be added to the configurator later, too.

core/src/Xios.cpp
jwallwork23: I tried this with `size = 40` and the test still passed. Although, just to note, the size here is string length rather than domain size. So the `size` here is now set across the whole class in f60e88fc19cc50ff0cbde3b0438daa6798b5d6ef.

TomMelt (approved):
Thanks for addressing my comments. Looks good to me 👍

PR #557, opened 2024-05-15
Dimension names redux

timspainNERSC:
# Dimension names redux
## Fixes \#550 (again)

---
# Change Description

Accept both the new (`xdim`, `ydim`, `zdim`) and old (`x`, `y`, `z`) dimension names to increase compatibility.

---
# Test Description

Added a test that reads in a file with the old dimension names are ensures the dimensions are still set correctly.

---
# Documentation Impact

No impact. This is a convenience detail and no users should be writing new restart files using the old dimension names.

---
# Other Details

The generated `old_names.nc` file will suffer the same deficiencies as the restart files used to but will never be updated.

### DIFF
--- a/core/test/old_names.py
+++ b/core/test/old_names.py
@@ -1,8 +1,6 @@
 import netCDF4
 import numpy as np
-import numpy.ma as ma
 import time
-import math
 
 # Create a restart file that specifically uses the old dimension names
 if __name__ == "__main__":
@@ -53,9 +51,7 @@ if __name__ == "__main__":
 
     # Array coordinates
     space = 25000. # 25 km in metres
-    node_x = np.zeros((nfirst + 1, nsecond + 1))
-    node_y = np.zeros((nfirst + 1, nsecond + 1))
-    
+
     x_vertex1d = np.arange(0, space * (nsecond + 1), space)
     y_vertex1d = np.arange(0, space * (nfirst +1), space)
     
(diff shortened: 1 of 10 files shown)


core/test/old_names.py
```diff
@@ -0,0 +1,121 @@
...
+    n_coords_comp = datagrp.createDimension("ncoords", n_coords)
+    
+    field_dims = ("y", "x")
+    coord_dims = ("yvertex", "xvertex", "ncoords")
+    zfield_dims = ("z", "y", "x")
+
+    # Array coordinates
+    space = 25000. # 25 km in metres
+    node_x = np.zeros((nfirst + 1, nsecond + 1))
+    node_y = np.zeros((nfirst + 1, nsecond + 1))
```
jwallwork23, line 57: These don't seem to be used anywhere else.

core/test/old_names.py
```diff
@@ -0,0 +1,121 @@
+import netCDF4
+import numpy as np
+import numpy.ma as ma
+import time
+import math
```
jwallwork23, line 5: `ma` and `math` appear to be unused.

jwallwork23 (approved):
I think I followed the login in `old_names.py` correctly. Looks good to me, there are just a couple of unused modules and arrays that are set to zero and never used.

core/test/old_names.py
timspainNERSC: Indeed not, removed.

core/test/old_names.py
timspainNERSC: Indeed not, removed.

Issue #537, opened 2024-05-02
Periodic boundary conditions


einola:
We need periodic boundary conditions for

 - Idealised setups
 - For regional setups with a zonal periodicity (e.g. the southern ocean)
 - For global "tri-polar" setup with a fold at the north pole

The main tasks are to

- [ ] Implement periodicity in the dynamics and advection code
- [ ] Make sure information about which borders are periodic flows between the restart file and the dynamics

einola:
We'll do this as part of the MPI communication.

PR #590, opened 2024-06-25
WIP: Grid boundary handling

timspainNERSC:
# Grid boundary handling
## Fixes \#537

---
# Change Description

## Periodic boundary

a periodic boundary element consists of
- an edge (given by its id)
- two elements on the left/right or bottom/top of the edge
- two indices specifying where the edge is as seen from the element

## Dirichlet-type conditions:

- dirichlet (v=0)
- open (free inflow/outflow with zero concentration on inflow)
- freeslip (normal velocity zero)

the all have the interface std::array<std::vector<size_t>, 4> where the outer index corresponds to elements that have boundary on the bottom, right, top, left and the inner index to corresponding id of the element.

## Code changes

Handling of periodic boundaries required several changes in DGTransport.cpp and cgParametricMomentum.cpp This mostly includes restructuring code and adding the edge (bottom, right, top, left) as template parameter.

Further, code is added to deal with open and freeslip boundaries.

---
# Test Description

There is a test in the (new) application folder.

---
# Documentation Impact

TODO

### DIFF
--- a/dynamics/CMakeLists.txt
+++ b/dynamics/CMakeLists.txt
@@ -4,6 +4,7 @@ add_subdirectory(src)
 #endif()
 
 add_subdirectory(tests)
+add_subdirectory(applications)
 
 set(NextsimSources
     "${NextsimSources}"
--- a/dynamics/applications/CMakeLists.txt
+++ b/dynamics/applications/CMakeLists.txt
@@ -1,0 +1,29 @@
+# Include dynamics headers
+set(DynamicsSrc "../src")
+set(SRC_DIR "${DynamicsSrc}")
+set(CoreSrc "../../core/src")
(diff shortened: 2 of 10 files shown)


winzerle:
Ok, I’ll change that auch the mesh is directly created.Sent from my iPhoneOn 25. Jun 2024, at 11:30, Tim Spain ***@***.***> wrote:﻿
@winzerle If we want to avoid divergence between you test cases and the dynamics used in nextSIM-DG, then not using ParametricMesh::readmesh() is essential.

timspainNERSC:
Handling of periodic boundaries (and to an extent open boundaries) will be moved to MPI, rendering much of this pull request redundant.

The improvements related to edge handling will be moved to another PR.

Issue #602, opened 2024-07-04
Using BBMDynamics causes a crash if the restart file does not contain a damage field


timspainNERSC:
The BBMDynamics expects to be able to initialize the damage field from the restart file. If the dataset from read from the restart file does not contain a field named `damage`, then the function `BBMDynamics::setData()` will crash the model with an exception when it tries to find the `damage` field. The damage field is considered optional on restart, as the other rheologies do not require it to be present.

PR #603, opened 2024-07-04
Define a default value of damage, if it is not provided.

timspainNERSC:
# Define a default value of damage, if it is not provided.
## Fixes \#602

---
# Change Description

Within `BBMDynamics`, separate the data fields that are obtained from the restart file into required and optional fields. Required fields are set immediately, just as before this change. Optional fields, which include the damage field, are mapped to an array type and default value. If the field is present, then the data is set immediately, as before. Otherwise a new `ModelArray` is created of the specified type and filled with the default value. This is then passed into the `setData()` function.

### DIFF
--- a/core/src/modules/DynamicsModule/BBMDynamics.cpp
+++ b/core/src/modules/DynamicsModule/BBMDynamics.cpp
@@ -12,7 +12,10 @@
 
 namespace Nextsim {
 
-static const std::vector<std::string> namedFields = { hiceName, ciceName, uName, vName, damageName };
+static const std::vector<std::string> namedFields = { hiceName, ciceName, uName, vName };
+static const std::map<std::string, std::pair<ModelArray::Type, double>> defaultFields = {
+    { damageName, { ModelArray::Type::H, 1.0 } },
+};
 BBMDynamics::BBMDynamics()
     : IDynamics(true)
     , kernel(params)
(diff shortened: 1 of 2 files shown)


core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -38,9 +41,24 @@ void BBMDynamics::setData(const ModelState::DataMap& ms)
...
+        const std::string& fieldName = entry.first;
+        if (ms.count(fieldName) > 0)
+            kernel.setData(fieldName, ms.at(fieldName));
+        // Fill data that is not supplied, masking if the mask is available
+        ModelArray data(entry.second.first);
+        data.resize();
+        // Fill the default value
+        data = entry.second.second;
+        // Mask the default data
+        kernel.setData(fieldName, mask(data));
```
einola, line 60: Shouldn't this bit be in an ``else`` clause?

einola (changes requested):
Very good. But you missed an ``else`` (or I really don't understand what's going on).

einola:
This made me realise that we have an issue with ``make_init_base.py``, whereby the restart files it produces have ``damage == 0`` everywhere by default. This is a one-line change (line 50). Should we rope it into this PR or make a separate issue?

core/src/modules/DynamicsModule/BBMDynamics.cpp
timspainNERSC: Yes, it should

timspainNERSC:
> This made me realise that we have an issue with `make_init_base.py`, whereby the restart files it produces have `damage == 0` everywhere by default. This is a one-line change (line 50). Should we rope it into this PR or make a separate issue?

Make a separate issue. It will be a quick review.

einola (approved):

Issue #604, opened 2024-07-04
poor performance due to IceGrowth::initializeThicknesses


Thanduriel:
I noticed that 95% of the total runtime in the mevp benchmarks is spend in `IceGrowth::initializeThicknesses`. This function should not take any significant time, especially since only dummy physics are used.

PR #605, opened 2024-07-04
Fix poor performance of `IceGrowth::initializeThicknesses`

Thanduriel:
# Fix poor performance of `IceGrowth::initializeThicknesses`
## Fixes \#604

# Change Description

The poor performance is a result of the complete `newice` array getting initialized to 0 for every single element. Instead the method `IceGrowth::initializeThicknessesElement` should only write the value for element `i`, as it does for every other array. This single line change gives me roughly a x20 speedup for 2.6e5 elements.

---
# Test Description

---
# Documentation Impact

---


timspainNERSC (approved):
Oops!

PR #606, opened 2024-07-05
Add DG components to the restart files

### LINKED ISSUE #592: Full DG components in restart files
For perfect restarts, the full set of DG components have to be output and ingested. Currently, only the finite volume DG0 components are.

### LINKED ISSUE #595: Specify the DG degree in one place
Currently, there are at least two (dissimilar) defintions of the number of DG components. The degree should be determined in one place, preferably outside the core model, so that both the dynamics kernels and the ModelArrayDetails can be set with the same, matching number of DG components

timspainNERSC:
# Add DG components to the restart files
## Fixes #592, fixes #595

---
# Change Description

Part of this PR moves the definition of the number of DG components and the CG degree from hard coded values spread between ModelArrayDetails.cpp and the dynamics module and kernel classes. These values are now defined as part of the CMake configuration step, with the CMake command line argument -DDynamicsType= with valid parameter values DG1 and DG2. A value of DG1 sets the advection DG degree to 1 and the number of advection DG components to 3. A value of DG2 sets the advection DG degree to 2 and the number of advection DG components to 6. In both cases the CG degree is set to 2 and the number of DG stress advection components to 8.

These values are implemented as values for preprocessor macros that are applied to all of the source files compiled into the nextsimlib compilation target. This is then used by the model binary and most of the unit tests.

The ModelArrayDetails.cpp file defines the number of components of the DG-related ModelArrays based on these values, but also defines values equivalent to the DG1 option as defaults.

In the dynamics, the number of advection DG components is used as the template parameter in instantiating the relevant dynamics kernel. The CG degree is used in the CGDynamicsKernel class to as the CGdegree template parameter and to calculate the number of DG components for stress. This latter value is calculated using the CG2DGCOMPONENTS function. These two numbers of DG components are used as the template parameters for the DynamicsKernel class. Within these various dynamics kernel classes the template parameters for the underlying dynamics classes such as DGVector and CGVector are set.

---
# Test Description

Added `PrognosticDataIO_test.cpp`, which tests the writing function `PrognosticData::writeRestartFile()`. This is currently in a separate source file so that `PrognosticData_test` can run with MPI enabled, which does not currently allow I/O related tests to pass. Once parallel I/O is fully integrated into the model, these two tests should be combined.


core/src/modules/DynamicsModule/include/BBMDynamics.hpp
```diff
@@ -12,6 +12,11 @@
 #include "include/MEBParameters.hpp"
 #include "include/IDynamics.hpp"
 
+#ifndef DGCOMP
+#define DGCOMP 3 // define to make red lines go away in the IDE
```
jwallwork23, line 16: Perhaps reword as "define to stop static analysis tools from raising unwanted errors"?

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
```diff
@@ -15,6 +15,11 @@
 #include "include/ModelComponent.hpp"
 #include "include/gridNames.hpp"
 
+#ifndef DGCOMP
+#define DGCOMP 3 // define to make red lines go away in the IDE
```
jwallwork23, line 19: Similarly here.

core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
```diff
@@ -18,6 +18,11 @@
 #include "include/ModelArray.hpp"
 #include "include/ModelComponent.hpp"
 
+#ifndef DGCOMP
+#define DGCOMP 3 // define to make red lines go away in the IDE
```
jwallwork23, line 22: And here.

docs/installation.rst
```diff
@@ -122,6 +122,22 @@ After all dependencies have been installed, we can build the code:
         cmake ..
         make
 
+Configuring the dynamics
+------------------------
+The dynamics for nextSIM are chosen at the point of configuring CMake. This is in contrast to most of the model configuration, which is done at model run time. The dynamics are set through the configuration option ``DynamicsType``. The available options for the dynamics are
```
jwallwork23, line 127: Should probably put nextSIM-DG.

docs/installation.rst
```diff
@@ -122,6 +122,22 @@ After all dependencies have been installed, we can build the code:
...
+
+* ``DG1``: First order discontinuous Galerkin dynamics. Advection calculations are performed with 3 DG components.
+
+  * This is the default option is no other option is provided to CMake.
```
jwallwork23, line 131: is->if

docs/installation.rst
```diff
@@ -122,6 +122,22 @@ After all dependencies have been installed, we can build the code:
         cmake ..
         make
 
+Configuring the dynamics
+------------------------
+The dynamics for nextSIM are chosen at the point of configuring CMake. This is in contrast to most of the model configuration, which is done at model run time. The dynamics are set through the configuration option ``DynamicsType``. The available options for the dynamics are
+
+* ``DG1``: First order discontinuous Galerkin dynamics. Advection calculations are performed with 3 DG components.
```
jwallwork23, line 129: Perhaps for clarity mention that this is since we use a 2D triangular mesh.

docs/installation.rst
```diff
@@ -122,6 +122,22 @@ After all dependencies have been installed, we can build the code:
...
+
+  * This is the default option is no other option is provided to CMake.
+
+* ``DG2``: Second order discontinuous Galerkin dynamics. Advection calculations are performed with 6 DG components.
```
jwallwork23, line 133: Again, could mention that this is for a 2D triangular mesh.

core/src/PrognosticData.cpp
```diff
@@ -131,27 +135,33 @@ ModelState PrognosticData::getState() const
...
 ModelState PrognosticData::getStateRecursive(const OutputSpec& os) const
 {
-    ModelState state(getState());
-    state.merge(pAtmBdy->getStateRecursive(os));
-    state.merge(iceGrowth.getStateRecursive(os));
-    state.merge(pDynamics->getStateRecursive(os));
+    ModelState state;
+    /* If allComponents is set on the OutputSpec, then for any duplicate fields, the subsystems
+     * take priority, otherwise the fields held by PrognosticData itself. Note that std::map::merge
+     * will not overwrite existing keys, so the first one that exists will survive.
```
jwallwork23, line 152: Interesting. I was unfamiliar with `map.merge` before reading this. If I understand correctly, it's a merge which always chooses the existing values in the case of a conflict? It's the opposite from Python's `dict.update` method.

core/test/PrognosticDataIO_test.cpp
```diff
@@ -0,0 +1,154 @@
+/*!
+ * @file PrognosticData_test.cpp
+ *
+ * @date 7 Sep 2023
```
jwallwork23, line 4: Update date.

core/test/PrognosticDataIO_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+#define TRUE true
+#endif
+#ifndef FALSE
+#define FALSE false
+#endif
+#define ON true
+#define OFF false
+
+TEST_SUITE_BEGIN("PrognosticDataIO");
+// Once MPI-enabled IO is fully working, integrate this test into PrognosticData_test.cpp
```
jwallwork23, line 41: Flag this with `TODO` prefix so it's more easily `grep`-able.

jwallwork23 (approved):
Thanks @timspainNERSC. It's really useful to read about how restarts are done here.

The code looks good as far as I understand the parts that were edited. I've left some fairly minor fixes and recommendations.

core/src/modules/DynamicsModule/include/BBMDynamics.hpp
timspainNERSC: I went with `// Define to prevent errors from static analysis tools`

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
timspainNERSC: Done

core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
timspainNERSC: Done

docs/installation.rst
timspainNERSC: No. The general plan is to refer to what we call nextSIM-DG  as plain nextSIM in the future. Plus, it is used without the DG elsewhere in the same document.

docs/installation.rst
timspainNERSC: Added '2D rectangular grid'.

docs/installation.rst
timspainNERSC: Added '2D rectangular grid'.

docs/installation.rst
timspainNERSC: Done

core/test/PrognosticDataIO_test.cpp
timspainNERSC: Done

core/test/PrognosticDataIO_test.cpp
timspainNERSC: Added the TODO marker.

docs/installation.rst
jwallwork23: Oh of course, I forgot that it's rectangular. You need 3 DoFs with linearly independent coordinates to define a plane so makes sense that it would still be 3.

PR #630, opened 2024-07-29
Drop XIOS `isDefined*` and `areDefined*` methods

### LINKED ISSUE #629: XIOS: Inline `isDefined` methods
There are various `isDefined` methods for XIOS but these can probably be inlined into the getters and setters and dropped from the API.

jwallwork23:
# Drop XIOS `isDefined*` and `areDefined*` methods
### Task List

---
# Change Description

It's unnecessary for us to provide API functions for checking whether XIOS attributes are defined. We can cut down API size, implement self-testing code, and reduce verbosity by just inlining the associated XIOS functions within the getter and setter methods.

---
### Pre-Request Checklist

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -1092,7 +1092,7 @@ xios::CFileGroup* Xios::getFileGroup()
     xios::CFileGroup* group = NULL;
     cxios_filegroup_handle_create(&group, groupId.c_str(), groupId.length());
     if (!group) {
-        throw std::runtime_error("Xios: Null pointer for file_definition group");
+        throw std::runtime_error("Xios: Null pointer for group 'file_definition'");
     }
     return group;
 }
@@ -1108,7 +1108,7 @@ xios::CFile* Xios::getFile(const std::string fileId)
     xios::CFile* file = NULL;
     cxios_file_handle_create(&file, fileId.c_str(), fileId.length());
     if (!file) {
-        throw std::runtime_error("Xios: Null pointer for file with ID '" + fileId + "'");
+        throw std::runtime_error("Xios: Null pointer for file '" + fileId + "'");
     }
     return file;
 }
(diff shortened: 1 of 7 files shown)


core/src/Xios.cpp
timspainNERSC, line 1: I'm a bit unclear on whether all the mentions of longitude and latitude actually refer to those global coordinates. It looks like some errors in setting the size of the i and j dimensions would result in errors referring to longitude or latitude. In the case where we are running a regional model domain (like the Arctic domain I usually test on), this might be confusing. Conversely, some of the mentions of longitude in the error messages do seem to be referring to errors in setting the values of the geographical coordinate.
Could you make sure that wherever longitude or latitude are mentioned in the error messages, that is really what is being referred to?

timspainNERSC (changes requested):
Looks good, except for my comment on the contents of some of the error messages (if I am understanding correctly).

core/src/Xios.cpp
jwallwork23: Ah yes, thanks for pointing this out @timspainNERSC. I've checked over the domain methods and think I addressed the issue in my latest commit. Would it be useful to use the naming pattern `setDomainLocalLatitude` etc., to more clearly differentiate between the local and global versions?

jwallwork23: Thanks for the elaboration. I think XIOS uses the terms latitude and longitude because until recently it's always been used on lat-long grids. But yes, we can certainly using different names. In addition to reworking the error messages, it's probably a good opportunity to switch to more appropriate names for the methods, too. How about the following patterns?
```
setDomainLocalXSize(...)
getDomainGlobalYValues(...)
```

timspainNERSC: > Thanks for the elaboration. I think XIOS uses the terms latitude and longitude because until recently it's always been used on lat-long grids. But yes, we can certainly using different names. In addition to reworking the error messages, it's probably a good opportunity to switch to more appropriate names for the methods, too. How about the following patterns?
> 
> ```
> setDomainLocalXSize(...)
> getDomainGlobalYValues(...)
> ```

I'm definitely in favour of that.

jwallwork23: Okay I've now updated all the names, docstrings, and error messages/warnings. Hopefully clearer now. I also made one minor reordering for consistency.

timspainNERSC (approved):
That all looks good now

jwallwork23:
Thanks @timspainNERSC

PR #631, opened 2024-07-31
Ignore empty modules

### LINKED ISSUE #608: Remnant Module dirs when switching branches
If a branch introduces a new Module, switching away from it will leave the Module subdirectory behind, containing the generated `module.cpp` file. The Module generation code should ignore or delete such directories. The telltale signature is a Module directory containing only a `module.cpp` file.

timspainNERSC:
# Pull Request Title
## Fixes #608

---
# Change Description

Ignores any modules that don't contain a `module.cfg` file. These can be left over when changing `git` branches where a module exists to one where it does not. Before this change, the presence of the directory alone would trigger `CMake` to run the module build scripts, which fail due to the lack of any module files. Without a `module.cfg` file, there is no module to build, so ignoring these sub-directories is the right thing to do, whether they are left over from changing branches or for whatever other reason.

---
# Test Description

Hand tested before creating this PR. I don't see a good way to integrate a test for this.

### DIFF
--- a/core/src/modules/CMakeLists.txt
+++ b/core/src/modules/CMakeLists.txt
@@ -5,6 +5,10 @@ set(AllModuleSources "")
 file(GLOB ModuleSubDirs LIST_DIRECTORIES TRUE "*Module")
 
 foreach(ModuleSubDir IN LISTS ModuleSubDirs)
+    # Ignore directories without a module.cfg file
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Seems like a good fix for the issue you mention.

PR #632, opened 2024-07-31
Wrap XIOS frequency splitting and fix timestep bug

### LINKED ISSUE #554: Support XIOS' frequency splitting option
The `split_freq` attribute of XIOS' file class allows separate output files to be created for different time intervals, as defined by a frequency (e.g., every 2 simulated days).

jwallwork23:
# Wrap XIOS frequency splitting and fix timestep bug
### Task List

---
# Change Description

Closes #554. Wraps the `split_freq` option for XIOS Files, so that output files can be separated according to some frequency.

While working on this, I spotted a bug: the timestep is interpreted as minutes rather than seconds. This is now fixed.

---
# Test Description

A test for the frequency splitting option is added, which checks a frequency of 2 timesteps works.

Another test is added to verify that the timestep is set up correctly.

---
### Pre-Request Checklist

### DIFF
--- a/core/test/XiosFile_test.cpp
+++ b/core/test/XiosFile_test.cpp
@@ -83,6 +83,10 @@ MPI_TEST_CASE("TestXiosFile", 2)
     xios_handler.setFileOutputFreq(fileId, freq);
     REQUIRE(xios_handler.isDefinedFileOutputFreq(fileId));
     REQUIRE(xios_handler.getFileOutputFreq(fileId) == freq);
+    // Split frequency
+    const std::string splitFreq { "1ts" };
+    xios_handler.setFileSplitFreq(fileId, splitFreq);
+    REQUIRE(xios_handler.getFileSplitFreq(fileId) == splitFreq);
     // Add field
     xios_handler.fileAddField(fileId, "field_A");
     std::vector<std::string> fieldIds = xios_handler.fileGetFieldIds(fileId);
--- a/core/test/XiosFile_test.cpp
+++ b/core/test/XiosFile_test.cpp
@@ -1,7 +1,7 @@
 /*!
  * @file    XiosFile_test.cpp
  * @author  Joe Wallwork <jw2423@cam.ac.uk
- * @date    24 July 2024
+ * @date    31 July 2024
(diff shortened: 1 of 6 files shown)


core/test/XiosFile_test.cpp
```diff
@@ -83,6 +83,10 @@ MPI_TEST_CASE("TestXiosFile", 2)
     xios_handler.setFileOutputFreq(fileId, freq);
     REQUIRE(xios_handler.isDefinedFileOutputFreq(fileId));
     REQUIRE(xios_handler.getFileOutputFreq(fileId) == freq);
+    // Split frequency
+    const std::string splitFreq { "1ts" };
```
timspainNERSC, line 87: Is this a typical way of setting a std::string? I can see that it works, and how, but I would always use
```
const std::string splitFreq = "1ts";
```
I'd suggest we stick to one or the other in the code base. If the `{}` version is currently preferred, then I'll switch all my `const std::string`s over as I touch their files.

timspainNERSC (changes requested):
Could you take a look at the `const std::string` assignment? Otherwise, this all looks reasonable.

core/test/XiosFile_test.cpp
jwallwork23: Oh good point. In other tests I followed Tom's approach elsewhere, which read like
```
std::string splitFreq = { "1ts"};
```
I'd meant to use the same notation but must have reverted to this instead.

While we're on the topic, though, the C++ course I did recently did recommend initialising variables with
```
std::string splitFreq { "1ts"};
```
rather than
```
std::string splitFreq = "1ts";
```
There are some reasons behind this listed here: https://stackoverflow.com/questions/18222926/what-are-the-advantages-of-list-initialization-using-curly-braces. (TLDR: _Prefer {} initialization over alternatives unless you have a strong reason not to._)

timspainNERSC (approved):

core/test/XiosFile_test.cpp
timspainNERSC: Excellent, thank you :)

PR #560, opened 2024-05-21
Add MPI support to ParaGrid

TomMelt:
# Add MPI support to ParaGrid
closes #534
closes #454
closes #448 

---
# Task List
- [x] modify `make_init_para24x30.py` to use new DG data layout and remove CG vars
- [x] add MPI doctest support to `ParaGrid_test.cpp`
- [x] move `extent` and `start` variables into `ModelArray`'s `DimensionSpec` 
- [x] add special MPI test case to `ParaGrid_test.cpp`
- [x] add MPI support to `getModelState`
- [x] add MPI support to `dumpModelState`
- [x] add MPI support to `writeDiagnosticTime`
- [x] reinstate test `ConfigOutput_test.cpp` for MPI builds
- [x] add MPI support to `ERA5`/`TOPAZ` tests

---
# Change Description

After #331 added MPI parallelisation for thermodynamics on the RectGrid, this PR does the same for the Parametric grid. This should then check off the second task in #120 (_"MPI parallelization of thermodynamics where all operations, except for I/O, are local to an MPI rank"_)

---
# Test Description

- `ParaGrid_test.cpp` tests core functionality of ParaGrid (serial and MPI)
- `./nextsim --config-file config_para24x30.cfg` should provide an integration test (serial and MPI) (based on #506)

---
# Further work (for a future PR)

- add MPI to dynamics (to close PR #120)
- implement halo comms
- implement boundary conditions (as part of MPI) 
- tidy naming of variables in Domain Decomp tool

PR #635, opened 2024-08-02
XIOS: use Domain for xy-plane and Axis for z

### LINKED ISSUE #634: XIOS: use Domain for xy-plane and Axis for z
Currently, most XIOS integration tests use Axes for all coordinate directions. We're more likely to want to use a Domain for the $xy$-plane and an Axis for the $z$-axis because we can partition over a Domain with MPI but not an Axis.

jwallwork23:
# XIOS: use Domain for xy-plane and Axis for z
### Task List

---
# Change Description

I gather the intention is to parallelise nextSIM-DG in the horizontal domain. As such, it makes sense to set the XIOS tests up such that they use a Domain in the horizontal (which supports MPI parallelisation) and an Axis in the vertical ~~(which does not)~~ (see discussion below).

A minor piece of new functionality is that it's now possible to deduce the local sizes of a domain when setting the values i.e., you can call `setDomainLocalXValues` without having set `setDomainLocalXSize` (and similarly for Y).

---
# Test Description

Tests are updated as above, along with some other minor improvements:
* Use `{}` initialisation, as discussed in #554.
* Simplified domains.
* Purged all mention of latitude and longitude.
* More concise notation.
* Clearer comments on what's actually being set up in the tests.

---
### Pre-Request Checklist

### DIFF
--- a/core/test/XiosWrite_test.cpp
+++ b/core/test/XiosWrite_test.cpp
@@ -52,7 +52,7 @@ MPI_TEST_CASE("TestXiosWrite", 2)
     xios_handler.setCalendarStart(TimePoint("2023-03-17T17:11:00Z"));
     xios_handler.setCalendarTimestep(Duration("P0-0T01:30:00"));
 
-    // Domain setup
+    // Create a 4x2 horizontal domain with a partition halving the x-extent
     xios_handler.createDomain("xy_domain");
     xios_handler.setDomainType("xy_domain", "rectilinear");
     xios_handler.setDomainGlobalXSize("xy_domain", 4);
@@ -62,18 +62,18 @@ MPI_TEST_CASE("TestXiosWrite", 2)
     xios_handler.setDomainLocalXValues("xy_domain", { -1.0 + rank, -0.5 + rank });
     xios_handler.setDomainLocalYValues("xy_domain", { -1.0, 1.0 });
 
-    // Axis setup
+    // Create a vertical axis with 2 points
     xios_handler.createAxis("z_axis");
     xios_handler.setAxisValues("z_axis", { 0.0, 1.0 });
 
-    // Grid setup
+    // Create a 2D grid comprised of the xy-domain and a 3D grid which also includes the z-axis
     xios_handler.createGrid("grid_2D");
     xios_handler.gridAddDomain("grid_2D", "xy_domain");
     xios_handler.createGrid("grid_3D");
     xios_handler.gridAddDomain("grid_3D", "xy_domain");
     xios_handler.gridAddAxis("grid_3D", "z_axis");
 
-    // Field setup
+    // Create fields on the two grids
     xios_handler.createField("field_2D");
     xios_handler.setFieldOperation("field_2D", "instant");
     xios_handler.setFieldGridRef("field_2D", "grid_2D");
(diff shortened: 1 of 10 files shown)


core/test/XiosWrite_test.cpp
```diff
@@ -52,95 +52,81 @@ MPI_TEST_CASE("TestXiosWrite", 2)
...
+    ModelArray::setDimension(ModelArray::Dimension::X, nx);
+    ModelArray::setDimension(ModelArray::Dimension::Y, ny);
+    ModelArray::setDimension(ModelArray::Dimension::Z, nz);
     // Create some fake data to test writing methods
     HField field_2D(ModelArray::Type::H);
     field_2D.resize();
-    for (size_t j = 0; j < n2; ++j) {
-        for (size_t i = 0; i < n1; ++i) {
-            field_2D(i, j) = 1.0 * (i + n1 * j);
+    for (size_t j {}; j < ny; ++j) {
```
timspainNERSC, line 105: I think we should be explicit that `j` is being initialized to `0`, as in
```
for (size_t j {0}; j < ny; ++j) {…
```

timspainNERSC (commented):
I think we need to discuss using list initializers everywhere. It looks weird in a for loop, but may just require getting used to.

jwallwork23:
> I think we need to discuss using list initializers everywhere. It looks weird in a for loop, but may just require getting used to.

Sure, we can discuss in today's meeting.

jwallwork23:
I was wrong - MPI is supported for `Axis`, too. (Thanks @TomMelt for recalling that this might be the case, which prompted me to check.)

Thoughts on where to go from here:
* Whilst we could support MPI parallelism with `Axis` as well as `Domain`, too, this would mean unnecessary API wrapping and I feel we should keep that minimal. So we should support MPI parallelism via `Axis` or `Domain` but not both.
* Dropping the current implementation and switching to `Axis` is one option. However, (if I understand correctly) nextSIM-DG's parallelism is always in the $xy$-plane so it makes sense to expose parallelism in XIOS' `Domain` concept (which is inherently 2D) but not the $z$-axis.
* When I contacted the XIOS developers about a bug with `Axis`, Julien did say that people usually use the `Domain` concept for fields defined on a 2D plane.

Two arguments for dropping the currently implementation and switching to `Axis` only are that
1. I'm wrong about the MPI parallelism above and we want to do something more general.
2. We'd like to cut down on the number of XIOS constructs we're wrapping. To be fair, `Domain` does add more methods than the others.

So I recommend going ahead with this approach but would also be open to switching to `Axis`-only if preferred.

core/test/XiosWrite_test.cpp
```diff
@@ -52,83 +52,70 @@ MPI_TEST_CASE("TestXiosWrite", 2)
...
     Module::setImplementation<IStructure>("Nextsim::ParametricGrid");
-    ModelArray::setDimension(ModelArray::Dimension::X, n1);
-    ModelArray::setDimension(ModelArray::Dimension::Y, n2);
-    ModelArray::setDimension(ModelArray::Dimension::Z, n3);
+    const size_t nx { xios_handler.getDomainLocalXSize("xy_domain") };
+    const size_t ny { xios_handler.getDomainLocalYSize("xy_domain") };
+    const size_t nz { xios_handler.getAxisSize("z_axis") };
+    ModelArray::setDimension(ModelArray::Dimension::X, nx);
+    ModelArray::setDimension(ModelArray::Dimension::Y, ny);
+    ModelArray::setDimension(ModelArray::Dimension::Z, nz);
```
jwallwork23, line 101: These will need updating once #560 is merged.

timspainNERSC (approved):
Looks good.

PR #637, opened 2024-08-05
XIOS duration handling

### LINKED ISSUE #628: XIOS: Private methods for converting between strings and durations
We already have private methods for converting between strings and dates:
```
std::string convertXiosDatetimeToString(const cxios_date datetime, const bool isoFormat = true);
cxios_date convertStringToXiosDatetime(const std::string datetime, const bool isoFormat = true);
```

We should implement methods for converting between strings and durations, too.

### LINKED ISSUE #636: Durations should not accept months
Currently, `Duration("P0-1-0T0:0:0")`, i.e., a 1-month duration, does not throw an error. Converting this duration to seconds gives some arbitrary number (different every time).

Of course, this is not the intended duration format. However, XIOS does accept durations in this format, so we should be careful when passing these around.

XIOS always sets the number of days in a month to 30, which could cause a lot of confusion. It would be better to just disallow such formats inside `Duration`.

jwallwork23:
# XIOS duration handling

Closes #628
Closes #636

### Task List

---
# Change Description

This PR addresses a few minor concerns:
* XIOS interface methods related to output frequency and split frequency now work with nextSIM-DG's `Duration` objects, rather than accepting XIOS' duration strings, which differ in confusing ways.
* Implement private methods for converting between XIOS and nextSIM-DG's time duration formats.
* Implement a private method for converting between C and C++ string formats, since this is done repeatedly in the XIOS interface.
* Disallow `"PY-M-DTH:M:S"`-style input to `Duration`. Namely, do not allow months to be specified because they have arbitrary length and the convention does not align with XIOS, which could cause confusion.

---
# Test Description

Unit tests are updated accordingly. I also thought it would be a good idea to test more rigorously that durations with different units are treated appropriately.

---
### Pre-Request Checklist

### DIFF
--- a/core/test/XiosFile_test.cpp
+++ b/core/test/XiosFile_test.cpp
@@ -77,10 +77,10 @@ MPI_TEST_CASE("TestXiosFile", 2)
     REQUIRE(xios_handler.getFileType(fileId) == fileType);
     // Output frequency
     xios_handler.setFileOutputFreq(fileId, timestep);
-    REQUIRE(xios_handler.getFileOutputFreq(fileId).seconds() == Approx(1.5 * 60 * 60));
+    REQUIRE(xios_handler.getFileOutputFreq(fileId).seconds() == 1.5 * 60 * 60);
     // Split frequency
     xios_handler.setFileSplitFreq(fileId, timestep);
-    REQUIRE(xios_handler.getFileSplitFreq(fileId).seconds() == Approx(1.5 * 60 * 60));
+    REQUIRE(xios_handler.getFileSplitFreq(fileId).seconds() == 1.5 * 60 * 60);
     // Add field
     xios_handler.fileAddField(fileId, "field_A");
     std::vector<std::string> fieldIds = xios_handler.fileGetFieldIds(fileId);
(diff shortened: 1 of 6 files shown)


core/test/XiosFile_test.cpp
```diff
@@ -73,19 +76,46 @@ MPI_TEST_CASE("TestXiosFile", 2)
...
+    xios_handler.setFileOutputFreq("minute", Duration("P0-0T00:01:00"));
+    xios_handler.setFileSplitFreq("minute", Duration("P0-0T00:02:00"));
+    REQUIRE(xios_handler.getFileOutputFreq("minute").seconds() == Approx(60));
+    REQUIRE(xios_handler.getFileSplitFreq("minute").seconds() == Approx(2 * 60));
+    xios_handler.createFile("second");
+    xios_handler.setFileOutputFreq("second", Duration("P0-0T00:00:01"));
+    xios_handler.setFileSplitFreq("second", Duration("P0-0T00:00:02"));
+    REQUIRE(xios_handler.getFileOutputFreq("second").seconds() == Approx(1));
+    REQUIRE(xios_handler.getFileSplitFreq("second").seconds() == Approx(2));
+
```
timspainNERSC, line 118: I know that comparing floating point values can be tricky, but in this case, the values should be exact. If the duration is being set to exactly one second, and the returned floating point value is not the exact value 1, this should also be a failure, and so on for all the othre exact integer values of seconds.

timspainNERSC (changes requested):
Looks good, but I think the use of `doctest::Approx()` should be reviewed.

core/test/XiosFile_test.cpp
jwallwork23: Thanks @timspainNERSC, good point. This should be fixed by 71ff796.

timspainNERSC (approved):
All good

PR #641, opened 2024-08-19
Issue459 oasis ocean coupling

einola:
# Pull Request Title
### Task List

---
# Change Description

*Please write a summary of the change and specify which Issue this fixes. This should cover the motivation for making the change and provide almost enough context to review the change in isolation. While the issue description is a good place to discuss the failure/requirements, this is the ideal space for explaining why this design fixes the issue and why you designed it the way you did.*

- *If there are important design decisions, particularly those that leverage functionality outside this change or were caused by existing limitations, they should be noted here.*
- *If there is a design specification which relates to this PR, please provide a link to it here.*
- *If the PR relates to an open issue, please link the issue to the PR by using the 'closes #issue_number' syntax.*
- *If the PR relates to several open issue, make sure to use 'closes #issue_1, closes #issue_2, etc.' syntax.*
- *If there are any dependencies (third-party or internal downstream) which are added or impacted, please note them here. And if their limitations informed your design, please note how they did so here (particularly if this information was not captured in a design specification).*

Text goes here

---
# Test Description

*Please describe the tests you have written that verify the enhancement works or that the bug fix has been remedied. For a bug fix, please ensure a complete description of the failure is available here or in the issue such that a reviewer can verify that the testing is suitable. This is a good place to highlight any manual testing you may have had to perform.*

*Please describe any operating system limitations placed on the testing (if appropriate).*

*If your tests fixes any issue, please explain that in the comment you provided in the test file by referencing the issue within the comment and please record the mapping here e.g.*
- *testA fixes issue 404*
- *testB fixes issue 999*

Text goes here

---
# Documentation Impact

*Please describe any changes to the documentation not captured above or, if made in a seperate pull request, please link to the corresponding PR or issue.*

Text goes here

---
# Other Details

*If you have run any static analysis, including complexity analysis, or run coverage testing which has not been captured by an automated tool, please link or copy it here.*

Text goes here

---
### Pre-Request Checklist

*Please complete your pull request description and delete the instructional text before submitting.*

*Congratulations and thank you for making your contribution to neXtSIM_DG!*


einola:
I set up this draft PR to help @andreapiacentini and me figure out how to get the OASIS coupling with NEMO working.

There's still a good deal of work left ...

core/src/main.cpp
```diff
@@ -9,6 +9,9 @@
 #ifdef USE_MPI
 #include <mpi.h>
 #endif
+#ifdef USE_OASIS
```
andreapiacentini, line 10: In the CMakeLists.txt you need to add close to the end
```
if (ENABLE_OASIS)
   target_include_directories(nextsim PRIVATE ${OASIS_INCLUDES})
endif ()
```

core/src/include/ModelMetadata.hpp
```diff
@@ -18,6 +18,9 @@
 #ifdef USE_MPI
 #include <mpi.h>
 #endif
+#ifdef USE_OASIS
```
andreapiacentini: The inclusion of OASIS in a `.hpp` included by main programs, require to add for many `XXX` targets in `core/test/CMakeLists.txt` and in `physics/test/CMakeLists.txt` a bunch of sections like  
```
if (ENABLE_OASIS)
   target_include_directories(XXX PRIVATE ${OASIS_INCLUDES})
endif ()
```

core/src/main.cpp
```diff
@@ -20,7 +23,18 @@
 int main(int argc, char* argv[])
 {
 #ifdef USE_MPI
```
andreapiacentini, line 24: The configuration parser is invoked by all the processes but does not perform any parallel communication and, therefore, does not need any knowledge of the MPI communicator.  
Would it be possible to postpone the MPI initialisation after the Configuration parsing so that the choice between `MPI_Init` and `oasis_c_init_comp` is not only depending on `USE_OASIS` but also on the activation of modules triggering OASIS ?

core/src/main.cpp
einola: It already says
```
target_include_directories(nextsimlib PRIVATE ${OASIS_INCLUDES})
```
at line 159 (within an ``if (ENABLE_OASIS) ... endif ()`` clause). Is that not working?

NB the directory here is ``nextsimlib``, not ``nextsim``, as in your example.

core/src/main.cpp
einola: I like this, as it also makes for cleaner code, in my opinion. See [d266c17](https://github.com/nextsimhub/nextsimdg/pull/641/commits/d266c17fa582f0635fd609a36244149be25364ef). I'm not sure if or how we can make the configuration decide if oasis is used, but the option is now there.

core/src/main.cpp
andreapiacentini: When I compiled on my machine, line 159 lead ok to the compilation of the shared lib nextsim.  But `main.cpp` (nor any of the test mains) is not compiled with the lib.  
When compiling the `nextsim` executable target, `main.cpp` is compiled just before being linked with the libs. And `main.cpp` or all the mains that include `ModelMetadata.hpp` now include `oasis_c.h` and do not find it. Adding these lines I got to the end of the compilation but it's "trial and error cmake". Please find a better solution, but I can confirm that something more than line 159 is needed. Also, I don't know at all what `PRIVATE` stands for. There is maybe a different keyword for making the include property apply to depending mains.

einola: That's strange because the whole point of the library was to compile it once and then link the main and tests to it. Lines 190 and 191 should take care of that by telling cmake that the main executable requires ``nextsimlib``. The library is also included in the ``target_link_libraries`` command in the ``CMakeLists.txt`` files in the ``test`` subdirectories.

Can you open a separate issue (or use the discussions board: https://github.com/nextsimhub/nextsimdg/discussions), including all the output etc?

andreapiacentini: Before opening an issue (I think a discussion would be better), I just point out that the in the `core/test/CMakeLists.txt` and `physics/test/CMakeLists.txt` there are several `target_include_directories` for the different executables. That's where I draw inspiration. Have a look at time. I guess that the `target_link_libraries` is ok for finding objects at link time, but the `target_include_directories` is needed at compile time.

andreapiacentini: I gave a try with `PUBLIC` instead of `PRIVATE` in the `target_include_directory` for `nextsimlib` and now it works for the mains as well. Are you happy with it (and will change it) or do I open the discussion for something explaining in detail how the properties inheritance works in cmake ?

einola: @Thanduriel explained to me why ``PUBLIC`` is the right choice, so I'll change it. It's a bit strange that it works on my setup, but it doesn't matter.

core/src/include/ModelMetadata.hpp
einola: Is this working better now that I put ``PUBLIC`` in the root ``CMakeLists.txt``? I don't have any problems here.

andreapiacentini: It's working allright with only two other changes:

1. Rename `Findoasis.cmake` to `FindOASIS.cmake` (my version seems to be case sensitive)
2. Replace `${OASIS_DIR}/inc` by `${OASIS_DIR}/include` in it

einola: Updated with the latest commit

1. I don't quite understand why the case matters, but my setup appears to be case-insensitive
2. It should have been ``include`` all along. It's a bit of a mystery why it worked for me the way it was!

physics/test/OASISCoupledOcean_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+    ModelArray::setDimensions(ModelArray::Type::Z, { 1, 1, 1 });
+
+    HField cice(ModelArray::Type::H);
+    cice = 1.0;
+    ModelComponent::getStore().registerArray(Protected::C_ICE, &cice, RO);
+    OASISCoupledOcean ocpl;
+    ModelMetadata metadata;
+
+    ocpl.setData(ModelState::DataMap());
+    ocpl.setMetadata(metadata);
```
andreapiacentini, line 76: The overridden `setMetadata` defines variables, yet it skips the `oasis_c_enddef` that should be issued by a generic call after all boundaries (ocean and or atmosphere) have finished defining variables.  
I added
```   
 OASIS_CHECK_ERR(oasis_c_enddef());
```

physics/test/OASISCoupledOcean_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+    ModelArray::setDimensions(ModelArray::Type::H, { 1, 1 });
+    ModelArray::setDimensions(ModelArray::Type::Z, { 1, 1, 1 });
+
+    HField cice(ModelArray::Type::H);
+    cice = 1.0;
+    ModelComponent::getStore().registerArray(Protected::C_ICE, &cice, RO);
+    OASISCoupledOcean ocpl;
+    ModelMetadata metadata;
+
```
andreapiacentini, line 73: I had to add here
```
    const std::vector<int> partInfo = { OASIS_Serial, 1 };
    OASIS_CHECK_ERR(oasis_c_def_partition(
        &metadata.OASISPartitionId, OASIS_Serial_Params, &partInfo[0], OASIS_No_Gsize, compName.c_str()));
```

physics/test/OASISCoupledOcean_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+    ocpl.setData(ModelState::DataMap());
```
andreapiacentini, line 74: I had to add a call to `configure` for getting the keys  
```
    ocpl.configure();
```
and inside `configure` from `OASISCoupledOcean.cpp` I had to update the string lists:  
```
    firstLayerDepth = Configured::getConfiguration(layerDepthConfigKey, FIRST_LAYER_DEPTH);

    cplStringsIn
      = { SSTKey, SSSKey, UOceanKey, VOceanKey, SSHKey };
    if (Configured::getConfiguration(exchangeFirstLayerConfigKey, EXCHANGE_FIRST_LAYER)) {
        cplStringsIn.push_back(MLDKey);
    }
    cplStringsOut
      = { TauXKey, TauYKey, TauModKey, EMPKey, QNoSunKey, QSWKey, SFluxKey, CIceKey };
```

physics/test/OASISCoupledOcean_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+    ocpl.setMetadata(metadata);
+    ocpl.updateBefore(TimestepTime());
```
andreapiacentini: There is a mismatch in `updateBefore` where `VOceanKey` is used twice.  
With it fixed, I had a first running test !

PR #644, opened 2024-08-19
Make the advected field type array non-const

### LINKED ISSUE #643: Make the advected field type array non-const
The `DynamicsKernel::fieldType` map is decalred const. That's not very useful for a member variable that should be read and written.

Should also fix the deleted constructor problem for the Intel compiler.

timspainNERSC:
# Make the advected field type array non-const
## Fixes #643 

The DynamicsKernel::fieldType map is decalred const. That's not very useful for a member variable that should be read and written.

Should also fix the deleted constructor problem for the Intel compiler.


TomMelt (approved):
Thanks @timspainNERSC , looks good to me.

PR #660, opened 2024-08-22
Add pre-commit for automatically updating date stamps

### LINKED ISSUE #658: Pre-commit hook for updating date stamp
This isn't something developers should have to remember to update themselves.

jwallwork23:
# Add pre-commit for automatically updating date stamps

Closes #658.

Works for me but worth please check this works for you too.

### DIFF
--- a/.pre-commit
+++ b/.pre-commit
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
 
 # An example pre-commit file for nextSIM-DG development
 
--- a/.pre-commit
+++ b/.pre-commit
@@ -6,5 +6,8 @@ for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
   # Apply clang-format for linting
   clang-format -i ${FILE} --verbose
   # Update date stamp to today's date
-  sed -i "s/$(grep '\* @date' ${FILE})/ @date $(date '+%d %b %Y')/" ${FILE}
+  LINE=$(grep '@date' ${FILE})
+  let SPACES=$(echo "${LINE}" | tr -cd ' \t' | wc -c)-4
+  NEWLINE=$(printf "%s%$((SPACES))s%s\n" " @date" " " "$(date '+%d %b %Y')")
+  sed -i "s/${LINE}/${NEWLINE}/" ${FILE}
 done
--- a/.pre-commit
+++ b/.pre-commit
@@ -1,0 +1,10 @@
+#!/usr/bin/bash
+
+# An example pre-commit file for nextSIM-DG development
+
+for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
+  # Apply clang-format for linting
+  clang-format -i ${FILE} --verbose
+  # Update date stamp to today's date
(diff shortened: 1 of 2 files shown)


einola:
It kind of works:
 1. You need GNU sed. BSD sed doesn't work.
 2. It modifies your files but just commits the parts I've already ``git add``-ed.

I don't know if we want to make a fuss about 1. Can this be done with plain BSD sed?

The behaviour in 2 can be considered a feature. What I added is committed, nothing more. But the hook kindly suggests how I should modify my files so I can review them and then add them. I guess this is normal behaviour, but it was unexpected to me, not having used pre-commit hooks before.

.pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/bash
```
einola:
```suggestion
#!/bin/bash
```

At least that's where it lives on the Mac. But this path should also work for Linux, right?

einola (approved):
Can you check if using /bin/bash instead of /usr/bin/bash works on linux? Other than that, it's good.

joewallwork:
> Can you check if using /bin/bash instead of /usr/bin/bash works on linux? Other than that, it's good.

Ah yes, I always forget about this. Works fine on Linux yeah. Thanks!

joewallwork:
@einola did the spaces come out okay for you? I had to add a few extra spaces to get things to align nicely. I'm currently using
```
#!/bin/bash

# An example pre-commit file for nextSIM-DG development

for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
  # Apply clang-format for linting
  clang-format -i ${FILE} --verbose
  # Update date stamp to today's date
  sed -i "s/$(grep '\* @date' ${FILE})/ @date    $(date '+%d %b %Y')/" ${FILE}
done
```

einola:
I didn't check the alignment :) ... but feel free to change it before merging.

joewallwork:
Okay I *think* the latest commit should preserve spaces. Do the additional commands work on Mac, too?

.pre-commit
```diff
@@ -6,5 +6,8 @@ for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
   # Apply clang-format for linting
   clang-format -i ${FILE} --verbose
   # Update date stamp to today's date
-  sed -i "s/$(grep '\* @date' ${FILE})/ @date $(date '+%d %b %Y')/" ${FILE}
+  LINE=$(grep '@date' ${FILE})
+  let SPACES=$(echo "${LINE}" | tr -cd ' \t' | wc -c)-4
+  NEWLINE=$(printf "%s%$((SPACES))s%s\n" " @date" " " "$(date '+%d %b %Y')")
+  sed -i "s/${LINE}/${NEWLINE}/" ${FILE}
```
einola:
```suggestion
  sed -i~ "s/${LINE}/${NEWLINE}/" ${FILE}
```

I am unsure why, but mac sed freaks out if you don't back up. The error message is not comprehensible to me "sed: 1: "physics/src/modules/inc ...": extra characters at the end of p command"

But with the backup option, everything runs fine, and backups are good, so I guess it's okay!

joewallwork: Oh okay thanks! Works on Ubuntu too so should be all good.

einola (approved):

PR #663, opened 2024-08-23
WIP: MPI-ify the dynamics

### LINKED ISSUE #120: MPI parallelization of Dynamics

TomMelt:
This PR has been superseded by #744

Fixes #120

### Change Description

To parallelize the dynamics we first need the halo region information for each process.

This information is located in the metadata file produced by the domain `decomp` tool.

This PR updates `ModelMetadata::getPartitionMetadata` with all the necessary information on halo regions.

### Test Description

This PR adds two new tests to check the functionality of the `getPartitionMetadata` method and the halo exchange.
- [x] `ModelMetadata_test.cpp` - test `getPartitionMetadata` can correctly load metadata from a partition metadata file into the appropriate data structure e.g., `neighbourExtents`, `neighbourHaloStarts` etc.
- [x] `HaloExchange_test.cpp` - check that we can do basic halo exchange by making use of the new slicing utilities e.g., `ModelArraySlice` #744 
- [ ] add halo exchange logic to nextsim

### Pre-Request Checklist

- [x] Add methods to read halo exchange info into metadata class
- [ ] Add halo exchange info to DGModelArray class
- [ ] Add exchange logic to DGModelArray.hpp
- [ ] Add option of in-/out-flows in the halo? or should this be separate?

### DIFF
--- a/core/src/ModelMetadata.cpp
+++ b/core/src/ModelMetadata.cpp
@@ -1,8 +1,9 @@
 /*!
  * @file ModelMetadata.cpp
  *
- * @date 21 August 2024
+ * @date 17 Jan 2025
  * @author Tim Spain <timothy.spain@nersc.no>
+ * @author Tom Meltzer <tdm39@cam.ac.uk>
  */
 
 #include "include/ModelMetadata.hpp"
@@ -47,8 +48,8 @@ void ModelMetadata::readNeighbourData(netCDF::NcFile& ncFile)
     netCDF::NcGroup neighbourGroup(ncFile.getGroup(neighbourName));
     std::string varName {};
     for (auto edge : edges) {
-        size_t nStart {}; // start point in metadata arrays
-        size_t count {}; // number of elements to read from metadata arrays
+        size_t nStart = 0; // start point in metadata arrays
+        size_t count = 0; // number of elements to read from metadata arrays
         std::vector<int> numNeighbours = std::vector<int>(mpiSize, 0);
         std::vector<int> offsets = std::vector<int>(mpiSize, 0);
 
@@ -66,7 +67,8 @@ void ModelMetadata::readNeighbourData(netCDF::NcFile& ncFile)
             // initialize neighbour info to zero and correct size
             neighbourRanks[edge].resize(count, 0);
             neighbourExtents[edge].resize(count, 0);
-            neighbourHaloStarts[edge].resize(count, 0);
+            neighbourHaloSend[edge].resize(count, 0);
+            neighbourHaloRecv[edge].resize(count, 0);
 
             varName = edgeNames[edge] + "_neighbour_ids";
             neighbourGroup.getVar(varName).getVar({ nStart }, { count }, &neighbourRanks[edge][0]);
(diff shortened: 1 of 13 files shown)


core/test/ModelMetadata_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+        REQUIRE(metadata.neighbourRanks[ModelMetadata::LEFT] == std::vector<int> { 0, 1 });
+        REQUIRE(metadata.neighbourExtents[ModelMetadata::LEFT] == std::vector<int> { 4, 5 });
+        REQUIRE(metadata.neighbourHaloStarts[ModelMetadata::LEFT] == std::vector<int> { 6, 6 });
+        REQUIRE(metadata.neighbourRanks[ModelMetadata::RIGHT].size() == 0);
+        REQUIRE(metadata.neighbourRanks[ModelMetadata::BOTTOM].size() == 0);
+        REQUIRE(metadata.neighbourRanks[ModelMetadata::TOP].size() == 0);
+    } else {
+        std::cerr << "only valid for 3 ranks" << std::endl;
+        exit(1);
+    }
```
joewallwork: Could this common part of the tests be pulled out into a function that's called in both places? It'd avoid duplication and reduce the likelihood of human errors when copy/pasting and reviewing.

core/test/ModelMetadata_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    // this metadata is specific to the periodic boundary conditions
+    if (test_rank == 0) {
+        // clang-format off
```
joewallwork: I'm guessing this is because of line length? Might be sufficient to put `using namespace std;` at the top.

core/test/ModelMetadata_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include <doctest/extensions/doctest_mpi.h>
+#include <iostream>
+
+#include "ModelMetadata.hpp"
+
+const std::string testFilesDir = TEST_FILES_DIR;
+const std::string filename = testFilesDir + "/paraGrid_test.nc";
+const std::string diagFile = "paraGrid_diag.nc";
+const std::string dateString = "2000-01-01T00:00:00Z";
```
joewallwork: Unused?

joewallwork (changes requested):
Looks great, thanks @TomMelt. I drew a diagram and convinced myself that the numbers in the CDL files are consistent with the expected values in the tests. I have a couple of suggestions, the main one being a bit of refactoring to avoid duplication.

core/src/ModelMetadata.cpp
```diff
@@ -39,6 +42,75 @@ void ModelMetadata::setMpiMetadata(MPI_Comm comm)
     MPI_Comm_rank(mpiComm, &mpiMyRank);
 }
 
+void ModelMetadata::readNeighbourData(netCDF::NcFile& ncFile)
+{
+    netCDF::NcGroup neighbourGroup(ncFile.getGroup(neighbourName));
+    std::string varName {};
+    for (auto edge : edges) {
+        size_t nStart {}; // start point in metadata arrays
+        size_t count {}; // number of elements to read from metadata arrays
```
timspainNERSC: Initialize numeric types using `= 0`, if indeed that is necessary (`nStart` and `count` seem to be set to other values before being read from).

core/test/ModelMetadata_test.cpp
timspainNERSC: Or define a `typedef` (or `using`)

timspainNERSC (changes requested):
A couple of minor edits.

timspainNERSC:
Having looked over the changes to ModelMetadata.cpp again, I'm not sure this code should be in there. Perhaps even the grid metadata should be moved out. You don't have to stuff the MPI metadata into existing classes.

Perhaps we should meet to discuss the architecture of the MPI data now that we have more of a handle on what it is going to look like. Possibly including moving the existing coordinate data and functions that are currently in ModelMetadata.

TomMelt:
> Having looked over the changes to ModelMetadata.cpp again, I'm not sure this code should be in there. Perhaps even the grid metadata should be moved out. You don't have to stuff the MPI metadata into existing classes.
> 
> Perhaps we should meet to discuss the architecture of the MPI data now that we have more of a handle on what it is going to look like. Possibly including moving the existing coordinate data and functions that are currently in ModelMetadata.

@timspainNERSC , yeah that makes sense. I will continue with my halo exchange tests for now and we can sort this before it merges into `develop`

TomMelt:
This PR has been superseded by #744

PR #664, opened 2024-08-26
Common ice-ocean stress calculation

### LINKED ISSUE #649: Coupling routines need to receive ice-ocean stresses
We need to send the ice-ocean stress to the ocean via the coupler. At the moment, these are calculated implicitly in the dynamics but not accessible directly or outside them.

timspainNERSC:
# Common ice-ocean stress calculation
Fixes #649

Moves the ice ocean stress to a common routine, with some necessary finagling with the parameter objects.

### DIFF
--- a/dynamics/src/include/DynamicsKernel.hpp
+++ b/dynamics/src/include/DynamicsKernel.hpp
@@ -9,6 +9,7 @@
 #define DYNAMICSKERNEL_HPP
 
 #include "DGTransport.hpp"
+#include "DynamicsParameters.hpp"
 #include "Interpolations.hpp"
 #include "ParametricMesh.hpp"
 #include "ParametricTools.hpp"
@@ -32,14 +33,20 @@
 namespace Nextsim {
 
 // forward define the class holding the potentially non-DG parts
-template <int DGdegree> class DynamicsInternals;
+template<int DGdegree> class DynamicsInternals;
 
-template <int DGadvection, int DGstress> class DynamicsKernel {
+template<int DGadvection, int DGstress> class DynamicsKernel {
 public:
     typedef std::pair<const std::string, const DGVector<DGadvection>&> DataMapping;
     typedef std::map<typename DataMapping::first_type, typename DataMapping::second_type> DataMap;
 
-    DynamicsKernel() = default;
+    DynamicsKernel(double cosOceanAngleIn, double sinOceanAngleIn, const DynamicsParameters& paramsIn)
+        : cosOceanAngle(cosOceanAngleIn)
+        , sinOceanAngle(sinOceanAngleIn)
+        , baseParams(paramsIn)
+        , deltaT(0.)
+    {
+    }
     virtual ~DynamicsKernel() = default;
     virtual void initialise(const ModelArray& coords, bool isSpherical, const ModelArray& mask)
     {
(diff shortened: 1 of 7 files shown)


dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -32,14 +33,20 @@
...
 public:
     typedef std::pair<const std::string, const DGVector<DGadvection>&> DataMapping;
     typedef std::map<typename DataMapping::first_type, typename DataMapping::second_type> DataMap;
 
-    DynamicsKernel() = default;
+    DynamicsKernel(double cosOceanAngleIn, double sinOceanAngleIn, const DynamicsParameters& paramsIn)
+        : cosOceanAngle(cosOceanAngleIn)
+        , sinOceanAngle(sinOceanAngleIn)
+        , baseParams(paramsIn)
+        , deltaT(0.)
```
einola, line 47: Since ocean_turning_angle is always in DynamicsParameters, wouldn't 

```suggestion
    DynamicsKernel(const DynamicsParameters& paramsIn)
        : cosOceanAngle(std::cos(paramsIn.ocean_turning_angle))
        , sinOceanAngle(std::sin(paramsIn.ocean_turning_angle))
        , baseParams(paramsIn)
        , deltaT(0.)
```

work? This gives fewer arguments for DynamicsKernel and the derived classes.

scripts/module_builder.py
```diff
@@ -101,6 +101,8 @@ def write_source_file(source, config, strings):
 {{
     addToConfiguredModules<{strings[class_name]}, {strings[module_class_name]}>();
 }}
+
+template class {module_templ};
```
einola, line 105: Does this belong in this PR?

einola (approved):
Very nice. One question and one suggestion, which you can do with what you like.

scripts/module_builder.py
timspainNERSC: Yes, if we want the Mac to build to tests successfully on Release. And it will be gone as soon as you rebase/merge from `develop` :P

dynamics/src/include/DynamicsKernel.hpp
timspainNERSC: Not so long as `VPCGDynamicsKernel` assigns fixed values to the cos and sin terms, as in
```
    VPCGDynamicsKernel(StressUpdateStep<DGadvection, DGstressComp>& stressStepIn,
        const DynamicsParameters& paramsIn)
        // Note that the ocean turning angle is explicitly zero at present
        : CGDynamicsKernel<DGadvection>(1., 0., paramsIn, u, v)
```
There's no other way (that I know of) to ensure these values are set. Once the `VPCGDynamicsKernel` calculation of stress correctly handles the ocean turning angle, then what you suggest can be done.

scripts/module_builder.py
einola: Ah, I see :)

dynamics/src/include/DynamicsKernel.hpp
einola: Ok, I see. It's not difficult to add ocean turning into ``VPCGDynamicsKernel``, but that should probably be a different issue. I'll make one and merge this.

PR #665, opened 2024-08-26
Reverse the order of dynamics and thermodynamics

### LINKED ISSUE #654: Make DG fields in the restart file self-consistent
Currently, the thermodynamics runs after the dynamics. Since the thermodynamics changes the mean, DG0 value of prognostic fields, this can mean that the higher DG components could be such that the value of the field exceeds its physical limits. To be fully consistent the final state of fields with both DG components and geophysical limits should be immediately after the dynamics, when the limiters have just been applied.

The solution to this would be to swap the order of the dynamics and thermodynamics within the `PrognosticData::update()`.

timspainNERSC:
# Reverse the order of dynamics and thermodynamics

Fixes #654

---
# Change Description

Currently, the thermodynamics runs after the dynamics. Since the thermodynamics changes the mean, DG0 value of prognostic fields, this can mean that the higher DG components could be such that the value of the field exceeds its physical limits. To be fully consistent the final state of fields with both DG components and geophysical limits should be immediately after the dynamics, when the limiters have just been applied.

The solution to this is to swap the order of the dynamics and thermodynamics within the `PrognosticData::update()`, as done in the PR.

---
# Test Description

Running the updated model through January 2010 doesn't look any *worse* than before. Arguably it is more realistic.

### DIFF
--- a/core/src/PrognosticData.cpp
+++ b/core/src/PrognosticData.cpp
@@ -81,16 +81,13 @@ void PrognosticData::update(const TimestepTime& tst)
     pOcnBdy->updateBefore(tst);
     pAtmBdy->update(tst);
 
-    // Fill the values of the true ice and snow thicknesses.
-    iceGrowth.initializeThicknesses();
-    // Fill the updated ice temperature array
-    ticeUpd.data().setData(m_tice);
-    pDynamics->update(tst);
-    updatePrognosticFields();
-
     // Take the updated values of the true ice and snow thicknesses, and reset hice0 and hsnow0
     // IceGrowth updates its own fields during update
     iceGrowth.update(tst);
+    updatePrognosticFields();
+
+    pDynamics->update(tst);
+
     updatePrognosticFields();
 
     pOcnBdy->updateAfter(tst);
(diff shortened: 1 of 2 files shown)


core/src/PrognosticData.cpp
```diff
@@ -81,18 +81,15 @@ void PrognosticData::update(const TimestepTime& tst)
     pOcnBdy->updateBefore(tst);
     pAtmBdy->update(tst);
 
-    // Fill the values of the true ice and snow thicknesses.
-    iceGrowth.initializeThicknesses();
-    // Fill the updated ice temperature array
-    ticeUpd.data().setData(m_tice);
```
einola, line 87: Were those not needed after all?

timspainNERSC: These two calls were just to fill arrays the the dynamics will read and change. The thermodynamics fills these arrays and so they do not need explicitly filling by hand in `PrognosticData::update()`.

`iceGrowth.initializeThicknesses()` is the first thing called by `IceGrowth::update()` (`iceGrowth.cpp` L 147), so it only needed to be called before updating the dynamics.

The updated ice temperature array `ticeUpd` is the one where the thermodynamic changes to the ice growth are written (but not read), so it doesn't need filling before the dynamics, either.

einola: Nice. That makes this change even more elegant ;)

einola (approved):

PR #668, opened 2024-08-28
Advection of arbitrary fields

### LINKED ISSUE #99: Advecting arbitrary fields
A component may depend on some field that needs to be advected. Some infrastructure (in ModelComponent?) should gather all the the fields to be advected from all the loaded Components, and provide them to the dynamics so that they can be.

### LINKED ISSUE #337: Implementation fields for advection
Individual Module implementations should be able to designate fields that are (otherwise) internal to that particular implementation to be advected by the dynamics. This should be possible without requiring the field be defined in the interface class and thereby polluting all the other implementations.

timspainNERSC:
# Advection of arbitrary fields

Fixes #99 & fixes #337

---
# Change Description

* Adds advection of the two non-dynamical but prognostic fields, snow thickness and ice temperature.
* Adds functions to advect an HField `ModelArray` for one advection time step. This is a wrapper around the `IDynamics::advectField()` function which will advect a 2D or 3D `ModelArray`, using the slicing infrastructure to reduce the 3D array to a series of 2D fields, advect them using the 2D advection functions provided by the kernel, and reconstruct the 3D field, again using `ModelArraySlice`s.
* Adds `DynamicsKernel::advectField()` which will interface with the dynamics code to advect a 2D `ModelArray` field by one advection time step.
* Adds a `AdvectionDynamicsKernel` to the dynamics, which just advects ice fields according to the ocean velocity.
* Initializes many fields to zero in the dynamics kernels to ensure there is no junk data in the higher DG components.
* Does *not* zero out the higher DG components every time new DG0 data is copied into the dynamics from the thermodynamics.

Incorporates the slicing changes (#736).

---
# Test Description

Adds two similar tests, `testFieldAdvection` in `dynamics/test` and `testDynamics` in `core/test`. These both set up a 51 x 51 domain, create a circular region of ice in the centre with solid body rotation as the velocity field. A snow depth field that is sinusoidal with angle around the centre is then advected and the field tested every time that the soldi body rotation field makes a quarter turn. The test is an 8 sample approximation to the Fourier transform for the degree 1 Fourier coefficients. That is, it extracts the amplitude of the sinusoidal and cosinusoidal variation around the disk. The two components are then compared to the original sinusoid field with appropriate phase shifts to ensure the field is rotating as a solid body.

In the case of the test in `core/test` a similar ice temperature 3D field is also advected and tested in the same manner.


core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -135,6 +133,8 @@ void BBMDynamics::update(const TimestepTime& tst)
 
     kernel.update(tst);
 
+    advectHField(hsnow, hsnowName);
```
timspainNERSC: Check this is needed here (cf. PrognosticData.cpp:96)

timspainNERSC: It is not

timspainNERSC:
Obsoleted by PR #877

Issue #614, opened 2024-07-16
incorrect resource lifetimes due to static data


Thanduriel:
Currently significant parts of the model are constructed and managed through `ConfiguredModule::configuredModules` and therefore have static lifetimes. This makes resource management via RAII challenging since the order of init/release is hard to track and destructors will only be called during the epilogue of `main`.

A concrete example for a problem this causes can be found in the GPU code. Usage of the GPU requires both manual `initialize` and `finalize` calls and the management of memory buffers. Currently the simulation will always terminate with an exception because the buffers are freed by a destructor which is only called after the `finalize`.

The easiest solution would probably be to make the `ConfiguredModule` data non-static and to introduce a finalize call.

timspainNERSC:
A greater problem is the static data at the foundation of every `Module<>`.

Also check the rest of the infrastructure in `core/src/include/*.hpp` for static data members.

timspainNERSC:
Classes to look at:

- [x] `ConfigMap`
- [x] `Configurator`
- [x] `Configured<>`
- [x] `ConfiguredModule`
- [x] `FileCallbackCloser`
- [x] `Iterator`
- [x] `MissingData`
- [x] `ModelComponent`
- [x] `ModelConfig`
- [x] `Module<>`
- [x] `ParaGridIO`
- [x] `ScopedTimer`
- [x] `Timer`

PR #674, opened 2024-09-03
Resolving static lifetime problems

timspainNERSC:
# Resolving static lifetime problems

Fixes #614 

---
# Change Description

## Deleted classes
- `ScopedTimer`
- `Timer`
    * Removed these classes as they are currently unused and need to be heavily refactored to be thread- and possibly MPI-aware.

### `ConfigMap`
Changed the `static const` type constants to an `enum`.

### `Configurator`
Made the static data into singletonoid functions. The data in question are:

   * The list of configuration sources
   * The true command line
   * The pointer to the `AdditionalConfiguration` object

The command line values were combined into a struct to avoid passing more than one return value.
The pointer to the additional configuration was set to a `nullptr` initially, with null checks used when the pointer value is used.

### `Configured` class template
Nothing depends on the keyMap template member variable being a template member variable. Redefine each key map as `static` (compilation unit localized) `const std::map`. Since many classes are configured, this affects many files. Only the source `.cpp` files are affected.

#### Changes to `Module<>`

* Make `Module` independent of Nextsim. It is now a separate Module base class that can be used for any C++ project.
* Introduce a `NextsimModule` class that specializes the newly generic `Module` for nextSIM-DG.
* Replace `#include "Module.hpp"` with `#include "NextsimModule.hpp"`.
* Add a map from interface name to implementation name in the `Module` namespace alongside the `Module` template class. This string-to-string map is not available elsewhere. The data is stored as a function-local `static` variable.
* Remove the class-static instance of the reference implementation from the `Module<>` template. This static instance is now a function-local `static` variable, so at least the construction order is well defined. The lifetime is still until the executable epilogue.
* Delete the module classes. Where there was originally (for example) a `DiagnosticOutputModule` class that inherited from `Module<IDiagnosticModule>`, everything now refers to the template instantiation. The `…Module` class was necessary to run the register the module at static start-up. With this no longer necessary,
  * the classes are no longer needed
  * the class headers (`…Module.hpp`) have been removed
  * the class header creation script `module_header.py` has been removed
  * the command in `CMakeLists.txt` to run the header script has been removed.
  * `#include`s of the `…Module.hpp` file have been replaced with an `#include` of `NextsimModule.hpp` and an `#include` of the interface header file.
 * Change the API of `Module<>`. It now depends on the definition of two classes before it is `#include`d: 
  * `HelpMap`
  * `Config` which defines a `std::string Config::getImpl(const std::string&)` function.
  * There may be a better way to handle this, but more template arguments was much too complex a solution.
* Add internal options to set the implementation without setting the unique instance to prevent recursive initialization. The external API remains the same.
* Move the class-static data in `Module<>` to be function-local statics. This requires a slight change to the API and a different format in the generated `module.cpp` files. 
* The `fn` type is now a plain function pointer, rather than a `std::function` object. This required replacing a lambda with an ad-hoc function in one test.
* The variables holding the implementation names are now namespaced into `namespace Module::InterfaceClass`.
* The necessary changes to `module_builder.py` to support the above changes.
* Adds a finalizer method to the `Module<>` template. This will call the destructor on the static instance, allowing all the data arrays the static instance holds to be deleted.

#### Changes to `ConfiguredModule`
* Remove static data from `ConfiguredModule`.  Module configurations are now parsed on demand.
* Remove the call from `main` to parse the configuration for the modules.
* With the change to lazy configuration, it is impossible to reconfigure modules automatically from the config text. Change to explicit calls to `setImplementation()`.
* The `ConfiguredModule_test` now only checks that the module configuration options are parsed correctly. Everything else is handeled by the `Module<>` template class.

#### Miscellaneous other changes
* Removed the directory `ModuleLoaderTestModules`. These are very defunct.
* A fix to `IceGrowth_test`. This is so far from the Module changes, I don't see how it could have been working before.

### `FileCallbackCloser`
Replaced the static `std::list` with a singleton function providing a `std::list`.

### `Iterator`
Removed the static implementation of `NullIterator`.

### `MissingData`
Replaced the `static double` missing data value with a singleton, inlined access function. The missing data value is created on first access, but still destroyed at the end of execution. This is fine, since it is a scalar double.

Removed the `MissingData.cpp` source file as it is no longer needed.

### `ModelComponent`
* Removed all references to the `registerModule` and the data members it uses. Also remove `hField()` and similar functions, which don't now do anything, and were originally planned to be used alongside the module register.
* Moved the `ModelArrayRefStore` to be lazily allocated.
* Moved the ocean mask and ocean index look-up vector to be lazily allocated.

### `ModelConfig`
Moved the key specification to an `.ipp` file to avoid repeating the text in `Model.cpp`.

### `ParaGridIO`
* Moved the maps between netCDF concepts and `ModelArray` equivalents to `const` instance variables that are initialized as part of the ctor.
* Moved the static maps from file names to netCDF files and current time indices to a singleton-like function. Also combined the maps, to ensure they remain consistent.
* Added the `Finalizer` class to to what `std::atexit` does, except at our convenience, rather than executable clean-up. This includes a new test for the new class. This change is the same as that contained in PR #683 

### `Finalizer`
* Added the `Finalize` class, which mimics `std::atexit()`, but can be called before the end of `main()`.
* Added the `Finalizer.cpp` source file to the CMake system.
* Added a test of the `Finalizer` class.
* Added a `Module::finalize<>` function template.
* Added the instantiation of the above function template to the module generation file.
* Added registration of finalization for a module wherever the module implementation if configured.
  * By calling `registerUnique()` each module instantiation is only called once when the finalization occurs.

---
# Test Description

Apart from deleted tests for deleted classes, all tests should still pass.

### `ModelComponent`
Adds some test to `ModelArray_test` to check sizes around assignment operations.

### `Finalizer`
There is a test for the new `Finalizer` class.

---
# Documentation Impact

No changes to existing documentation. All classes should provided the same interfaces as previously.


core/src/include/Module.hpp
```diff
@@ -20,96 +16,218 @@
...
+            std::throw_with_nested(std::runtime_error(
+                "No implementation named " + implName + " found for Module " + moduleName()));
+        }
+    }
+
+    static void setExternalImplementationInternal(fn generator, bool setStaticInstance)
+    {
+        getGenerationFunction() = generator;
+        if (setStaticInstance)
+            getUniqueInstance(true) = std::move(getGenerationFunction()());
```
Thanduriel, line 221: no std::move needed

core/src/include/Module.hpp
```diff
@@ -20,96 +16,218 @@
...
-using ConfigType = Nextsim::ConfigurationHelp::ConfigType;
-
-template <typename I> HelpMap& getHelpRecursive(HelpMap& map, bool getAll);
-
 template <typename I> class Module {
 public:
-    typedef std::function<std::unique_ptr<I>()> fn;
-    typedef std::map<std::string, fn> map;
-
+    using fn = std::unique_ptr<I> (*)();
```
Thanduriel: a bit pedantic, but type names should be CamelCase according to our naming convention, i.e. `fn` -> `Fn` and `map` -> `Map`

core/src/include/Module.hpp
```diff
@@ -20,96 +16,218 @@
...
+     */
+    static std::unique_ptr<I> getInstance() { return getInstanceInternal(true); }
+
+    /*!
+     * Returns a std::unique_ptr to the static instance of the current implementation.
+     */
+    static std::unique_ptr<I>& getUniqueInstance(bool suppressInit = false)
+    {
+        static std::unique_ptr<I> staticInstance
+            = std::move((suppressInit) ? std::unique_ptr<I>(nullptr) : getInstanceInternal(false));
```
Thanduriel: std::move is redundant because the right-hand-side already is an rvalue

core/src/include/Module.hpp
Thanduriel: no std::move needed

core/src/include/Module.hpp
```diff
@@ -20,96 +16,218 @@
...
+        return staticInstance;
+    }
 
-    static I& getImplementation() { return *staticInstance; }
+    /*!
+     * Returns a reference to the static instance of the current implementation.
+     */
+    static I& getImplementation()
+    {
+        if (!getUniqueInstance().get())
```
Thanduriel: `.get()` is not needed here because std::unique_ptr has a cast to bool to test directly whether something is currently owned

core/src/ConfiguredModule.cpp
```diff
@@ -17,59 +17,36 @@ namespace Nextsim {
...
-    for (auto entry : configuredModules) {
-        std::string implString = vm[addPrefix(entry.first)].as<std::string>();
-        // Only do anything if the retrieved option is not the default value
-        if (implString != defaultStr) {
-            entry.second.first(implString);
-        }
-    }
-}
+// A default string that can never be a valid C++ class name
+static const std::string defaultStr = "+++DEFAULT+++";
```
TomMelt: is empty string also not a valid default? And also not a valid class name?

core/src/ConfiguredModule.cpp
```diff
@@ -17,59 +17,36 @@ namespace Nextsim {
...
-std::string ConfiguredModule::getModuleConfiguration(const std::string& module)
-{
-    return configuredModules[module].second();
+    std::string implString = Configurator::parse(opt)[addPrefix(module)].as<std::string>();
+    // Only do anything if the retrieved option is not the default value
+    if (implString != defaultStr) {
+        return implString;
+    } else {
+        return "";
+    }
```
TomMelt: Can you explain the use-case here? Is the idea we could have modules that have not been fully set up yet i.e., they have `defaultStr`. Or are there cases where modules will necessarily have `defaultStr` for some reason?

core/src/Finalizer.cpp
```diff
@@ -0,0 +1,61 @@
...
+    fns.push_back(fn);
+}
+
+bool Finalizer::registerUnique(const FinalFn& fn)
+{
+    if (contains(fn))
+        return false;
+    registerFunc(fn);
+    return true;
+}
```
TomMelt, line 24: is there a reason this function returns a `bool`. Looks like it is mostly used like a `void` function.

core/src/ParaGridIO.cpp
```diff
@@ -27,56 +28,53 @@
...
+          // clang-format on
+      })
+{
+    if (doOnce()) {
+        // Register the finalization function here
+        Finalizer::registerUnique(closeAllFiles);
+        // Since it should only ever run once, do further one-off initialization: allow distant
+        // classes to close files via a callback.
+        FileCallbackCloser::onClose(ParaGridIO::close);
+        doOnce() = false;
```
TomMelt: As an aside I found an alternate way of doing this... but I think yours is more follow-able. (see example below.... you would need to move the logic into the function though. But it doesnt require manually setting the bool to false). See [here](https://stackoverflow.com/a/36596693/9491878) for more info.

```
void doOnce()
{
    static bool once = [](){
        std::cout << "running once only" << std::endl;
        return true;
    } ();
}

int main()
{

  doOnce();
  doOnce();
}
```

core/src/Iterator.cpp
```diff
@@ -10,12 +10,7 @@
 
 namespace Nextsim {
 
-Iterator::NullIterant Iterator::nullIterant;
-
-Iterator::Iterator()
-    : iterant(&nullIterant)
-{
-}
+Iterator::Iterator() { iterant = new NullIterant(); }
```
Thanduriel: This is a memory leak because this `NullIterant` is never deleted. The ownership model of `Iterator` is not clear to me since raw pointers are used and the ownership is not documented, so I don't have a suggestion on how to solve this.

Thanduriel (changes requested):
These changes greatly improve resource lifetime management. In fact, it is now possible to get useful information out of valgrind:
```
==41968== LEAK SUMMARY:
==41968==    definitely lost: 512 bytes in 3 blocks
==41968==    indirectly lost: 3,569,016 bytes in 24 blocks
==41968==      possibly lost: 1,186,944 bytes in 21 blocks
==41968==    still reachable: 110,717 bytes in 719 blocks
==41968==         suppressed: 0 bytes in 0 blocks
```
Not bad :D. Two of the leaks are from raw pointers holding heap data in the dynamics and they can be addressed in a different issue. The third one is in `Iterator`. I think this leak should be addressed before merging the changes.

My other comments are just for cosmetic code changes, so they are not essential.

For the big picture of resource management, a possible pitfall which remains is that modules have to be manually registered for destruction via `Module::finalize`. That seems manageable but is there a reason why this should be done manually instead of automatically, e.g. on construction?

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
```diff
@@ -46,6 +46,8 @@ ConfigurationHelp::HelpMap& ERA5Atmosphere::getHelpRecursive(HelpMap& map, bool
 
 void ERA5Atmosphere::configure()
 {
+    Finalizer::registerUnique(Module::finalize<IFluxCalculation>);
```
TomMelt, line 49: just a thought... is there a way to force finalization without relying on the user to remember they have to do it?

core/test/Finalizer_test.cpp
```diff
@@ -0,0 +1,119 @@
...
+namespace Nextsim {
+
+// A mock Model class. Does nothing except call copy the behaviour of the real
+// Model class in its destructor
+class Model
+{
+public:
+    Model() = default;
+    ~Model() { Finalizer::finalize(); }
+};
```
TomMelt: is this used?

core/test/Finalizer_test.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void throwHappy() { throw HappyTime(); }
+
+TEST_SUITE_BEGIN("ModelFinalizer");
+TEST_CASE("Duplicate functions")
+{
+    TheCount::reset();
+    REQUIRE(TheCount::count() == 0);
+//    Finalizer::clear();
```
TomMelt: can we remove this comment? and the one below? i.e.,
```
//    Finalizer::clear();
```

TomMelt (changes requested):
Thanks @timspainNERSC. Just a few comment/questions from my side.

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
timspainNERSC: Almost certainly! But what does this use case look like? Which classes would be automatically finalized? I think this is a big enough piece of work on its own that is should be a separate issue where we could discuss the scope and implementation.

core/src/Finalizer.cpp
timspainNERSC: The return value indicates whether the function was successfully registered (`true`) or not (`false`). I this is useful enough functionality, given the minimal cost. Even if it is not currently used.

core/src/ConfiguredModule.cpp
timspainNERSC: You know, that never occurred to me. I still think a positive indication that the user wants to use the default implementation is better than just letting it be an empty string.

core/src/ConfiguredModule.cpp
timspainNERSC: No, it's a use case I had planned for, but no longer expect to ever be used.

core/src/ParaGridIO.cpp
timspainNERSC: I've distilled this down to a version that is both readable and doesn't require manual setting of the bool.

core/src/Finalizer.cpp
TomMelt: I guess my concern was, if you thought you were using it but you were not. Or if it was left in for some use case that we no longer need.

Do you think it would be more useful to raise an exception if the user tries to register an already registered function? Given the use in the rest of the code, my concern is that new developers are not likely to test the boolean return value.

core/src/ParaGridIO.cpp
TomMelt: the best of both worlds :ok_hand:

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
TomMelt: My use case was more around the fact the someone may forget to register the module finalization. By default we could finalize everything at the end if it hasn't been manually triggered beforehand. Maybe I am missing something though. Happy to discuss in the next meeting

TomMelt: So let's say I forget to do this step. How bad is it?

TomMelt (approved):

TomMelt:
Thanks @timspainNERSC , I have approved from my side. I left two conversations "unresolved" just so we can find them easily in the next team meeting and discuss. No outstanding issues. Thanks for addressing my comments

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
timspainNERSC: In the CPU version? It's what `develop` is doing right now without this code, so no problem at all.
In the GPU version, my understanding is that not cleaning up data buffers before the GPU process finishes is a big headache. That is what a lot of the changes in this PR are aiming to address.

core/src/Finalizer.cpp
timspainNERSC: It's there as a service if the developer wants to know if the function was already registered. How it is used currently, I don't care, so I don't check. If a future developer does care, then they can check.

Thanduriel (approved):

timspainNERSC:
> For the big picture of resource management, a possible pitfall which remains is that modules have to be manually registered for destruction via `Module::finalize`. That seems manageable but is there a reason why this should be done manually instead of automatically, e.g. on construction?

I thought that there were cases where registering the final methods at construction would not work. But I cannot identify any right now. This might be worth looking at in the (not to distant) future, but can be a separate issue.

PR #677, opened 2024-09-03
Issue614 missing data

timspainNERSC:
# Replace static non-const `MissingData` with a singleton double

Fixes #614 (partially)

---
# Change Description

Replace the `static double` missing data value with a singleton, inlined access function. The missing data value is created on first access, but still destroyed at the end of execution. This is fine, since it is a scalar double.

Removed the `MissingData.cpp` source file as it is no longer needed.


Thanduriel (approved):
While the new pattern guarantees that any access of `value` will see at least see the `defaultValue` instead of zero, the behavior is similarly error prone if a user provided value should be used instead. Since the value from the config is set after static initialization, the implicit usage restriction is the same in that case.

So, either keeping the old way or merging this is fine.

PR #684, opened 2024-09-04
Finalizer class

### LINKED ISSUE #683: Finalize classes and objects without using `std::atexit`
The C-derived library function `std::atexit()` allows functions to be run during the clean-up phase of the executable, along with running destructors of objects with static storage. It would be beneficial to be able to run the same functions with more control over when these functions are run. One example is cleaning up the OASIS coupler before shutting down MPI.

timspainNERSC:
# Finalizer class

Fixes #683 

---
# Change Description

Adds a finalization class `Finalizer` which is designed to mimic `std::atexit()`, but with more flexibility. The class has a `static` function `finalize()`, which will execute all of the functions designated to be executed. As with `std::atexit()`, the functions must have a signature `void()`.

The class also allows the finalization functions to be cleared, to be queried by a `contains(fn)` function, counted and also run without clearing the functions.

---
# Test Description

A test is included in this PR.

---
# Documentation Impact

The class functions have Doxygen comments describing them.

### DIFF
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -14,6 +14,7 @@ set(BaseSources
     "RectGridIO.cpp"
     "ParaGridIO.cpp"
     "FileCallbackCloser.cpp"
+    "Finalizer.cpp"
     "DevStep.cpp"
     "StructureFactory.cpp"
     "MissingData.cpp"
--- a/core/src/Finalizer.cpp
+++ b/core/src/Finalizer.cpp
@@ -1,0 +1,54 @@
+/*!
+ * @file Finalizer.cpp
+ *
+ * @date Sep 11, 2024
+ * @author Tim Spain <timothy.spain@nersc.no>
+ */
+
+#include "include/Finalizer.hpp"
+
+namespace Nextsim {
+
+void Finalizer::registerFunc(const FinalFn& fn)
+    {
+        auto& fns = functions();
+        fns.push_back(fn);
+    }
+
+bool Finalizer::registerUnique(const FinalFn& fn)
+    {
+        if (contains(fn))
+            return false;
+        registerFunc(fn);
+        return true;
+    }
+
+bool Finalizer::contains(const FinalFn& fn)
+    {
+        auto& fns = functions();
+        for (const auto& stored : fns) {
+            if (stored == fn)
+                return true;
+        }
+        return false;
+    }
+
+void Finalizer::finalize()
+    {
+        while (!functions().empty()) {
+            auto fn = functions().back();
+            // Remove the function from the finalization list before executing it.
+            functions().pop_back();
+            fn();
+        }
+    }
+
+size_t Finalizer::count() { return functions().size(); }
+
+Finalizer::FinalFnContainer& Finalizer::functions()
+    {
+        static FinalFnContainer set;
+        return set;
+    }
+
+}
--- a/core/src/Finalizer.cpp
+++ b/core/src/Finalizer.cpp
@@ -10,45 +10,45 @@
 namespace Nextsim {
 
 void Finalizer::registerFunc(const FinalFn& fn)
-    {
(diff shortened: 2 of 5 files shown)


Thanduriel (commented):
All in all it looks fine to me for its purpose but I have a few nitpicks:
* Prefer `using` instead of `typedef`. This is something I have noticed throughout the code base and is a matter of taste but the consensus in the C++ community is to always use `using` since the syntax is more intuitive and it is more powerful (cf. the [coreguidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using))
* The names `atfinal` and `atfinalUnique` are not ideal. They are not verbs and not properly in camelCase. I suggest `register` and `registerUnique` instead.
* I don't see a reason why the functions need to be implemented `inline` in the header. Can you move the methods into a source file?
* I am not entirely sure about the `run` functionality. I would not expect a destructor-like function to be valid if called twice. So providing this functionally seems like a recipe for bugs. Similarly `clear` looks like an invitation for resource leaks but since the philosophy in the project so far has been "the OS is going to clean up in the end" instead of careful lifetime management, such functionality might be useful.
* There are some unnecessary copies of somewhat expensive `std::function` objects but the Finalizer should not be performance critical so it does't really matter.

timspainNERSC:
> All in all it looks fine to me for its purpose but I have a few nitpicks:
> 
>     * Prefer `using` instead of `typedef`. This is something I have noticed throughout the code base and is a matter of taste but the consensus in the C++ community is to always use `using` since the syntax is more intuitive and it is more powerful (cf. the [coreguidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using))
> 
>     * The names `atfinal` and `atfinalUnique` are not ideal. They are not verbs and not properly in camelCase. I suggest `register` and `registerUnique` instead.
> 
>     * I don't see a reason why the functions need to be implemented `inline` in the header. Can you move the methods into a source file?
> 
>     * I am not entirely sure about the `run` functionality. I would not expect a destructor-like function to be valid if called twice. So providing this functionally seems like a recipe for bugs. Similarly `clear` looks like an invitation for resource leaks but since the philosophy in the project so far has been "the OS is going to clean up in the end" instead of careful lifetime management, such functionality might be useful.
> 
>     * There are some unnecessary copies of somewhat expensive `std::function` objects but the Finalizer should not be performance critical so it does't really matter.

- I have converted to using here and elsewhere as I touch the files. And yes, the `typedef` syntax is always confusing!
- Not `register`! That's a C reserved word. I went with `registerFunc` and `registerUnique`.
- Okay, sure.
- Removed `run()` I was using it for testing and kept it in case I though it would be useful. But what you say makes sense.
- Perhaps regular, old function pointers would be enough.

Thanduriel:
>     * Perhaps regular, old function pointers would be enough.

I think `std::function` is the right choice. The way to prevent copies i was thinking about was to introduce perfect forwarding and move semantics:
```C++
template<typename Fn>
static void registrFn(Fn&& fn)
{
    auto& fns = functions();  
    fns.emplace_back(std::forward<Fn>(fn));  
}
```
Or with slightly less potential for confusing error messages:
```C++
static void registrFn(FinalFn fn)
{
    auto& fns = functions();  
    fns.emplace_back(std::move(fn));  
}
```
But the added complexity is probably not worth the performance gain here. The second location where to my mind unnecessary copies occur are in `finalize()`. There only the order would need to be changed, i.e. call the function from a reference and then pop the entry from the list.

Really, it is best to just ignore my comments on performance. We can always do these optimizations later if it ever becomes an issue. Lets just merge this once the other changes are done.

timspainNERSC:
My design was to mimic `std::atexit()`, which only accepts regular functions (I think?), so only accepting function pointers is enough.

timspainNERSC:
Merged as part of #674

PR #736, opened 2024-11-18
ModelArraySlice and slicing infrastructure

### LINKED ISSUE #735: Slice ModelArrays
Both advection (#99) and the halo regions needed for MPI (#132) require moving around subsets of `ModelArray` data. A unified slicing class/library to provide a common interface would be a useful addition to the model.

- [x] Create a `Slice` class
- [x] Create a class to iterate through a `Slice`
- [x] Implement basic indexing
- [x] Implement 'rest of the array' arguments (what would be written as `[::2]` in Python).
- [x] Implement reverse steps
- [x] Throw an exception when passed a step value of zero
- [x] Implement negative indices
- [x] Implement iterator equality
- [x] Implement the standard `start` and `end` iterators
~- [ ] Implement the standard `rstart` and `rend` iterators~
- [x] Create a class that gives access to a slice of a `ModelArray`
- [x] Assign a scalar to a slice of a `ModelArray`.
- [x] Assign one slice of a `ModelArray` to another.
- [x] Assign a slice of one `ModelArray` to the entirety of another
- [x] Assign the entirety of one `ModelArray` to a slice of another
- [x] Assign a buffer (standard container?) to a slice
- [x] Assign a slice to a buffer (standard container?)
- [x] Doxygen comments

timspainNERSC:
# ModelArraySlice and slicing infrastructure

Fixes #735 
---
# Change Description

Adds a class to allow slicing `ModelArray`s in order to use part of the data within. Also includes the `ArraySlicer::Slice` class that provides the slicing syntax and an iterator class for accessing data.

---
# Test Description

Several test cases that check that the slicing syntax is correctly interpreted and that making reasonable slices produces expected results. To fully test the slice syntax, a comprehensive suite of cases is defined. As far as possible, the class is designed to behave like the numpy Python library. 

---
# Documentation Impact

A section will need to be added to the eventual ModelArray developer documentation. As this does not yet exist, there is nothing to add as part of this PR.


TomMelt:
just mentioning for future reference... If we move to c++23 we may wish to consider using [`mdspans`](https://en.cppreference.com/w/cpp/container/mdspan). This would allow us to easily switch between 1D and N-D array views.

core/src/include/ModelArraySlice.hpp
```diff
@@ -0,0 +1,176 @@
...
+    MASIter(ModelArraySlice& mas);
+    double& operator*() const { return data[iter.index()]; }
+    MASIter& operator++()
+    {
+        ++iter;
+        return *this;
+    }
+    bool operator!=(const MASIter& other) const
+    {
+        //        std::cout << this->iter << "!=" << other.iter << "?" << std::endl;
```
TomMelt:
```suggestion
```

Think this was left in by mistake

core/src/include/ModelArraySlice.hpp
TomMelt, line 1: Could you add docstrings to the methods/functions in this file please?

core/src/include/Slice.hpp
```diff
@@ -0,0 +1,526 @@
...
+        os << m_slice << "[";
+        for (size_t i = 0; i < ndims; ++i) {
+            os << current[i] << ((i < noComma) ? "," : "");
+        }
+        os << "]";
+        return os;
+    }
+
+private:
+    Int dimEnd(size_t dim) const
```
TomMelt, line 479: similarly couple of methods/functions here missing docstrings

core/src/include/indexer.hpp
TomMelt, line 1: similarly missing docstings here

core/test/ModelArraySlice_test.cpp
TomMelt, line 1: Thanks @timspainNERSC , this test was very helpful for me when planning the halo exchange work. As mentioned in the main review summary, it might be worth adding a test for taking the last row of a 2d structure e.g., the following should work
```cpp
    Slice sliceTop   {{{  },{-1}}}; // unexpected behaviour (-1 doesnt work)
```
But currently it doesn't and I have to use this instead.
```cpp
    Slice sliceTop   {{{  },{ny-1}}};
```

TomMelt (changes requested):
@timspainNERSC thanks so much for this functionality. It is really helpful for my halo exchange implementation.

I have been using your slicing class in my current `halo-exchange` branch and it works brilliantly.

Just one issue we have discussed already in the team meeting:

According to my understanding the following slice should work when copy from a sliced ModelArray to a buffer i.e., `ModelArray[Slice].copyToBuffer()`
```cpp
    Slice sliceTop   {{{  },{-1}}}; // unexpected behaviour (-1 doesnt work)
```
But currently it doesn't and I have to use this instead.
```cpp
    Slice sliceTop   {{{  },{ny-1}}};
```

So if possible, my only request is fix the `Slice sliceTop   {{{  },{-1}}};` bug and add a test to the `ModelArraySlice_test.cpp` test.

core/src/include/ModelArraySlice.hpp
timspainNERSC: Removed

core/src/include/Slice.hpp
timspainNERSC: Added docstrings to all public functions and at least some kind of comment to all private ones.

core/src/include/indexer.hpp
timspainNERSC: Added docstrings to all public functions and at least some kind of comment to all private ones.

core/test/ModelArraySlice_test.cpp
timspainNERSC: This should be fixed now.

core/src/include/ModelArraySlice.hpp
timspainNERSC: Added docstrings to all public functions and at least some kind of comment to all private ones.

core/src/include/indexer.hpp
timspainNERSC: Added docstrings.

TomMelt (approved):
Thanks @timspainNERSC. It looks good to me

PR #737, opened 2024-11-18
Abort git commit if pre-commit applies formatting

joewallwork:
In #660 we added a `.pre-commit` file that developers can use to automatically apply formatting and update datestamps. It's a minor annoyance that the commit still goes through even if changes were applied. Then you either have separate "Update datestamps" / "Formatting" commits or have to squash/fixup with `git rebase`.

This PR changes the pre-commit so that it aborts the commit if any changes were made. Then you just re-run the `git commit` command with the formatting/date changes included.


.pre-commit
```diff
@@ -11,3 +13,10 @@ for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
   NEWLINE=$(printf "%s%$((SPACES))s%s\n" " @date" " " "$(date '+%d %b %Y')")
   sed -i~ "s/${LINE}/${NEWLINE}/" ${FILE}
 done
+
+# Abort commit if formatting was applied
+git diff >nextsim_precommit_after.patch
+if [ "$(diff nextsim_precommit_before.patch nextsim_precommit_after.patch)" ]; then
+  echo "Aborting git commit because the pre-commit hook made changes."
```
einola, line 22:
```suggestion
  echo "Aborting git commit because the pre-commit hook made changes."
  rm ${BEFORE} ${AFTER}
```

.pre-commit
```diff
@@ -2,6 +2,8 @@
 
 # An example pre-commit file for nextSIM-DG development
 
+git diff >nextsim_precommit_before.patch
```
einola:
```suggestion
BEFORE=$(mktemp)
git diff > ${BEFORE}
```

.pre-commit
```diff
@@ -11,3 +13,10 @@ for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
   NEWLINE=$(printf "%s%$((SPACES))s%s\n" " @date" " " "$(date '+%d %b %Y')")
   sed -i~ "s/${LINE}/${NEWLINE}/" ${FILE}
 done
+
+# Abort commit if formatting was applied
+git diff >nextsim_precommit_after.patch
```
einola:
```suggestion
AFTER=$(mktemp)
git diff > ${AFTER}
```

einola (approved):
Suggestions to make the .patch files inte temporary files with mktemp. They should also be removed afterwards.

This is indeed a minor problem :) I always just do

$ git commit
$ git commit --amend

But this is a bit better.

joewallwork:
The Mac CI hit what's presumably an unrelated error:
```
==> Pouring pkgconf--2.3.0_1.arm64_sonoma.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /opt/homebrew
```
I tried re-triggering it but this didn't help.

PR #738, opened 2024-11-19
Handle XIOS contexts

### LINKED ISSUE #731: XIOS: Enable several tests in same script
Currently, the XIOS server isn't fully destroyed at the end of a doctest `TEST_CASE`. This means that if we have two or more `TEST_CASE`s in the same test script then we get errors related to XIOS attributes being created twice.

See splitting of tests in #667.

joewallwork:
# Handle XIOS contexts

Fixes #731.

This PR does several things:
* Provide an `enableXios` utility that is used to configure across all tests.
* Allow a context ID to be specified when creating an `Xios` instance, which allows several such instances to be created in the same run.
* Separate the reading and writing tests by spatial dimension but combine them into one file.
* Arbitrarily set an hour-long timestep for XIOS on initialisation, otherwise strange bugs can crop up.
* Improve clarity of `Xios::finalize` docstring.


core/test/XiosReadWrite_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+    HField field_2D(ModelArray::Type::H);
+    field_2D.resize();
+
+    // Verify fields are read in correctly
+    readFile(&xios_handler, field_2D, "field_2D");
+    const size_t nx = xios_handler.getDomainLocalXSize("xy_domain");
+    const size_t ny = xios_handler.getDomainLocalYSize("xy_domain");
+    for (size_t j = 0; j < ny; ++j) {
+        for (size_t i = 0; i < nx; ++i) {
+            assertIsClose(field_2D(i, j), i + nx * j);
```
joewallwork, line 195: As mentioned in the docstring for `assertIsClose`, there is a bizarre bug that crops up if I try to call `REQUIRE` here. I derived a minimum failing example [here](https://github.com/nextsimhub/nextsimdg/blob/bugfix731_xios-tests-mfe2/core/test/XiosReadWrite_test.cpp) but figuring out the exact issue was taking a while and I feel that this isn't important to track down right now, unless the reviewer(s) think so. If we are happy to leave it for now then I'll open an issue as a reminder.

core/test/XiosReadWrite_test.cpp
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    // Create ParametricGrid and ParaGridIO instances
+    Module::setImplementation<IStructure>("Nextsim::ParametricGrid");
+    ParametricGrid grid;
+    ParaGridIO* pio = new ParaGridIO(grid);
+    grid.setIO(pio);
+
+    // Initialize an Xios instance called xios_handler
+    // TODO: Create XIOS handler along with ParaGridIO instance
```
joewallwork, line 57: As in the previous test scripts, these TODOs are indicators for the design process for closer integration of XIOS.

timspainNERSC (approved):
I haven't even tried to understand the finer points of the XIOS read/write test, but assuming that tests what needs to be tested, this all looks good.

joewallwork:
> I haven't even tried to understand the finer points of the XIOS read/write test, but assuming that tests what needs to be tested, this all looks good.

The same test code is all still there, just in a different format. Thanks @timspainNERSC!

core/test/XiosReadWrite_test.cpp
joewallwork: Opened #740 to follow up.

PR #744, opened 2024-11-25
halo exchange integration

### LINKED ISSUE #132: Support padding of data arrays to enable halo exchanges in distributed-memory version

TomMelt:
Fixes #132

### Change Description

This PR adds necessary changes to support halo cells into `nextsimDG`.

- `ModelArray`'s are padded with halo cells at the moment they are read from disk
- This happens in the appropriate IO routines, e.g.,
  - `core/src/ParaGridIO.cpp`
  - `core/src/RectGridIO.cpp`
- Similarly when data is written to disk the halo cells are discarded

This PR **does not** add halo exchange logic into `nextsimDG`. This will be done
in a following PR. But all of the functionality is here and it is tested in
`core/test/HaloExchangeCB_test.cpp` and `core/test/HaloExchangePB_test.cpp`
(closed and periodic boundaries).

As part of this work, 2 singleton classes have been added:
- `ModelMetadata` (already existed but has been refactored as a singleton)
- `ModelMPI` (new class)

`ModelMPI` stores the MPI communication, rank and size. `ModelMetadata` contains
metadata on the `Model` which includes domain decomposition information which is
required for halo exchange.

Domain decomposition data is produced by the domain `decomp` tool, which is then
read by `ModelMetadata` when it is first initialized.

I have tried to privatize as much of the class data/methods as possible to
simplify their APIs and avoid mis-use. Therefore member data is privatized as
much as possible and public getters have been made available.

### Test Description

This PR adds 4 new tests to check the new functionality, for closed and periodic
boundaries (CB/PB respectively)
- [x] `ModelMetadataCB_test.cpp` and `ModelMetadataPB_test.cpp` - test
  `getPartitionMetadata` can correctly load metadata from a partition metadata
  file into the appropriate data structure e.g., `neighbourExtents`, etc.
- [x] `HaloExchangeCB_test.cpp` and `HaloExchangePB_test.cpp` - check that we
  can do basic halo exchange by making use of the new slicing utilities e.g.,
  `ModelArraySlice`


joewallwork:
Note that I changed the logic for determining whether the Docker workflow is run. I an attempt to fix the test failure https://github.com/nextsimhub/nextsimdg/actions/runs/12009927720/job/33475689562 might it be feasible to rebase on top of or merge in `develop`?

Compare https://github.com/nextsimhub/nextsimdg/blob/halo-exchange/.github/workflows/docker.yml vs https://github.com/nextsimhub/nextsimdg/blob/develop/.github/workflows/docker.yml.

core/src/include/ModelArraySlice.hpp
```diff
@@ -100,11 +100,11 @@ class ModelArraySlice {
     }
 
     ModelArray& copyToModelArray(ModelArray& target) const;
-    template <typename T> T& copyToBuffer(T& buffer)
+    template <typename T> T& copyToBuffer(T& buffer, size_t startIndex = 0)
     {
         // make no especial attempt at efficiency here
         SliceIter thisIter(slice, data.dimensions());
-        auto biter = buffer.begin();
+        auto biter = std::next(buffer.begin(), startIndex);
```
TomMelt: @timspainNERSC this was a bit hacky from me. I want to populate the buffer but from a specific starting index. Let me know if there's a better way of doing this. 

I was trying to avoid creating temporary arrays for each of the edges.

core/src/include/Slice.hpp
```diff
@@ -142,7 +142,18 @@ class Slice {
         }
         std::ostream& print(std::ostream& os) const
         {
-            return os << start << ":" << stop << ":" << step;
+            // TODO remove me before merging into develop
```
TomMelt: TODO: remove this before merging this PR. I was using this for debugging purposes

core/test/HaloExchange_test.cpp
```diff
@@ -0,0 +1,444 @@
...
+#include "mpi.h"
+
+// aliases
+using Indexer::indexer;
+using Slice = ArraySlicer::Slice;
+using VBounds = ArraySlicer::Slice::VBounds;
+using SliceIter = ArraySlicer::SliceIter;
+
+static const int DG = 3;
+static const bool debug = true;
```
TomMelt: TODO: change this to `false` before merging

core/src/include/Slice.hpp
TomMelt: removed

core/test/HaloExchange_test.cpp
TomMelt: done

core/src/include/ModelArraySlice.hpp
TomMelt: this has now been removed as it is no longer needed

core/src/ModelMetadata.cpp
```diff
@@ -29,25 +36,104 @@ const std::string& ModelMetadata::structureName() const
...
+        std::vector<int> numNeighbours = std::vector<int>(mpiSize, 0);
+        std::vector<int> offsets = std::vector<int>(mpiSize, 0);
+
+        // non-periodic neighbours
+        varName = edgeNames[edge] + "_neighbours";
+        neighbourGroup.getVar(varName).getVar(
+            { 0 }, { static_cast<size_t>(mpiSize) }, &numNeighbours[0]);
+
+        // compute start index for each process
+        MPI_Exscan(&numNeighbours[mpiMyRank], &nStart, 1, MPI_INT, MPI_SUM, modelMPI.getComm());
```
dorchard: What if we are at rank = 0? This will leave `nStart` uninitialised (and similarly on line 98) due to `MPI_Exscan` computing the exclusive scan.

core/src/ModelMetadata.cpp
```diff
@@ -29,25 +36,104 @@ const std::string& ModelMetadata::structureName() const
...
+            varName = edgeNames[edge] + "_neighbour_halo_recv";
+            neighbourGroup.getVar(varName).getVar(
+                { nStart }, { count }, &neighbourHaloRecv[edge][0]);
+        }
+
+        // periodic neighbours
+        varName = edgeNames[edge] + "_neighbours_periodic";
+        neighbourGroup.getVar(varName).getVar(
+            { 0 }, { static_cast<size_t>(mpiSize) }, &numNeighbours[0]);
+
```
dorchard: Code after here seems repeated but only the `varName` is changed so this could be refactored to reduce duplication.

core/src/ModelMetadata.cpp
```diff
@@ -29,25 +36,104 @@ const std::string& ModelMetadata::structureName() const
...
+    for (auto edge : edges) {
+        size_t nStart = 0; // start point in metadata arrays
+        size_t count = 0; // number of elements to read from metadata arrays
+        std::vector<int> numNeighbours = std::vector<int>(mpiSize, 0);
+        std::vector<int> offsets = std::vector<int>(mpiSize, 0);
+
+        // non-periodic neighbours
+        varName = edgeNames[edge] + "_neighbours";
+        neighbourGroup.getVar(varName).getVar(
+            { 0 }, { static_cast<size_t>(mpiSize) }, &numNeighbours[0]);
```
dorchard: I would use `numNeighbours.data()` (https://cplusplus.com/reference/vector/vector/data/) instead of `&numNeighbours[0]` as the latter is technically undefined for empty vectors (which shouldn't happen here but could be a strange edge case!). 
(and then in the other cases which have a similar pattern in the code below).

core/src/ModelMetadata.cpp
dorchard: Maybe it's fine because nStart is initialized to `0` and so at rank `0` this will just stay at `0`. But I'm not certain that's how MPI_Exscan behaves so might be worth checking (or doing something explicit to be safe).

core/src/ModelMetadata.cpp
TomMelt: Thanks @dorchard , this has been addressed in 786c157b

core/src/ModelMetadata.cpp
TomMelt: Thanks @dorchard , hopefully this has been addressed in 786c157b. Let me know if you have any other suggestions.

core/src/ModelMetadata.cpp
TomMelt: Thanks @dorchard , this has been addressed in 786c157b. It doesn't fix my problems but it is definitely a great addition and makes the code much safer.

core/src/ModelMetadata.cpp
dorchard: Looks good!

core/src/ModelMetadata.cpp
dorchard: Ah well worth tying down just in case- thanks! It looks good.

core/src/ModelMetadata.cpp
```diff
@@ -26,25 +35,84 @@ const std::string& ModelMetadata::structureName() const
...
+                    varName = edgeNames[edge] + suffixes[i];
+                    neighbourGroup.getVar(varName).getVar(
+                        { nStart }, { count }, arrays[i].get().data());
+                }
+            }
+        }
+    }
 }
 
+// rename
```
timspainNERSC: Either do the renaming or get rid of this comment.

core/test/HaloExchangeCB_test.cpp
timspainNERSC, line 1: Could add a comment on what is tested within this file. And how it differs from the `HaloExchangePB` test.

core/test/HaloExchangePB_test.cpp
timspainNERSC, line 1: Could add a comment on what is tested within this file. And how it differs from the HaloExchangeCB test.

dynamics/src/ParametricMesh.cpp
```diff
@@ -180,6 +180,7 @@ void ParametricMesh::readmesh(std::string fname)
 void ParametricMesh::coordinatesFromModelArray(const ModelArray& coords)
 {
     // Fill in the array sizes from the ModelArray dimensions
+    // TODO check do we need to make any additional changes here
```
timspainNERSC: This sounds like a thing that should be checked before merging.

timspainNERSC (changes requested):
A few changes. Some amount of explanation on what some of the tests are doing, and resolving, or deleting, some TODO comments.

core/src/ModelMetadata.cpp
TomMelt: haha thanks. I think I left it in by mistake. It's now been removed

core/test/HaloExchangeCB_test.cpp
TomMelt: great will do. For anyone reading this comment CB = closed boundaries and PB = Periodic Boundaries. But I will update both tests with a comment explaining this.

dynamics/src/ParametricMesh.cpp
TomMelt: Good point. AFAIK we dont need any changes here any more. Now that the halo exchange logic has also changed. So I will remove this comment. 

But I will raise an issue for a separate PR to handle what we do for different vertex grids. It is not obvious what the most sensible choice is for vertex points in the halo region.

TomMelt: issuse #923 has been raised.

core/test/HaloExchangeCB_test.cpp
TomMelt: Add description in dc5d3208

core/test/HaloExchangePB_test.cpp
TomMelt: Add description in dc5d3208

core/test/XiosDomain_test.cpp
joewallwork, line 1: It looks like you reintroduced this file that'd previously been removed? Was this for a reason?

core/src/ParaGridIO.cpp
```diff
@@ -165,17 +168,46 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
...
+                size_t startIndex = dim.start;
+                size_t localLength = dim.localLength;
+#ifdef USE_MPI
+                if (dt == Dim::X or dt == Dim::Y) {
+                    localLength = localLength - 2 * Halo::haloWidth;
+                }
+                if (dt == Dim::XVERTEX or dt == Dim::YVERTEX) {
+                    localLength = localLength - 2 * Halo::haloWidth;
+                }
+#endif
```
joewallwork, line 185: I think this section could benefit from some comments to say why `X`, `Y`, `XVERTEX`, and `YVERTEX` are being treated in a particular way.

Also note that the two conditionals could be merged.

core/src/ParaGridIO.cpp
```diff
@@ -165,17 +168,46 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
...
             // dims are looped in [dg], x, y, [z] order so start and count
             // order must be reveresed to match order netcdf expects
             std::reverse(start.begin(), start.end());
             std::reverse(count.begin(), count.end());
 
+#ifdef USE_MPI
+            // TODO
+            // at this point we should add the halo exchange logic for each modelarray
+            // need to check what happens for non-H-field modelarrays
+            // need to figure out what happens w.r.t to coords in non-periodic and periodic case
```
joewallwork: Will this be done in this PR or a subsequent one?

core/src/ParaGridIO.cpp
```diff
@@ -213,17 +245,32 @@ ModelState ParaGridIO::readForcingTimeStatic(
...
+            auto startIndex = dim.start;
+            auto localLength = dim.localLength;
+#ifdef USE_MPI
+            if (dt == Dim::X or dt == Dim::Y) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+            if (dt == Dim::XVERTEX or dt == Dim::YVERTEX) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+#endif
```
joewallwork, line 259: See comment above.

core/src/ParaGridIO.cpp
```diff
@@ -298,10 +358,20 @@ void ParaGridIO::dumpModelState(
...
             auto dim = entry.second.definedDimensions.at(dt);
+            size_t localLength = dim.localLength;
+#ifdef USE_MPI
+            if (dt == Dim::X or dt == Dim::Y) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+            if (dt == Dim::XVERTEX or dt == Dim::YVERTEX) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+#endif
```
joewallwork, line 366: As above.

core/src/ParaGridIO.cpp
```diff
@@ -380,11 +460,21 @@ void ParaGridIO::writeDiagnosticTime(
...
             auto dim = ModelArray::definedDimensions.at(dt);
+            auto localLength = dim.localLength;
+#ifdef USE_MPI
+            if (dt == Dim::X or dt == Dim::Y) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+            if (dt == Dim::XVERTEX or dt == Dim::YVERTEX) {
+                localLength = localLength - 2 * Halo::haloWidth;
+            }
+#endif
```
joewallwork, line 467: Again.

core/src/ParaGridIO.cpp
```diff
@@ -416,18 +506,23 @@ void ParaGridIO::writeDiagnosticTime(
...
-                            .at(ModelArray::typeDimensions.at(ModelArray::Type::H)[0])
-                            .localLength,
-            ModelArray::definedDimensions.at(ModelArray::typeDimensions.at(ModelArray::Type::H)[1])
-                .localLength };
+        for (auto dt : ModelArray::typeDimensions.at(ModelArray::Type::H)) {
+            auto dim = ModelArray::definedDimensions.at(dt);
+            auto localLength = dim.localLength;
+#ifdef USE_MPI
+            localLength = localLength - 2 * Halo::haloWidth;
+#endif
```
joewallwork, line 511: Same applies here - this isn't clear.

core/src/ModelMPI.cpp
```diff
@@ -0,0 +1,36 @@
+/*!
+ * @author  Tom Meltzer <tdm39@cam.ac.uk>
+ */
```
joewallwork, line 3: This could benefit from some `@brief` and `@details`.

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,427 @@
+/*!
+ * @author  Tom Meltzer <tdm39@cam.ac.uk>
+ */
```
joewallwork, line 13: This could benefit from some `@brief` and `@details`.

core/src/include/ModelMetadata.hpp
```diff
@@ -24,32 +26,57 @@ class CommonRestartMetadata;
...
+#ifdef USE_MPI
+    inline static ModelMetadata& getInstance(std::string partitionFile = "")
+    {
+        static ModelMetadata instance = ModelMetadata(partitionFile);
+        if (instance.isInitialized) {
+            return instance;
+        } else {
+            throw std::runtime_error("ModelMetadata :: Object needs to be initialized before use.");
+        }
+    }
```
joewallwork, line 61: It looks like you put the default `partitionFile = ""` because the `partitionFile` argument only gets used the first time you call `getInstance`. However, I doubt `""` will work as a partition filename and will likely raise an error that may or may not have a helpful message. Do we not need to add a catch for this?

core/test/ConfigOutput_test.cpp
```diff
@@ -18,10 +18,15 @@
 #include "include/ModelArray.hpp"
 #include "include/ModelArrayRef.hpp"
 #include "include/ModelComponent.hpp"
+#include "include/ModelMPI.hpp"
 #include "include/ModelMetadata.hpp"
 #include "include/ModelState.hpp"
 #include "include/NextsimModule.hpp"
 #include "include/gridNames.hpp"
+#ifdef USE_MPI
+#include "ModelMPI.hpp"
```
joewallwork: `ModelMPI.hpp` included twice.

joewallwork (changes requested):
Nice that function signatures are more consistent between the MPI and non-MPI cases with this change.

I've not gone through all of the new test code but I have a few comments and requests on the source code and changes to the XIOS tests.

I think some of the changes you made to `ParaGridIO.cpp` need to be reflected in `ParaGridIO_Xios.cpp`, but that's my problem, not yours. I'm working on the relevant bits of code in #916 and #917 anyway.

core/test/XiosDomain_test.cpp
TomMelt: Ahh apologies @jwallwork23 . I think this must have snuck back in during the mega merge :scream: ... It was an accident. I can remove it.

I am assuming you dont need it either?

joewallwork: No worries. If you could just revert the add here that'd be great.

core/src/ParaGridIO.cpp
TomMelt: Good point. I have added a description and I made a utility function to simplify the logic and make it a bit easier to understand (see commit f38263bf)

core/src/ParaGridIO.cpp
TomMelt: sorry Tim had already pointed this out and I did remove it. I think the rebase brought it back.

Removed in 80f159bb

core/src/ParaGridIO.cpp
TomMelt: fixed in f38263bf

core/src/ParaGridIO.cpp
TomMelt: fixed in f38263bf

core/test/XiosDomain_test.cpp
TomMelt: 9ebbb9cd

core/src/ParaGridIO.cpp
TomMelt: fixed in f38263bf

core/src/ParaGridIO.cpp
TomMelt: fixed in f38263bf

core/test/ConfigOutput_test.cpp
TomMelt: 769b0e0d (school boy error :disappointed: )

core/src/ModelMPI.cpp
TomMelt: good point. I will write one for this, `Halo` and `ModelMetadata`. Just a very minor question. Do you think it should be in the `.hpp` file or the `.cpp`. I feel it should go in the `.hpp` because that's like the overview and `.cpp` is the implementation. For Halo there's only the `.hpp` so it doesn't matter. But just curious in general. (Also somewhere is better than nowhere currently :sweat_smile: )

core/src/include/ModelMetadata.hpp
TomMelt: Good point. I will add a `std::runtime_error` for empty strings. Because NetCDF has a horrible habit of not reporting any useful errors when the file is there, it would be good for me to put a standard c++ "check if file exists" and raise error if not. That way even if the name isn't empty, it still needs to exist. I have been stung too many times but the partition file not existing :sweat_smile:

core/src/ModelMPI.cpp
joewallwork: Hm good point. I don't actually know which one (if any) Doxygen picks up. The XIOS one has it in both but maybe we should put it in the headers more consistently.

core/src/ParaGridIO.cpp
joewallwork: Thanks, this makes it much clearer :+1:

timspainNERSC (approved):

core/src/include/ModelMetadata.hpp
TomMelt: fixed in 6fd44c5e

TomMelt: had to move it to a different place because it breaks when you use the singleton (see 1830de1b)

core/src/ModelMPI.cpp
TomMelt: I have added doxygen info to  `Halo`, `ModelMPI` and `ModelMetadata` ae334ca9

core/test/ConfigOutput_test.cpp
```diff
@@ -110,26 +129,30 @@ TEST_CASE("Test periodic output")
...
-
-    for (size_t j = 0; j < ny; ++j) {
+#ifdef USE_MPI
+    for (size_t j = 0; j < localNY - 2 * Halo::haloWidth; ++j) {
+        for (size_t i = 0; i < localNX - 2 * Halo::haloWidth; ++i) {
+            hice(i + 1, j + 1) = 0 + 0.01 * (j * nx + (i + offsetX));
+            cice(i + 1, j + 1) = 0.1 + 0.01 * (j * nx + (i + offsetX));
+            hsnow(i + 1, j + 1) = 0.2 + 0.01 * (j * nx + (i + offsetX));
+            tsurf(i + 1, j + 1) = 0.4 + 0.01 * (j * nx + (i + offsetX));
+            topMelt(i + 1, j + 1) = 0.6 + 0.01 * (j * nx + (i + offsetX));
```
joewallwork, line 143: I assume the shift-by-1 of index was intentional to only set the interior points? Might be worth a comment.

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+#ifndef HALOWIDTH
+#define HALOWIDTH 1
+#endif
+
+namespace Nextsim {
+
+/*!
+ * @brief A class to facilitate halo exchange between MPI ranks
+ *
+ * @details
```
joewallwork: Empty details?

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+    /**
+     * @brief Initializes halo metadata including buffers and slice information
+     *
+     * Sets up the send/receive buffers for each component and defines the boundary
+     * slices for halo exchanges. Handles both vertex and non-vertex cases.
+     *
+     * - Calculates total number of halo cells based on halo width
+     * - Allocates send/receive buffers for each component
+     * - Defines edge lengths
+     * - Sets up inner and outer slices for Bottom, Right, Top, Left edges
```
joewallwork, line 99: This part could be under `@details`.

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+     */
+    void intializeHaloMetadata()
+    {
+        // number of halo cells (should be general for any halo width)
+        m_numHaloCells = 2 * haloWidth * (m_innerNx + m_innerNy + 2 * haloWidth);
+
+        // need send / recv buffers for each componenet (e.g., each DGCOMP)
```
joewallwork:
```suggestion
        // need send / recv buffers for each component (e.g., each DGCOMP)
```

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+    std::vector<std::vector<double>>
+        send; // buffer to store halo region that will be read by other ranks
+    std::vector<std::vector<double>>
+        recv; // buffer to store halo region which is read from other ranks
+    ModelArray::DataType tempBuffer;
+
+    /*!
+     * @brief Open memory window to exchange send buffer between MPI ranks.
+     *
+     * @ details this is not intended to be used manually. It should only be called as part of the
```
joewallwork:
```suggestion
     * @details this is not intended to be used manually. It should only be called as part of the
```

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+        MPI_Win_create(&send[idx][0], m_numHaloCells * sizeof(double), sizeof(double),
+            MPI_INFO_NULL, modelMPI.getComm(), &m_win);
+        // remove fence and check that no proceding RMA calls have been made
+        MPI_Win_fence(MPI_MODE_NOPRECEDE, m_win);
+    }
+
+    /*!
+     * @brief Initialise memory window to exchange send buffer between MPI ranks.
+     *
+     * @ details this is not intended to be used manually. It should only be called as part of the
```
joewallwork:
```suggestion
     * @details this is not intended to be used manually. It should only be called as part of the
```

core/src/include/Halo.hpp
```diff
@@ -0,0 +1,452 @@
...
+                        size_t disp = metadata.neighbourHaloSendPeriodic[edge][i];
+                        size_t recvOffset = metadata.neighbourHaloRecvPeriodic[edge][i];
+                        if (isVertex) {
+                            vertexAdjustedPositions(
+                                count = count, disp = disp, recvOffset = recvOffset, edge = edge);
+                        }
+                        MPI_Get(&recv[comp][recvOffset], count, MPI_DOUBLE, fromRank, disp, count,
+                            MPI_DOUBLE, m_win);
+                    }
+                }
```
joewallwork, line 319: This duplication could be reduced by giving `ModelMetadata` getters that can say periodic or not but let's leave it for another time.

core/test/HaloExchangeCB_test.cpp
joewallwork, line 1: I find the test hard to follow because it's difficult to visualise the test data. It's fine to leave it like this for now but might it be possible to refactor later to consider a mesh with fewer data points? Also, we could turn off clang-format to reshape the reference data arrays how their DoFs are distributed spatially.

core/test/HaloExchangePB_test.cpp
joewallwork, line 1: Similarly for this test.

core/test/ModelMetadataCB_test.cpp
```diff
@@ -0,0 +1,78 @@
+/*!
+ * @author  Tom Meltzer <tdm39@cam.ac.uk>
+ */
```
joewallwork, line 8: Would benefit from some info about what the test does.

core/test/ModelMetadataPB_test.cpp
```diff
@@ -0,0 +1,109 @@
+/*!
+ * @author  Tom Meltzer <tdm39@cam.ac.uk>
+ */
```
joewallwork, line 8: And here.

core/test/ParaGrid_test.cpp
```diff
@@ -50,42 +54,51 @@ const double scale = 1e5;
...
+    auto localNY = ModelArray::definedDimensions.at(dimY).localLength;
+#ifdef USE_MPI
+    auto haloWidth = Halo::haloWidth;
+#else
+    auto haloWidth = 0;
+#endif
+
+    // In the following loops we only set the "inner" data
+
+    // Hfield's, DGField and mask
```
joewallwork:
```suggestion
    // HFields, DGField and mask
```
for consistency

core/test/ParaGrid_test.cpp
```diff
@@ -130,43 +151,29 @@ TEST_CASE("Write and read a ModelState-based ParaGrid restart file")
...
+    HField y;
+
+    // Initialize (resize and set to zero) all ModelArrays
+    ModelArray* arrays[] = { &fractional, &fractionalDG, &mask, &coordinates, &x, &y };
+    for (auto arr : arrays) {
+        arr->resize();
+        *arr = 0.;
+    }
+
+    // populate model arrays with dummy data
```
joewallwork: Suggestion to use 'placeholder', 'stub', or 'sample' rather than 'dummy' (https://www.phoronix.com/news/LF-AOUSD-AWSF-Inclusive-Naming)

core/test/ParaGrid_test.cpp
```diff
@@ -308,42 +309,28 @@ TEST_CASE("Write a diagnostic ParaGrid file")
...
```
joewallwork: And here

joewallwork (changes requested):
Thanks for all your hard work on this, @TomMelt! It mostly looks great but I have a few minor requests - mostly related to docs and rewording. Then it should be good to go.

I also have a few suggestions for code changes to ease understanding and reduce duplication but I think they can wait until after this is merged.

core/src/include/Halo.hpp
TomMelt: woops this was added by mistake. The details are at the top of the file. I have removed this.

core/src/include/Halo.hpp
TomMelt: thanks for spotting. It's fixed now.

core/test/ConfigOutput_test.cpp
TomMelt: Thanks added :ok_hand:

core/test/ModelMetadataCB_test.cpp
TomMelt: Thanks see 231126c0

joewallwork (approved):
Thanks for addressing those @TomMelt. Let's merge it! :partying_face:

PR #745, opened 2024-11-26
Fix Mac CI

joewallwork:
Same fix as in https://github.com/nextsimhub/domain_decomp/pull/71.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -99,6 +99,9 @@ jobs:
         brew install netcdf
         brew install netcdf-cxx
         brew install boost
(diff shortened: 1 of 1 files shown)


joewallwork:
Hm, it seems the same fix doesn't work here.

joewallwork:
Reordering the packages seemed to do the trick - see https://github.com/nextsimhub/nextsimdg/pull/745/commits/cddd3754430dda76cb6c7711fcba541e6e76cbca. It seems to be important to do the unlink-relink of pkgconf near the start but after boost.

einola (approved):

PR #749, opened 2024-12-04
Fix up the Model class help.

### LINKED ISSUE #748: Model class help text
Some of the help text for the `Model` class is outdated or just plain missing. Specifically, the run length can be an integer number of second and the help text for the output restart file name is missing.

timspainNERSC:
# Fix up the Model class help.

Fixes #748 

---
# Documentation Changes

* Correct a typo in the description of the model end data.
* Add that the run length and timestep can be set as an integer number of seconds.
* Make explicit that the initial restart file is initial.
* Add the missing help documenting the option for the name or name pattern for the output restart file(s).


joewallwork (approved):
Thanks @timspainNERSC, this does indeed clear things up.

TomMelt (approved):
Thanks @timspainNERSC , also looks good to me :+1:

PR #750, opened 2024-12-04
Allow timestep to be set when XIOS handler is created

joewallwork:
Refinement of #751

# Allow timestep to be set when XIOS handler is created

This PR makes a few minor modifications to the `Xios` handler object:
* Improve clarity of docstrings related to configuration.
* Allow the user to set the timestep, calendar type and context ID when creating the `Xios` instance.
* Set the calendar origin and start as part of the configuration.
* Update tests accordingly.

These steps should make it easier to set up XIOS and reduce the amount of additional boilerplate when using it to drive I/O rather than the existing approach.

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -49,13 +49,16 @@ void enableXios()
  *
  * @param dt Timestep to use for the model
  * @param contextid ID for the XIOS context
+ * @param starttime Datetime string for the start of the simulation
  * @param calendartype Type of calendar to use
  */
-Xios::Xios(const std::string dt, const std::string contextid, const std::string calendartype)
+Xios::Xios(const std::string dt, const std::string contextid, const std::string starttime,
+    const std::string calendartype)
 {
-    calendarType = calendartype;
-    contextId = contextid;
     timestep = Duration(dt);
+    startTime = TimePoint(starttime);
+    contextId = contextid;
+    calendarType = calendartype;
     configure();
 }
 
@@ -127,7 +130,7 @@ void Xios::configureServer()
 
     // Set default calendar origin and start
     setCalendarOrigin(TimePoint("1970-01-01T00:00:00Z")); // Unix epoch
-    setCalendarStart(TimePoint("2023-03-17T17:11:00Z")); // TODO: Read from config
+    setCalendarStart(TimePoint(startTime));
 }
 
 /*!
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -125,9 +125,9 @@ void Xios::configureServer()
     cxios_create_calendar(clientCalendar);
     cxios_update_calendar_timestep(clientCalendar);
 
-    // Set default calendar origin and start // TODO: Pick something sensible
-    setCalendarOrigin(TimePoint("2020-01-23T00:08:15Z"));
-    setCalendarStart(TimePoint("2023-03-17T17:11:00Z"));
+    // Set default calendar origin and start
+    setCalendarOrigin(TimePoint("1970-01-01T00:00:00Z")); // Unix epoch
+    setCalendarStart(TimePoint("2023-03-17T17:11:00Z")); // TODO: Read from config
 }
 
 /*!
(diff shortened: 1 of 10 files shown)


core/src/Xios.cpp
```diff
@@ -116,10 +120,14 @@ void Xios::configureServer(const std::string calendarType)
...
-    cxios_duration timestep { 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0 };
-    cxios_set_calendar_wrapper_timestep(clientCalendar, timestep);
+    cxios_set_calendar_wrapper_type(clientCalendar, _calendarType.c_str(), _calendarType.length());
+    cxios_set_calendar_wrapper_timestep(clientCalendar, convertDurationToXios(_timestep));
     cxios_create_calendar(clientCalendar);
+    cxios_update_calendar_timestep(clientCalendar);
+
+    // Set default calendar origin and start // TODO: Pick something sensible
+    setCalendarOrigin(TimePoint("2020-01-23T00:08:15Z"));
+    setCalendarStart(TimePoint("2023-03-17T17:11:00Z"));
```
joewallwork: Any thoughts on suitable default calendar origin and start times?

joewallwork: I guess the start will be read from the config file.

timspainNERSC: The time classes in the rest of nextSIM use the Unix epoch as their calendar origin.

core/src/include/Xios.hpp
```diff
@@ -146,7 +147,9 @@ class Xios : public Configured<Xios> {
     bool isEnabled;
 
     std::string clientId;
+    std::string _calendarType;
     std::string _contextId;
+    Duration _timestep;
```
timspainNERSC: Unless this is an XIOS convention, I'd discourage leading underscores in nextSIM code, since we generally try to align with the WebKit coding standard (albeit imperfectly).

core/test/XiosCalendar_test.cpp
```diff
@@ -48,6 +48,7 @@ MPI_TEST_CASE("TestXiosInitialization", 2)
     REQUIRE(start == xios_handler.getCalendarStart());
     REQUIRE(start.format() == "2023-03-17T17:11:00Z");
     // Timestep
+    REQUIRE(xios_handler.getCalendarTimestep().seconds() == doctest::Approx(3600.0)); // Default
     Duration timestep("P0-0T01:30:00");
     REQUIRE(timestep.seconds() == doctest::Approx(5400.0));
```
timspainNERSC: I know that testing floating point values can be imprecise, but I'd suggest that if we don't get back the exact integer values of 3600 and 5400, that should also be a test failure.

timspainNERSC (changes requested):
Some suggested changes.

core/src/include/Xios.hpp
joewallwork: Sure, addressed in 197d7473.

core/test/XiosCalendar_test.cpp
joewallwork: Good point - fixed in 2eec43ffe8ec19a2918ee5645ca9b9225786956a.

core/src/Xios.cpp
joewallwork: Thanks, now addressed in 542da5b84767f9dc35057b94cfea63446f73d58c.

joewallwork:
I also provided the option to set the calendar start upon creating an `Xios` instance.

joewallwork:
I just pushed a couple of small commits:
* https://github.com/nextsimhub/nextsimdg/pull/750/commits/cd35f66cc034f7c0a188496ad91a39641f80031a throws errors if the getter for a calendar attribute is called before it's been set.
* https://github.com/nextsimhub/nextsimdg/pull/750/commits/0eae60aa97eb6ba7d1e311719478e9607df64f6d fixes a minor bug in the pre-commit.

joewallwork:
@timspainNERSC please could you re-review this? I think I addressed all your concerns.

.pre-commit
timspainNERSC, line 1: Do we want to merge these changes as part of this PR, rather than in a separate PR?

timspainNERSC (approved):
All looks good. But just make sure that the changes to `.per-commit` are best added as part of this PR.

.pre-commit
joewallwork: I'd figured it was a very small change so didn't know if it warranted it's own PR. But yeah, it'd be better practice to put it in a separate PR. I'll do that now.

joewallwork:
> All looks good. But just make sure that the changes to `.per-commit` are best added as part of this PR.

Thanks @timspainNERSC!

PR #754, opened 2024-12-06
Get the ice-ocean stress into the ModelArrayRef system

### LINKED ISSUE #649: Coupling routines need to receive ice-ocean stresses

einola:
# Get the ice-ocean stress into the ModelArrayRef system
### Task List

---
# Change Description

Introduces a getIceOceanStress function into CGDynamicsKernel, called by getDG0Data when that function gets the right arguments. All derived DynamicsKernels (VPCGDynamicsKernel, BrittleCGDynamicsKernel, and FreeDriftDynamicsKernel) must implement getIceOceanStress.

This is a partial re-implementation of code previously rejected from the develop branch.

---
# Test Description

No CI test. I just ran the benchmark code and checked that the values returned are of the right order of magnitude.

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/core/src/include/ModelComponent.hpp
+++ b/core/src/include/ModelComponent.hpp
@@ -64,6 +64,8 @@ namespace Protected {
     inline constexpr TextTag WIND_V = "WIND_V"; // y(north)-ward component of wind, m s⁻¹
     inline constexpr TextTag ICE_U = "ICE_U"; // x(east)-ward ice velocity, m s⁻¹
     inline constexpr TextTag ICE_V = "ICE_V"; // y(north)-ward ice velocity, m s⁻¹
+    inline constexpr TextTag IO_STRESS_U = "IO_STRESS_U"; // x(east)-ward ice-ocean stress, Pa
+    inline constexpr TextTag IO_STRESS_V = "IO_STRESS_V"; // y(north)-ward ice-ocean stress, Pa
     // Slab ocean fields
     inline constexpr TextTag SLAB_SST = "SLAB_SST"; // Slab ocean sea surface temperature, ˚C
     inline constexpr TextTag SLAB_SSS = "SLAB_SSS"; // Slab ocean sea surface salinity, ˚C
--- a/core/src/include/ModelComponent.hpp
+++ b/core/src/include/ModelComponent.hpp
@@ -1,7 +1,7 @@
 /*!
  * @file ModelComponent.hpp
  *
- * @date 1 Jul 2024
+ * @date 06 Dec 2024
  * @author Tim Spain <timothy.spain@nersc.no>
  * @author Einar Ólason <einar.olason@nersc.no>
  */
@@ -64,8 +64,8 @@ namespace Protected {
     inline constexpr TextTag WIND_V = "WIND_V"; // y(north)-ward component of wind, m s⁻¹
     inline constexpr TextTag ICE_U = "ICE_U"; // x(east)-ward ice velocity, m s⁻¹
     inline constexpr TextTag ICE_V = "ICE_V"; // y(north)-ward ice velocity, m s⁻¹
-    inline constexpr TextTag IO_STRESS_U = "IO_STRESS_U"; // x(east)-ward ice-ocean stress, Pa
-    inline constexpr TextTag IO_STRESS_V = "IO_STRESS_V"; // y(north)-ward ice-ocean stress, Pa
+    inline constexpr TextTag IO_STRESS_X = "IO_STRESS_X"; // x(east)-ward ice-ocean stress, Pa
+    inline constexpr TextTag IO_STRESS_Y = "IO_STRESS_Y"; // y(north)-ward ice-ocean stress, Pa
     // Slab ocean fields
     inline constexpr TextTag SLAB_SST = "SLAB_SST"; // Slab ocean sea surface temperature, ˚C
     inline constexpr TextTag SLAB_SSS = "SLAB_SSS"; // Slab ocean sea surface salinity, ˚C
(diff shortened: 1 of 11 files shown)


core/src/modules/include/ProtectedArrayNames.ipp
```diff
@@ -40,3 +40,5 @@
     { "sss_slab", "SLAB_SSS" }, // Slab ocean surface salinity PSU
     { "qdw", "SLAB_QDW" }, // Slab ocean temperature nudging heat flux, W m⁻²
     { "fdw", "SLAB_FDW" }, // Slab ocean salinity nudging water flux, kg s⁻¹ m⁻²
+    { "taux", "IO_STRESS_U" }, // Ice-ocean stress x(east) direction, Pa
+    { "tauy", "IO_STRESS_V" }, // Ice-ocean stress x(east) direction, Pa
```
timspainNERSC: At least one of these should be 'y (north)'. In fact exactly one.

core/src/include/ModelComponent.hpp
```diff
@@ -64,6 +64,8 @@ namespace Protected {
     inline constexpr TextTag WIND_V = "WIND_V"; // y(north)-ward component of wind, m s⁻¹
     inline constexpr TextTag ICE_U = "ICE_U"; // x(east)-ward ice velocity, m s⁻¹
     inline constexpr TextTag ICE_V = "ICE_V"; // y(north)-ward ice velocity, m s⁻¹
+    inline constexpr TextTag IO_STRESS_U = "IO_STRESS_U"; // x(east)-ward ice-ocean stress, Pa
+    inline constexpr TextTag IO_STRESS_V = "IO_STRESS_V"; // y(north)-ward ice-ocean stress, Pa
```
timspainNERSC: I feel like x & y make more sense for stress components than u & v, but if u & v is the convention, then of course we should use that.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -163,6 +163,22 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
...
+        const double uOceanRel = uOcean(i) - avgU(i);
+        const double vOceanRel = vOcean(i) - avgV(i);
+        const double cPrime = FOcean * std::hypot(uOceanRel, vOceanRel);
+
+        if (name == uIOStressName)
+            return cPrime * (uOceanRel * cosOceanAngle - vOceanRel * sinOceanAngle);
+        else if (name == vIOStressName)
+            return cPrime * (vOceanRel * cosOceanAngle + uOceanRel * sinOceanAngle);
+        else
+            return std::nan("");
```
timspainNERSC: Perhaps `std::numeric_limits<double>::quiet_NaN()`. Creating a NaN with a string is just bizarre looking. Standard and well defined, yes, but just weird.

dynamics/src/include/FreeDriftDynamicsKernel.hpp
```diff
@@ -69,6 +69,22 @@ template <int DGadvection> class FreeDriftDynamicsKernel : public CGDynamicsKern
...
+        const double uOceanRel = uOcean(i) - u(i);
+        const double vOceanRel = vOcean(i) - v(i);
+        const double cPrime = FOcean * std::hypot(uOceanRel, vOceanRel);
+
+        if (name == uIOStressName)
+            return cPrime * (uOceanRel * cosOceanAngle - vOceanRel * sinOceanAngle);
+        else if (name == vIOStressName)
+            return cPrime * (vOceanRel * cosOceanAngle + uOceanRel * sinOceanAngle);
+        else
+            return std::nan("");
```
timspainNERSC: As in `BrittleCGDynamicsKernel`.

dynamics/src/include/VPCGDynamicsKernel.hpp
```diff
@@ -89,6 +89,22 @@ template <int DGadvection> class VPCGDynamicsKernel : public CGDynamicsKernel<DG
...
+        double uOcnRel = u(i) - uOcean(i);
+        double vOcnRel = v(i) - vOcean(i);
+        double absocn = sqrt(SQR(uOcnRel) + SQR(vOcnRel));
+
+        if (name == uIOStressName)
+            return FOcean * absocn * uOcnRel;
+        else if (name == vIOStressName)
+            return FOcean * absocn * vOcnRel;
+        else
+            return std::nan("");
```
timspainNERSC: As in `BrittleCGDynamicsKernel`.

timspainNERSC (changes requested):
Who decided that you need to supply a string argument to generate a NaN value!?
Also u & v versus x & y.

core/src/include/ModelComponent.hpp
einola: Yes, x & y is better. Not sure why I did it like this, but there's no standard really. Using u and v for velocities and x and y for everything else is good, though.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
einola: I didn't know about ``std::numeric_limits<double>::quiet_NaN()``. I'll use that because I agree about the string bit o_o It's a bit of a verbose solution, but it's fine.

timspainNERSC (approved):
Okay, that looks good now

PR #757, opened 2024-12-09
Align update of nextSIM-DG and XIOS calendars

### LINKED ISSUE #740: Interaction between XIOS and doctest?
> As mentioned in the docstring for `assertIsClose`, there is a bizarre bug that crops up if I try to call `REQUIRE` here. I derived a minimum failing example [here](https://github.com/nextsimhub/nextsimdg/blob/bugfix731_xios-tests-mfe2/core/test/XiosReadWrite_test.cpp) but figuring out the exact issue was taking a while and I feel that this isn't important to track down right now, unless the reviewer(s) think so. If we are happy to leave it for now then I'll open an issue as a reminder.

_Originally posted by @jwallwork23 in https://github.com/nextsimhub/nextsimdg/pull/738#discussion_r1848701150_

### LINKED ISSUE #756: Have `ModelMetadata::incrementTime` increment the XIOS calendar, too
nextSIM-DG and XIOS have separate calendar concepts. We should align them as far as possible, including how they are updated.

joewallwork:
# Align update of nextSIM-DG and XIOS calendars

Fixes #756
Fixes #740
Refinement of #751

### Task List

---
# Change Description

This PR adds functionality for incrementing XIOS' calendar and hooks it up in the `ModelMetadata::incrementTime` method.

More specifically:
* Define `Xios::incrementCalendar`.
* Rename `Xios::updateCalendar` and `Xios::setCalendarStep` (for consistency with `Xios::getCalendarStep`.
* Reconfigure the build system so that XIOS can be used more widely (i.e., in the library as well as the tests).

---
# Test Description

The new method is used in the XIOS read/write test. Unfortunately, I found a bug in XIOS (see comments below). However, this is fixed with a minor hack in the `install-xios.sh` script. We should be able to avoid this hack when we update to XIOS3 (see #761).

---
### Pre-Request Checklist


core/test/XiosReadWrite_test.cpp
```diff
@@ -128,6 +130,11 @@ Xios setupXiosHandler(int dim, bool read)
     xios_handler.fileAddField(fileId, fieldId);
 
     xios_handler.close_context_definition();
+    // FIXME: Why do we need to re-set the calendar timestep and start here?
+    //        These are already set in the Xios constructor and the test fails if the following two
+    //        lines are moved above the close_context_definition.
+    xios_handler.setCalendarTimestep(Duration("P0-0T01:30:00"));
+    xios_handler.setCalendarStart(TimePoint("2023-03-17T17:11:00Z"));
```
joewallwork: I'm currently in discussion with the XIOS developers on why this is occurring. Apparently the MFE I provided works fine in XIOS3. (I'm using XIOS2 at revision 2638.)

joewallwork: I considered updating to XIOS3 but found that some of the demos hang with that so figured it'd be better to address this specific issue first.

Fixed the issue above locally with a hack as in https://github.com/nextsimhub/nextsimdg/pull/757/commits/fe1c9b22c4e30054fd44040c2fccdd6ef8d181a2. My latest commit triggered the Docker image to be rebuilt so hopefully it will work in there, too.

joewallwork: Okay so initially the test suite with XIOS failed but then once the Docker image had built, I re-triggered it and it passed!

.github/workflows/test_suite.yml
```diff
@@ -50,9 +50,43 @@ jobs:
...
+            for file in $(find test* -maxdepth 0 -type f)
+            do
+              echo $file
+              nprocs=$(echo $file | sed -r "s/.*MPI([0-9]+)/\1/")
+              mpirun --allow-run-as-root --oversubscribe -n $nprocs ./$file
+            done
+            cd -
+        done
+
+  test-ubuntu-mpi-xios:
```
joewallwork, line 119: Note that this change is the same as the one in #758.

core/src/ParaGridIO.cpp
```diff
@@ -227,7 +227,7 @@ ModelState ParaGridIO::readForcingTimeStatic(
         std::vector<double> timeVec(timeDim.getSize());
         timeVar.getVar(timeVec.data());
         // Get the index of the largest TimePoint less than the target.
-        targetTIndex = std::find_if(begin(timeVec), end(timeVec), [time](double t) {
+        targetTIndex = std::find_if(std::begin(timeVec), std::end(timeVec), [time](double t) {
```
joewallwork, line 230: This is required to avoid compile errors about `begin` and `end` being overloaded and it not being clear which version to use.

core/src/Xios.cpp
```diff
@@ -37,7 +37,6 @@ static const std::map<int, std::string> keyMap = { { Xios::ENABLED_KEY, "xios.en
 //! Enable XIOS in the 'config'
 void enableXios()
 {
-    Configurator::clearStreams();
```
joewallwork, line 40: This change is important when you set up XIOS after you've already configured other parts of nextSIM-DG.

joewallwork:
Oh nice - turns out the XIOS update fixes issue #740. (Dropped workaround in 28b24531.)

core/src/include/ModelMetadata.hpp
```diff
@@ -117,6 +129,9 @@ class ModelMetadata {
 #ifdef USE_MPI
     const std::string bboxName = "bounding_boxes";
 #endif
+#ifdef USE_XIOS
+    Xios* xiosHandler = nullptr;
```
timspainNERSC: Is `ModelMetadata` the best place for the XIOS handler? It doesn't (instinctively) feel like metadata data to me. Perhaps implementing the infrastructure necessary to return an `Xios` singleton.

Also, wherever it ends up, might a smarter pointer be better than a raw pointer?

timspainNERSC (changes requested):
The XIOS handler living in `ModelMetadata` feels out of place. Perhaps define it as a singleton in `Xios.hpp`/`.cpp`?

core/src/include/ModelMetadata.hpp
joewallwork: The way I currently have it set up in #751 is that multiple classes (`ParaGridIO` and `ModelMetadata`) have pointers to it. A singleton would make sense in general, yes. It might make test files containing multiple tests that use XIOS a little trickier to set up, but that's not really the important thing. I'll look into how to implement it. Also good point on using smart pointers.

joewallwork: Opened PR #763 to address this issue, which would merge into this PR if approved.

TomMelt (approved):
Once #763 is merged in this looks good to go :rocket:

core/src/include/ModelMetadata.hpp
TomMelt: @timspainNERSC what kind of smart pointer were you considering? I guess we could do something like `unique_ptr` but this slightly different to the singleton functionality we need. We really want a single instance rather than a unique pointer. But perhaps both is extra protection?

timspainNERSC: I think a `shared_ptr` is more likely to be the type of smart pointer that we need for a singleton.

timspainNERSC: Thinking further, the singleton function should return a reference, rather than any kind of smart or dumb pointer. If an object needs to store a pointer to the object, then a plain pointer should be fine, as the ownership of the object is clear: it belongs to the singleton function.

TomMelt: @AdelekeBankole , place the `unique_ptr` into `xios.hpp`

joewallwork:
~~Let's merge this following #807.~~

Oops, wrong PR.

core/src/include/ModelMetadata.hpp
joewallwork: This discussion will be resolved upon merging #763 with recent changes.

joewallwork:
[Rebased on top of `develop`.]

joewallwork:
The same is true here as in #758 because of the (duplicate) workflow change.
> Ah, with this PR we have an issue that the repo settings expect the test-ubuntu-mpi action to pass, but it's been split out into test-ubuntu-mpi-noxios and test-ubuntu-mpi-xios. So we'll need to edit the repo settings in order to merge it.

timspainNERSC (approved):
That looks good, now

joewallwork:
> That looks good, now

Perfect, thanks @timspainNERSC. Please could you temporarily disable the requirement for the `test-ubuntu-mpi` test to pass before merging? Once this has been merged, there will then be `test-ubuntu-mpi-noxios` and `test-ubuntu-mpi-xios` jobs that we can require instead. (I guess I could alternatively rename `test-ubuntu-mpi-noxios` as `test-ubuntu-mpi` but I feel it's good to be explicit here.)

joewallwork:
Will merge once the periodic Docker build passes.

PR #758, opened 2024-12-09
Set up `ParaGridIO_Xios` as a copy of `ParaGridIO` for now

### LINKED ISSUE #552: Setup framework for implementing `ParaGridIO` using XIOS
The `ParaGridIO` class provides NetCDF I/O functionality for the `ParaGrid` mesh type, which is implemented manually.

As part of the XIOS integration, we will provide an alternative implementation (perhaps called `ParaGridIO_Xios`), which replicates the functionality of `ParaGridIO` using XIOS.

joewallwork:
# Set up `ParaGridIO_Xios` as a copy of `ParaGridIO` for now

Closes #552
Refinement of #751

### Task List

---
# Change Description

This PR creates a `ParaGridIO_Xios` class for us to work on, but just copies what's in `ParaGridIO` for now. This means that we can convert the functionality to use XIOS piece by piece, rather than doing it all at once.

I can't tick the box for "no new warnings are generated" because warnings are indeed generated to say that we're using the XIOS implementation but that it hasn't been completed yet.

---
# Test Description

CMake for the tests needed to be reordered so that XIOS includes can be picked up. `undef INFO` is added in `ParaGrid_test` to avoid a redefinition warning.

An extra CI test workflow is added that builds and runs without XIOS, so we can check that both I/O implementations are working.

---
### Pre-Request Checklist


core/src/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 # Sources for the main neXtSIM model
 
+if(ENABLE_XIOS)
+  set(ParaGridIO_Impl "ParaGridIO_Xios")
+else()
+  set(ParaGridIO_Impl "ParaGridIO")
+endif()
+
```
timspainNERSC: Posibbly add the `.cpp` suffix here to make it clear this is a source file.

core/src/include/ParaGridIO_Xios.hpp
timspainNERSC, line 1: What are the expected differences between the current and `_Xios` versions of the header? Keeping the same header name would mean a lot of the changes in this PR would be unnecessary. If the header does need to change, one option would be to create both `ParaGridIO_NoXios.hpp` and `ParaGridIO_Xios.hpp` with the actual header info in, and to have `#ifdef #include` statements include the correct file.

timspainNERSC (commented):
I'd suggest instead using conditional inclusion in `ParaGirdIO.hpp` if the non-XIOS and XIOS versions of that header need to diverge.

core/src/CMakeLists.txt
joewallwork: Addressed in 1291ae453cfc581d90c42d677709715070466f1e.

core/src/include/ParaGridIO_Xios.hpp
joewallwork: Great point! There will be parts of it that are needed by one implementation and not the other, but we can start out with selective `#ifdef`s and consider splitting if they diverge too far.

Addressed in 1291ae453cfc581d90c42d677709715070466f1e.

TomMelt (approved):
Thanks @jwallwork23 . Tested locally and it all seems fine. I think it also satisfies the requirement of 

> - [ ] no new warnings generated 

I didn't get any at least.... am I missing something?

core/src/CMakeLists.txt
timspainNERSC, line 1: I think that rather than the rather clunky if … set … dereference chain currently implemented, it might be cleaner set a `ParaGridIOImplSource` variable and add that as another option to `NextsimSourcesTemp`.

```
if(ENABLE_XIOS)
  set(ParaGridIOImplSource "ParaGridIO_Xios.cpp")
else()
  set(ParaGridIOImplSource "ParaGridIO.cpp")
endif()
```
…
```
set(NextsimSourcesTemp
 "${BaseSources}"
 "${NextsimSources}"
 "${ModuleSources}"
 "${ParaGridIOImplSource}"
)
```

timspainNERSC (changes requested):
Revisit the way the source file is chosen in the `CMakeLists.txt`.

core/src/CMakeLists.txt
joewallwork: How does this look? 950f703a8e4df094aa997a2be8462659965a190d

joewallwork:
> Thanks @jwallwork23 . Tested locally and it all seems fine. I think it also satisfies the requirement of
> 
> > * [ ]  no new warnings generated
> 
> I didn't get any at least.... am I missing something?

Thanks @TomMelt, not sure what this was about because I don't see any warnings either. Marked it as done!

core/src/ParaGridIO.cpp
```diff
@@ -1,10 +1,11 @@
 /*!
  * @file ParaGridIO.cpp
  *
- * @date Oct 24, 2022
+ * @date 09 Dec 2024
  * @author Tim Spain <timothy.spain@nersc.no>
  */
 
+#ifndef USE_XIOS
```
timspainNERSC: If we are only including this source file when building without XIOS, why do we need the include guards here and in `ParaGridIO_Xios.cpp`?

timspainNERSC (changes requested):
Check whether the include guards in the .cpp files are really necessary. They may be, and I am just missing something.

core/src/ParaGridIO.cpp
joewallwork: Fair point, thanks @timspainNERSC. Addressed in 3463c41379d5d3c22f4f6de8f3680724153d544d and 672da0937350ad24ad48b10bce2b7e2f4b3a3b8a.

joewallwork:
Ah, with this PR we have an issue that the repo settings expect the `test-ubuntu-mpi` action to pass, but it's been split out into `test-ubuntu-mpi-noxios` and `test-ubuntu-mpi-xios`. So we'll need to edit the repo settings in order to merge it.

timspainNERSC (approved):

PR #759, opened 2024-12-10
Fix pre-commit hook; add CMake formatting pre-commit

joewallwork:
# Fix pre-commit hook; add CMake formatting pre-commit

There was a minor issue in the pre-commit hook because the final lines checked for differences in `nextsim_precommit_before.patch` and `nextsim_precommit_after.patch` but these were replaced with calls to `mktemp`.

While fixing this, I realised we aren't currently formatting `CMakeLists.txt` files but tools for this exist. Let me know if this is something we'd like to include.

I also added some docs on how to get the pre-commit hook set up.

(Almost all of the changes in this PR are auto-formatting.)

### DIFF
--- a/.cmake-lint-config.py
+++ b/.cmake-lint-config.py
@@ -1,0 +1,241 @@
+# ----------------------------------
+# Options affecting listfile parsing
+# ----------------------------------
+with section("parse"):
+
+  # Specify structure for custom cmake functions
+  additional_commands = { 'foo': { 'flags': ['BAR', 'BAZ'],
+             'kwargs': {'DEPENDS': '*', 'HEADERS': '*', 'SOURCES': '*'}}}
+
+  # Override configurations per-command where available
+  override_spec = {}
+
+  # Specify variable tags.
+  vartags = []
+
+  # Specify property tags.
+  proptags = []
+
+# -----------------------------
+# Options affecting formatting.
+# -----------------------------
+with section("format"):
+
+  # Disable formatting entirely, making cmake-format a no-op
+  disable = False
+
+  # How wide to allow formatted cmake files
+  line_width = 100
+
+  # How many spaces to tab for indent
+  tab_size = 4
+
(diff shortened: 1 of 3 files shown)


timspainNERSC:
I like the idea of autoformatting the `CMakeLists.txt`. But I don't like the formatting changes included in this PR. I _like_ the lists of sources (&c.) with one item per line. I think it's easier to read and also easier to add or remove items, as this corresponds to adding or removing lines, rather than editing lines.

joewallwork:
> I like the idea of autoformatting the `CMakeLists.txt`. But I don't like the formatting changes included in this PR. I _like_ the lists of sources (&c.) with one item per line. I think it's easier to read and also easier to add or remove items, as this corresponds to adding or removing lines, rather than editing lines.

Let me see if I can set configure options for that.

.github/workflows/linting.yml
```diff
@@ -24,3 +21,16 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang-format
+      run: |
+        pip install cmakelang
+        cmake-lint $(find . -name CMakeLists.txt) -c .cmake-lint-config.py
+        # NOTE: We ignore C0103, which would enforce that variables are all-caps
+        # NOTE: We ignore C0113, which would enforce that comments are used
+        # NOTE: We set the tab size to 4
+        # NOTE: We set the maximum line length to 100
```
joewallwork, line 36: Okay, I figured out how to set a config file with these four edits. With these edits, the lint check passes without modification of the code!

.cmake-lint-config.py
joewallwork, line 1: This file was generated with
```
cmake-lint --disabled {C0103,C0113} --tab-size 4 --line-width 100 --dump-config python > .cmake-lint-config.py
```
so it consists of the defaults plus the four edits. I figure it'd be useful to use the current defaults, given our experience with the default settings of clang-format changing.

timspainNERSC (approved):
👍

PR #760, opened 2024-12-10
Workflow handling

joewallwork:
# Workflow handling

In the interests of green computing, development efficiency and avoiding unnecessary CI runs, this PR sets up paths for workflow handling, so that the linters are only run if the source code or build code change and the test suite is only run if the test suite, build system, or config files change.

The PR also fixes a copy-paste error in the linting workflow (introduced in #759).

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -6,6 +6,7 @@ on:
   pull_request:
     branches: [ main, develop ]
     paths:
+      - '.github/workflows/linting.yml'
       - '**.cpp'
       - '**.hpp'
       - '**CMakeLists.txt'
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -5,6 +5,10 @@ on:
     branches: [ main, develop ]
   pull_request:
     branches: [ main, develop ]
+    paths:
+      - '**.cpp'
+      - '**.hpp'
+      - '**CMakeLists.txt'
 
 jobs:
   clang-format:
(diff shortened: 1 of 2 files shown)


joewallwork:
If you check the Actions logs, you should see that:
* My initial push (at https://github.com/nextsimhub/nextsimdg/pull/760/commits/9d1059e21087960b6340e80a5c8e96b0822e0954) didn't trigger any CI jobs because nothing in the source code, build system, or config files changed.
* My second push (at https://github.com/nextsimhub/nextsimdg/pull/760/commits/61523e8025fe14c0832e32fdcae29f044027c9d1) triggered the linting workflows (but not test workflows) because I added that they should be run if the `linting.yml` workflow file itself changes.
* My third push (at https://github.com/nextsimhub/nextsimdg/pull/760/commits/562bd6dddb5212f0f4d62fbc7be7f40773cdc78f) also triggered the test suite workflows because I added that they should be run if the `test_suite.yml` workflow file itself changes.

.github/workflows/linting.yml
```diff
@@ -5,13 +5,18 @@ on:
...
+      - '**.hpp'
+      - '**CMakeLists.txt'
 
 jobs:
   clang-format:
     runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v2
-    - name: clang-format
+    - name: cmake-format
```
einola: Doesn't this change belong in line 29/34?

einola (changes requested):
Sounds like a good idea!

But isn't there a mistake on line 14/19 cf line 29/34 in linting.yml?

.github/workflows/linting.yml
joewallwork: Oops! Good spot - fixed in 470d3bd9.

einola (approved):

PR #763, opened 2025-01-02
Implement Xios handler as a singleton

joewallwork:
Note that this PR merges into the branch for #757, not `develop`.

This PR implements the `Xios` handler object as a singleton, as requested. This means that nextSIM-DG classes don't need to own pointers to the current working `Xios` instance. There can be only one `Xios` instance. The first time the static `Xios::getInstance` method is called, it creates the singleton and returns a pointer to it. On subsequent calls, this method simply returns a pointer to the singleton.

Unfortunately, I had to rework the XIOS read/write test. The singleton design pattern isn't suitable for the way it's currently written, so I re-separated out the reading and writing parts and re-combined the 2D and 3D cases.


TomMelt (approved):
Thanks @jwallwork23, looks good to me. I have tested it locally and all tests pass.

core/src/include/Xios.hpp
```diff
@@ -30,12 +30,37 @@ namespace Nextsim {
 void enableXios();
 
 class Xios : public Configured<Xios> {
-public:
+private:
+    static Xios* instancePtr;
```
timspainNERSC: Avoid static members in classes. Wrap this in a function to allow it to at least be lazily constructed. See for example [Configurator.hpp:185](https://github.com/nextsimhub/nextsimdg/blob/100e2b742713ce6a809bc2d7fa9386864a1f1852/core/src/include/Configurator.hpp#L185)

Making the pointer a static variable in `getInstance()` would achieve this.

core/src/include/Xios.hpp
```diff
@@ -30,12 +30,37 @@ namespace Nextsim {
...
+public:
+    Xios(const Xios& xiosHandler) = delete;
     ~Xios();
 
+    /*
+     * Define Xios handler Singleton
+     *
+     * NOTE: The arguments will only be used the first time this is called.
+     */
+    inline static Xios* getInstance(const std::string dt = "P0-0T01:00:00",
```
timspainNERSC: Most of the changes in this PR would be unnecessary if references were used instead of pointers. Change the return type to `Xios& getInstance()`.

core/src/include/Xios.hpp
```diff
@@ -30,12 +30,37 @@ namespace Nextsim {
...
+        const std::string contextId = "nextSIM-DG",
+        const std::string startTime = "1970-01-01T00:00:00Z",
+        const std::string calendarType = "Gregorian")
+    {
+        if (isNull()) {
+            instancePtr = new Xios(dt, contextId, startTime, calendarType);
```
timspainNERSC: If the `Xios` class was a static function variable, then it wouldn't have to be dynamically allocated.

timspainNERSC (changes requested):
This needs serious re-working. Most of the changes would be unnecessary if a reference to the `Xios` class was used instead of returning a pointer.

core/src/include/Xios.hpp
joewallwork: Thanks for this suggestion and for pointing to an example @timspainNERSC. This is a much neater implementation. Done in 10d6207d42813c5009e7a69bba47c07bfd05f110.

core/src/include/Xios.hpp
joewallwork: See 10d6207d42813c5009e7a69bba47c07bfd05f110.

core/src/include/Xios.hpp
joewallwork: Again, see 10d6207d42813c5009e7a69bba47c07bfd05f110.

joewallwork:
[Cherry-picked @TomMelt's NetCDF CI fix from 9d36ed17183e74d8c034a05b3dcf829faafa4469 to see if it fixes the NetCDF/MPI issue this PR gave in the CI just now.]

joewallwork:
> [Cherry-picked @TomMelt's NetCDF CI fix from [9d36ed1](https://github.com/nextsimhub/nextsimdg/commit/9d36ed17183e74d8c034a05b3dcf829faafa4469) to see if it fixes the NetCDF/MPI issue this PR gave in the CI just now.]

Hm it seems it does not! will investigate.

joewallwork:
> > [Cherry-picked @TomMelt's NetCDF CI fix from [9d36ed1](https://github.com/nextsimhub/nextsimdg/commit/9d36ed17183e74d8c034a05b3dcf829faafa4469) to see if it fixes the NetCDF/MPI issue this PR gave in the CI just now.]
> 
> Hm it seems it does not! will investigate.

Ah, I do get this failure locally. I was running in the wrong build directory...

joewallwork:
[Rebased on top of `issue756_xios-calendar-increment`.]

core/test/ConfigOutput_test.cpp
timspainNERSC, line 1: I wasn't envisaging using XIOS in ConfigOutput, but instead having a separate class (`XiosOutput`, for example) for handling output when XIOS is being used.

timspainNERSC (approved):
Approved, with a note on the use of XIOS in `ConfigOutput`, which does not require any action at this point.

joewallwork:
> Approved, with a note on the use of XIOS in `ConfigOutput`, which does not require any action at this point.

Great, thanks @timspainNERSC!

PR #765, opened 2025-01-02
Enable tests for PRs that don't merge into main or develop

### LINKED ISSUE #764: Tests don't run on PRs that don't merge into `main` or `develop`
I have a PR (#763) that makes edits to a branch other than `main` or `develop`. To run the tests, I needed to modify the GitHub Actions workflow. Is this intended?

### LINKED ISSUE #769: add branch protection to develop
I have just realised branch protection of ensuring tests pass is not on for `develop` (possibly not `main` either). We should check this. I cannot because I don't have admin privileges.

We should check this in the next team meeting.

joewallwork:
Closes #764
> I have a PR (https://github.com/nextsimhub/nextsimdg/pull/763) that makes edits to a branch other than `main` or `develop`. To run the tests, I needed to modify the GitHub Actions workflow. Is this intended?

Closes #769

I recommend we use this approach described [here](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs) i.e.,:

> Workflows that would otherwise be triggered using `on: push` or `on: pull_request` won't be triggered if you add any of the following strings to the commit message in a push, or the `HEAD` commit of a pull request:
> - `[skip ci]`
> - `[ci skip]`
> - `[no ci]`
> - `[skip actions]`
> - `[actions skip]`

I think we should avoid editing `.github/workflows/test_suite.yml`.

There is another popular approach using something like `if: github.event.pull_request.draft == false` in the workflow, but this comes with many problems. See [here](https://github.com/orgs/community/discussions/25722) for more info.


.github/workflows/test_suite.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main, develop ]
   pull_request:
-    branches: [ main, develop ]
```
joewallwork, line 7: @TomMelt to look into draft PRs. If not, add a comment here for how to ignore certain branches.

AdelekeBankole (approved):
@TomMelt, thank you. This looks good to me at least from what I picked up from yesterday's meeting.

TomMelt:
@timspainNERSC , I just raised #769 because I realised I could merge without the tests running at all. We should be able to fix this next meeting.

TomMelt:
@timspainNERSC , can you take a look at branch protection rules for `develop` and `main`

timspainNERSC:
> @timspainNERSC , can you take a look at branch protection rules for `develop` and `main`

This may be something to do with the branch protection rule 'Require status checks to pass before merging' but I don't understand the github status check system well enough to be confident in enabling this option without further investigation.

joewallwork:
Let's merge this following #807.

joewallwork:
Hm okay so the linting workflows didn't run because they only run when certain changes... but we now require them to pass in order for the PR to be mergeable. So in https://github.com/nextsimhub/nextsimdg/pull/765/commits/b3522a36d08a50258fb267ff4bc1bd5b799d5c0e I add workflow triggers upon review. This isn't ideal because tests because there will be unnecessary test runs in some cases, but it's what we need to meet the new criteria.

TomMelt (approved):
~LGTM~ :rocket: 

I have unapproved for now just till we decide best way forward

.github/workflows/linting.yml
```diff
@@ -10,6 +10,7 @@ on:
       - '**.cpp'
       - '**.hpp'
       - '**CMakeLists.txt'
+  pull_request_review:
```
TomMelt, line 13: hmmm it's a shame we have to trade efficiency of running tests "unnecessarily" though.

although the checks are only required going into `develop`  right (or is it `main`)? I wonder if we can we can force runs of tests for `main` and leave the others to just merge anyway? what do you think?

TomMelt (changes requested):
not sure how we want to proceed. Is it possible to force runs on `pull_request_review` only for `main`/`develop` branch?

.github/workflows/linting.yml
joewallwork: Oh yeah good point. Added in https://github.com/nextsimhub/nextsimdg/pull/765/commits/587e2d81e41c83da65e898b334ed75e5394467f4.

joewallwork:
The annoying thing now is that the tests get triggers twice because there's both a `pull_request` and a `pull_request_review` trigger. I would've thought there's a way to avoid this duplication but haven't found anything as yet.

TomMelt (approved):
LGTM

PR #776, opened 2025-01-16
fixed NH generation scripts

### LINKED ISSUE #770: high-res realistic setup is unstable
I have been running the setup from the integration test (i.e. northern hemisphere, TOPAZ4 + ERA5 forcing) for the different available grids. It works fine up to 6.25 km, but the 3.125 km grid becomes unstable after just a few steps, even for very small time-steps (e.g. 2s).  I can't pinpoint a specific point where the instability originates but it does look like it expands from the coast-line (showing hice steps 3 and 4):
step3
step4

The setup was created using the scripts `era5_topaz4_maker.py` and `make_initNH.pyy` with the only modification being an added single cell boundary around the domain. The whole setup is available [here](https://cloud.ovgu.de/s/8KeMDacLBa6nnoq) and can be run with the current develop version. I also tried various tweaks to pin down the problem including a landmask consisting of just the outer boundary and dummy thermodynamics but it seems to persist.

Thanduriel:
# Pull Request Title

Fixes #770

---
# Change Description

The script `era5_topaz4_maker.py` now uses the mesh coordinates from the init file, ensuring that the grid is the same. The generated forcing files therefore no longer have undefined values on sea-ice cells for any grid spacing and the manual fix with hard-coded indices was removed.

Furthermore, both scripts feature more cmd args such as the paths of the required files, making them more flexible.

---
# Test Description

The generated files where tested manually. The simulation no longer becomes unstable instantly. I am not sure if this impacts the integration test which is based on these scripts.

---
# Documentation Impact

The script `make_init25kmNH.py` was replaced with `make_initNH.py`. I am not sure if this  script is mentioned anywhere.

### DIFF
--- a/run/era5_topaz4_maker.py
+++ b/run/era5_topaz4_maker.py
@@ -51,18 +51,18 @@ def create_topaz_times(start_tm, stop_tm):
     return (unix_times, hour_times - topaz4_hours)
 
 # Returns the file name that holds the ERA5 data for a given field at a given time
-def era5_source_file_name(field, unix_time):
+def era5_source_file_name(field, unix_time, path):
     file_year = time.gmtime(unix_time).tm_year
-    return f"ERA5_{field}_y{file_year}.nc"
+    return f"{path}/ERA5_{field}_y{file_year}.nc"
 
 # Returns the file name that holds the TOPAZ data for a given field at a given time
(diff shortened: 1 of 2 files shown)


timspainNERSC (approved):
That all looks reasonable.

PR #778, opened 2025-01-20
Handle higher DG components at restart

### LINKED ISSUE #777: Handle higher DG components correctly, especially at restart
Higher DG components are not being handled correctly when passing data to the dynamics.

timspainNERSC:
# Handle higher DG components at restart

Fixes #777

---
# Change Description

Makes the necessary changes to support reading the full DG components from a restart file.

- Explicitly copy only the DG0 component of the initial `ModelState` data into the arrays of `PrognosticData`.
- Consolidate the setting of data for `CGDynamicsKernel` is a single function called with several different arguments, rather than copying the same (small) code fragment over and over.
- Remove the line of code that zeroes out the higher DG components whenever data is copied.
  - Tidy up commented-out code in `DGModelArray`.
- Initialize dynamics prognostic fields to 0 (all components).
- Update the default dynamics to BBM.
- Make the mock dynamics module class used in the Prognostic data tests persist data (especially higher DG components) like the real dynamics modules do.
- Make the DGModelArray test zero the higher DG components, rather than it automatically happen on assignment from the ModelArray to the DG vector.

### DIFF
--- a/core/src/PrognosticData.cpp
+++ b/core/src/PrognosticData.cpp
@@ -54,7 +54,7 @@ void PrognosticData::configure()
 }
 
 // Copies an HField from a source ModelArray that is either an HField or a DGField.
-void copyFiniteDifference(const ModelArray& source, ModelArray& sink)
+void copyMeanComponent(const ModelArray& source, ModelArray& sink)
 {
     if (source.nComponents() > 1) {
         sink.setData(source.data().col(0));
@@ -72,13 +72,13 @@ void PrognosticData::setData(const ModelState::DataMap& ms)
         noLandMask();
     }
 
-    copyFiniteDifference(ms.at(hiceName), m_thick);
-    copyFiniteDifference(ms.at(ciceName), m_conc);
-    copyFiniteDifference(ms.at(ticeName), m_tice);
-    copyFiniteDifference(ms.at(hsnowName), m_snow);
+    copyMeanComponent(ms.at(hiceName), m_thick);
+    copyMeanComponent(ms.at(ciceName), m_conc);
+    copyMeanComponent(ms.at(ticeName), m_tice);
+    copyMeanComponent(ms.at(hsnowName), m_snow);
     // Damage is an optional field, and defaults to 1, if absent
     if (ms.count(damageName) > 0) {
-        copyFiniteDifference(ms.at(damageName), m_damage);
+        copyMeanComponent(ms.at(damageName), m_damage);
     } else {
         m_damage.resize();
         m_damage = 1.;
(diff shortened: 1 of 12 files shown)


core/src/PrognosticData.cpp
```diff
@@ -53,22 +53,32 @@ void PrognosticData::configure()
     tryConfigure(iceGrowth);
 }
 
+// Copies an HField from a source ModelArray that is either an HField or a DGField.
+void copyFiniteDifference(const ModelArray& source, ModelArray& sink)
```
einola:
```suggestion
void copyDGMean(const ModelArray& source, ModelArray& sink)
```

You're copying the mean value of the DG field, hence the rename. Calling it finite difference is not right. I guess you were thinking about finite volumes ... but that's strictly not correct either, even if zero-order DG is almost finite volumes (but not exactly if I understand it correctly).

core/src/PrognosticData.cpp
```diff
@@ -53,22 +53,32 @@ void PrognosticData::configure()
...
     }
 
-    m_thick = ms.at("hice");
-    m_conc = ms.at("cice");
-    m_tice = ms.at("tice");
-    m_snow = ms.at("hsnow");
+    copyFiniteDifference(ms.at(hiceName), m_thick);
+    copyFiniteDifference(ms.at(ciceName), m_conc);
+    copyFiniteDifference(ms.at(ticeName), m_tice);
+    copyFiniteDifference(ms.at(hsnowName), m_snow);
```
einola:
```suggestion
    copyDGMean(ms.at(hiceName), m_thick);
    copyDGMean(ms.at(ciceName), m_conc);
    copyDGMean(ms.at(ticeName), m_tice);
    copyDGMean(ms.at(hsnowName), m_snow);
```

core/src/PrognosticData.cpp
```diff
@@ -53,22 +53,32 @@ void PrognosticData::configure()
...
     // Damage is an optional field, and defaults to 1, if absent
     if (ms.count(damageName) > 0) {
-        m_damage = ms.at(damageName);
+        copyFiniteDifference(ms.at(damageName), m_damage);
```
einola:
```suggestion
        copyDGMean(ms.at(damageName), m_damage);
```

einola (approved):
All fine, except some nit-picking on a function name.

timspainNERSC:
> All fine, except some nit-picking on a function name.

I don't particularly like referencing DG outside of the dynamics, so how about `copyMeanComponent()`?

einola (approved):

Issue #774, opened 2025-01-15
Error checking


einola:
The model needs to have sensible error checks. The purpose of those is:
 1. Make sure user-supplied options are valid
 2. Make sure forcing data is valid
 3. Stop the model if it enters an unrealistic or non-physical state

Some checks should always be done, while others may only be active in debug mode.

einola:
For 1. we need to check
 - If the section is valid.
 - If the options in each section are valid
 - If the values given to options are valid

The checks in 2 should be done in the given modules, e.g., ERA5Atmosphere or TOPAZOcean. Options in these modules allow for enabling or disabling these checks.

We should probably always perform basic checks for 3 in PrognosticData, but additional, optional checks can be performed in modules (e.g., ThermoWinton or BBMDynamics).

einola:
What to do when the check fails?

For 1, we can just gracefully bring down the model. For 2 and 3, I suggest the modules throw an exception, which something (maybe PrognosticData) catches, writes out special diagnostic files, and brings down the model. The diagnostic files should include as many variables as possible to help diagnose the crash.

PR #784, opened 2025-01-28
Issue774 check fields

einola:
# Checks whether model fields have physically reasonable values
Fixes parts of #774

### Task List

---
# Change Description

This PR adds a checkFields function to the ModelComponent class. This means that all derived classes, including all modules, can use it to check the validity of the model fields. The function checks if a given field is in a given range, i.e. min < field[i] < max for all ocean points.

This PR includes two example implementations for the ERA5Atmosphere and TOPAZOcean modules, as well as for PrognosticData. It also includes some testing of the new functionality.

When field values outside the prescribed bounds are found, the function throws a runtime_error, which is caught by the Iterator class. When the exception is caught, the iteration is stopped and the iterator returns a non-zero exit code to the calling function. This means all netCDF files should be closed properly and a final restart file should be written out. 

I also used boost/stacktrace functionality to print out the stack trace when the exception is caught. This does not work on all platforms, but the code should detect functioning and non-functioning platforms and compile accordingly. A more portable approach would be preferrable, but I didn't find one.

At the moment, the stack trace doesn't print file name and line number of the offending call, only the function name. This can by fixed by not linking to nextsimdglib, but directly to the object files themselves.

This PR addresses points 2 and 3 of issue #774 

---
# Test Description

The additional code is tested in two ways, through the realistic integration test.
 1. The integration test now includes test_fields = true for the ERA5Atmosphere and TOPAZOcean modules
 2. I added a second integration test run with erroneous input data and a test to check that the model returns a non-zero exit code.

---
# Documentation Impact

Addresses points 2 and 3 of issue #774.

We should have a developer's guide to writing modules and include a description of how to check module fields. But this needs to be in a separate issue.

---
# Other Details

N/A

---
### Pre-Request Checklist


core/src/PrognosticData.cpp
```diff
@@ -51,6 +51,13 @@ void PrognosticData::configure()
     tryConfigure(pDynamics);
 
     tryConfigure(iceGrowth);
+
+    // clang-format off
+    fieldsToCheck.push_back({"m_conc",   &m_conc,   std::make_pair(0, 1+1e4)});
+    fieldsToCheck.push_back({"m_damage", &m_damage, std::make_pair(0, 1+1e-4)});
+    fieldsToCheck.push_back({"m_snow",   &m_snow,   std::make_pair(0, 10)});
+    fieldsToCheck.push_back({"m_thick",  &m_thick,  std::make_pair(0, 50)});
```
timspainNERSC: Use the names names in `gridNames.hpp`.
Use the `ModelArrayRef` system, since the model has infrastructure for mapping names to `ModelArray` fields.
Use an initializer list, rather than `std::make_pair`.

timspainNERSC (changes requested):
Rather than turning on checks per-module, I think it would be better to take a list of labels for `ModelArrayStore` (that is the same ones used in `ConfigOutput.field_names`) to be checked. This would allow the fields that are checked in any particular run to be set at run time, without hard coding the set of fields in each `ModelComponent`.

einola:
Very good suggestions and I think everything is in place now.

I'm keeping the ability to do per-module error checking. I think this can be very useful for debugging a model that is too big or the run is too long to put through a debugger.

test/run_test_error_handling_test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+
+# Prepares and executes nextSIM-DG integration test based on the January 2010
+# Arctic simulation, but with erroneous forcing data. Tests if the return code
+# is non-zero.
+
+restart_file=init_25km_NH.nc
+era5_file=25km_NH.ERA5_2010-01-01T000000_test_data.nc
+topaz_file=25km_NH.TOPAZ4_2010-01-01T000000_test_data.nc
+out_file=out.integration_test.nc
```
TomMelt: do these do anything? or are they more comments?

core/src/ModelComponent.cpp
```diff
@@ -122,6 +123,42 @@ ModelArray ModelComponent::mask(const ModelArray& data)
...
+            nLayers = 1;
+        else
+            throw std::logic_error(
+                "ModelComponent::checkFields expected a field with 2 or 3 dimensions.\n");
+
+        // layerToCheck must be in scope for checkFieldElement as well
+        for (layerToCheck = 0; layerToCheck < nLayers; ++layerToCheck)
+            overElements(std::bind(&ModelComponent::checkFieldsElement, this, std::placeholders::_1,
+                             std::placeholders::_2),
+                tst);
```
TomMelt: Not a problem, but just curious. Is there a reason to use `bind` over lambdas?

core/src/include/PhysicalBounds.ipp
```diff
@@ -0,0 +1,67 @@
...
+
+/* List of reasonable physical bounds for all variables listed in ProtectedArrayNames.ipp and
+ * SharedArrayNames.ipp */
+
+// Shared arrays
+{ "DAMAGE0", {0, 1} }, // Ice thickness, cell average, m
+{ "H_ICE_cell", {0, 50} }, // Ice thickness, cell average, m
+{ "C_ICE0", {0, 1} }, // Ice concentration
+{ "H_SNOW_cell", {0, 10} }, // Snow depth, cell average, m
+{ "T_ICE0", {-100, 0} }, // Ice temperature, ˚C
```
TomMelt: Also not a big deal, but just raising unicode chars could cause issues for some tools (even in comments). No need to change just raising awareness

TomMelt (approved):
Thanks @einola this will be really helpful to reduce user error.

TomMelt:
@einola , not exactly what you asked for i.e., stack trace, but it is possible to get the filename and line number via macros which should be a truly cross-platform solution e.g.,:

```cpp
std::cout << __FILE__ << " " << __LINE__ << "\n";
```

core/src/ModelComponent.cpp
```diff
@@ -1,13 +1,13 @@
...
  *
- * @date Feb 28, 2022
+ * @date 02 Feb 2025
  * @author Tim Spain <timothy.spain@nersc.no>
  */
 
 #include "include/ModelComponent.hpp"
-
 #include "include/MissingData.hpp"
+#include <cmath>
```
timspainNERSC: I've been laying out the headers in source files as:
"HeaderForThisClass.hpp"
[blank line]
"OtherNextSIMHeader.hpp"
…
[blank line]
<system_headers>
…

This system groups the headers together and avoids clang-format from complaining about alphabetization.
Could you revert to that layout here? (Add a blank line between each of the headers)

core/src/PrognosticData.cpp
```diff
@@ -88,6 +110,51 @@ void PrognosticData::setData(const ModelState::DataMap& ms)
...
+    }
+
+    // Populate fieldsToCheck with user supplied fields
+    for (const auto& x : storeData) {
+        // Only check arrays that are in use (have been set/resized)
+        if (x.second->data().rows() != 0) {
+            fieldsToCheck.push_back(
+                { pfx + ": " + reverseNames.at(x.first), x.second, bounds.at(x.first) });
+        }
+    }
```
timspainNERSC: This should not be in `PrognosticData`. There should probably should be an independent error checking class.

core/src/include/ConfigurationHelp.hpp
```diff
@@ -38,6 +38,14 @@ class ConfigurationHelp {
...
         return output.str();
     }
+
+    static std::string toString(const bool input)
+    {
+        if (input)
+            return { "true" };
+        else
+            return { "false" };
+    }
```
timspainNERSC: This is already a feature of the standard library, see [Converting bool to text in C++](https://stackoverflow.com/questions/29383/converting-bool-to-text-in-c) on StackOverflow.

core/src/include/ModelArrayReferenceStore.hpp
```diff
@@ -72,6 +72,18 @@ class ModelArrayReferenceStore {
...
+    {
+        if (const auto returnValue = storeRW.find(field); returnValue != storeRW.end())
+            return returnValue->second;
+
+        if (const auto returnValue = storeRO.find(field); returnValue != storeRO.end())
+            return returnValue->second;
+
+        return nullptr;
+    }
+
```
timspainNERSC: `storeRW` and `storeRO` are already `std::map`s from the `std::string name`. You don't need to use `find()`.

core/src/include/PhysicalBounds.ipp
```diff
@@ -0,0 +1,67 @@
+/*!
+ * @file PrognosticData.cpp
```
timspainNERSC: Correct the name of the file.

core/src/include/PhysicalBounds.ipp
timspainNERSC, line 1: This could be a class (probably derived from some `std::map`) that returns the limits `std::pair` give a name (or preferably `TextTag`).

core/src/include/PrognosticData.hpp
timspainNERSC, line 1: Move these additions to a new error checking class.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -114,8 +114,8 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
         // The timestep for the brittle solver is the solver subtimestep
         deltaT = tst.step.seconds() / nSteps;
 
-        avgU.zero();
-        avgV.zero();
+        avgU.setZero();
+        avgV.setZero();
```
timspainNERSC: Why change this? `zero()` works perfectly well, and is part of the `cgVector` interface, as opposed to the parent `Eigen::Matrix` interface.

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
```diff
@@ -49,6 +54,20 @@ void ERA5Atmosphere::configure()
...
+    if (doChecks) {
+        // clang-format off
+        fieldsToCheck.push_back({pfx+": tair",  &tair,  {  -100, 80}});
+        fieldsToCheck.push_back({pfx+": tdew",  &tdew,  {  -100, 80}});
+        fieldsToCheck.push_back({pfx+": pair",  &pair,  { 700e2, 1100e2}});
+        fieldsToCheck.push_back({pfx+": sw_in", &sw_in, {    -1, 2e3}});
+        fieldsToCheck.push_back({pfx+": lw_in", &lw_in, {     0, 1e3}});
+        fieldsToCheck.push_back({pfx+": wind",  &wind,  {  -0.1, 150}});
+        fieldsToCheck.push_back({pfx+": uwind", &uwind, {  -150, 150}});
+        fieldsToCheck.push_back({pfx+": vwind", &vwind, {  -150, 150}});
```
timspainNERSC: I think this would look neater if you merged in a initializer list literal. There would at least be only one function call. But this one is a matter of taste.

physics/src/modules/IceOceanHeatFluxModule/BasicIceOceanHeatFlux.cpp
```diff
@@ -27,6 +27,8 @@ void BasicIceOceanHeatFlux::updateElement(size_t i, const TimestepTime& tst)
     // Use the timestep length as the relaxation time scale
     if (cice[i] > 0.) {
         qio[i] = doOne(tf[i], sst[i], mlBulkCp[i], tst.step.seconds());
+    } else {
+        qio[i] = 0.;
```
timspainNERSC: This feels like it should be in its own bug fix PR.

physics/src/modules/IceThermodynamicsModule/ThermoIce0.cpp
```diff
@@ -87,6 +87,7 @@ void ThermoIce0::calculateElement(size_t i, const TimestepTime& tst)
     if (hice[i] == 0. || cice[i] == 0.) {
         deltaHi[i] = 0.;
         snowToIce[i] = 0.;
+        tice.zIndexAndLayer(i, 0) = freezingPointIce;
```
timspainNERSC: This should be in a bug fix PR, probably along with `BasicIceOceanHeatFlux:30`.

physics/src/modules/OceanBoundaryModule/TOPAZOcean.cpp
```diff
@@ -47,7 +50,25 @@ void TOPAZOcean::configure()
...
+    doChecks = Configured::getConfiguration(keyMap.at(CHECKS_KEY), false);
+
+    if (doChecks) {
+        // clang-format off
+        fieldsToCheck.push_back({pfx+": sstExt", &sstExt, {  -5, 50}});
+        fieldsToCheck.push_back({pfx+": sssExt", &sssExt, {   0, 50}});
+        fieldsToCheck.push_back({pfx+": mld",    &mld,    {1e-3, 11e3}});
+        fieldsToCheck.push_back({pfx+": u",      &u,      {  -5, 5}});
+        fieldsToCheck.push_back({pfx+": v",      &v,      {  -5, 5}});
+        fieldsToCheck.push_back({pfx+": ssh",    &ssh,    {  -5, 5}});
```
timspainNERSC: See the comment on `ERA5Atmosphere.cpp`.

physics/src/modules/OceanBoundaryModule/TOPAZOcean.cpp
```diff
@@ -97,9 +120,6 @@ void TOPAZOcean::setData(const ModelState::DataMap& ms)
     slabOcean.setData(ms);
 }
 
-void TOPAZOcean::updateTf(size_t i, const TimestepTime& tst)
-{
-    tf[i] = Module::getImplementation<IFreezingPoint>()(sss[i]);
-}
+void TOPAZOcean::updateTf(size_t i, const TimestepTime& tst) { tf[i] = (*pFreezingPoint)(sss[i]); }
```
timspainNERSC, line 127: I'd rather this continued to use the `Module` infrastructure.

physics/src/modules/OceanBoundaryModule/include/TOPAZOcean.hpp
```diff
@@ -43,14 +45,19 @@ class TOPAZOcean : public IOceanBoundary, public Configured<TOPAZOcean> {
...
     static std::string filePath;
+    static bool doChecks;
 
     HField sstExt;
     HField sssExt;
 
     SlabOcean slabOcean;
+
+    IIceOceanHeatFlux* pIOHeatFlux;
+    IFreezingPoint* pFreezingPoint;
```
timspainNERSC: These shouldn't be raw pointers. In fact, I don't think they should be here at all. Follow the method in the other ocean modules, and use the `Module` infrastructure.

timspainNERSC (changes requested):
Several changes to make (and possibly discuss).

core/src/include/ModelArrayReferenceStore.hpp
einola: Yes, but ``storeRW``and ``storeRO`` may not contain field. I now use nested ``try`` instead and also try and catch in the calling function. I'm not sure which is better.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
einola: Why indeed? I was reading the Eigen::Matrix interface for something else and got confused. Reverted.

physics/src/modules/OceanBoundaryModule/TOPAZOcean.cpp
einola: Ah, yes. I should have asked you about this one. See commit message for bb47c1e1

> I don't understand why, but using Module::getImplementation<Interface>
> directly reset nOcean, oceanIndex, and p_oceanMaskH. But if I first get
> a pointer to the implementation everything works. A bonus is that we can
> call tryConfigure for these modules in a transparent way.

physics/src/modules/OceanBoundaryModule/include/TOPAZOcean.hpp
einola: I copied what you did in ERA5Atmosphere with the IFluxCalculation reference #cargocultcoding I tried a bit with smart pointers, but I wasn't smart enough to make it work.

core/src/PrognosticData.cpp
einola: I guess the point is that this is not a part of the core functionality of ``PrognosticData``, which I agree with. But this check should only be done once and only in ``PrognosticData``, so shouldn't it then be confined to ``PrognosticData`` itself?

physics/src/modules/IceOceanHeatFluxModule/BasicIceOceanHeatFlux.cpp
einola: Now in #789

physics/src/modules/IceThermodynamicsModule/ThermoIce0.cpp
einola: Now in #789

einola:
I think this should be it except
 - The use of raw pointers - which I don't know how to fix
 - Should setFieldsToCheck be in a separate class? I'm not sure

Maybe a discussion in meat space would help.

einola:
Using std::unique_ptr and a separate class for consistency checking. Can you have another look @timspainNERSC

core/src/CheckingModelComponent.cpp
```diff
@@ -0,0 +1,100 @@
+/*!
+ * @file ModelComponent.cpp
```
timspainNERSC: Change the file string to match the file name

core/src/CheckingModelComponent.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    // Populate fieldsToCheck with user supplied fields
+    for (const auto& x : storeData) {
+        PhysicalBounds bounds;
+        fieldsToCheck.emplace_back(
+            prefix + ": " + reverseNames.at(x.first), x.second, bounds.getBounds(x.first));
+    }
+}
+
+} /* namespace Nextsim */
```
timspainNERSC: Github marks the lack of a final newline with a big red symbol, so presumably its important. Add this.

core/src/include/CheckingModelComponent.hpp
```diff
@@ -0,0 +1,49 @@
+/*!
+ * @file ModelComponent.cpp
```
timspainNERSC: Change the file string to match the file name

core/src/CheckingModelComponent.cpp
```diff
@@ -0,0 +1,100 @@
...
+        int nLayers;
+        if (array->nDimensions() == 3)
+            nLayers = array->dimensions()[2];
+        else if (array->nDimensions() == 2)
+            nLayers = 1;
+        else
+            throw std::logic_error(
+                "ModelComponent::checkFields expected a field with 2 or 3 dimensions.\n");
+
+        // layerToCheck must be in scope for checkFieldElement as well
```
timspainNERSC: The checking be better performed using `Eigen` library functions, such as `isnan()`, `max()`, `maxCoeff(posn)`, rather than trying to hack around `overElements()`.

core/src/PrognosticData.cpp
```diff
@@ -51,6 +62,10 @@ void PrognosticData::configure()
     tryConfigure(pDynamics);
 
     tryConfigure(iceGrowth);
+
+    if (getConfiguration(keyMap.at(CHECKFIELDS_KEY), checkFieldsDefault)) {
+        setFieldsToCheck(getConfiguration(keyMap.at(FIELDNAMES_KEY), fieldNamesDefault), pfx);
+    }
```
timspainNERSC: Put the common code and configuration keys into the `CheckingModelComponent` class.

core/src/PrognosticData.cpp
```diff
@@ -177,13 +194,29 @@ ModelState PrognosticData::getStateRecursive(const OutputSpec& os) const
...
+            "The special value \""
+                + all + "\" will output all available fields." },
+        { checkFieldsKey, ConfigType::BOOLEAN, { "true", "false" },
+            ConfigurationHelp::toString(checkFieldsDefault), "",
+            "Switch to check if the fields listed in field_names are within a reasonable "
+            "physical range and not NaN." },
+    };
+    return map;
+}
+
```
timspainNERSC: Put the common help test pertaining to checking into the `CheckingModelComponent` class.

core/src/Iterator.cpp
```diff
@@ -37,16 +37,26 @@ TimePoint Iterator::parseAndSet(const std::string& startTimeStr, const std::stri
...
+int Iterator::run()
 {
+    int return_code = 0;
+
     iterant.start(startTime);
 
     for (auto t = startTime; t < stopTime; t += timestep) {
         TimestepTime tsTime = { t, timestep };
-        iterant.iterate(tsTime);
+        try {
```
timspainNERSC, line 44: Why catch the exception at this level, just to pass an integer return value up to the next layer?

core/src/main.cpp
```diff
@@ -54,11 +56,11 @@ int main(int argc, char* argv[])
         // Apply the model configuration
         model.configure();
         // Run the Model
-        model.run();
+        return_code = model.run();
```
timspainNERSC: Since the exceptions are not actually being handled at the lower levels, if returning a return code is necessary, rather than just letting the unhandled exception escape main, do it handle the exception here.

core/src/CheckingModelComponent.cpp
```diff
@@ -0,0 +1,100 @@
...
+                             std::placeholders::_1, std::placeholders::_2),
+                tst);
+    }
+}
+
+void CheckingModelComponent::checkFieldsElement(size_t i, const TimestepTime& tst) const
+{
+    const auto& x = fieldsToCheck[numToCheck];
+    const double& value = std::get<1>(x)->zIndexAndLayer(i, layerToCheck);
+    const auto& bounds = std::get<2>(x);
```
timspainNERSC: Make this tuple into an actual `class` (or perhaps a `struct`).

timspainNERSC (changes requested):
I think this needs a fair bit of restructuring.

core/src/Iterator.cpp
einola: I did that so that the output buffers would be flushed, the file would be closed, and the model would write an end-of-run restart file, which may contain more interesting information. Not flushing the buffers before crashing is a significant problem in the model's current state.

But this can maybe be done more elegantly, e.g.

```
        } catch (const std::exception& e) {
            interant.stop(tsTime);
            throw;
        }
```

And then catch the exception at a higher level.

core/src/PrognosticData.cpp
einola: I don't know how to do that.

``checkFieldsDefault`` and ``fieldNamesDefault`` are unique to the child classes ``PrognosticData``, ``ERA5Atmosphere``, or ``TOPAZOcean``, or other unimplemented uses of ``CheckingModelComponent``. The default field list should depend on the child class, and whether to check or not should be true for ``PrognosticData`` and false for everyone else.

``keyMap`` is also unique to the child class because the string can have different prefixes, e.g. ``ERA5Atmosphere.check_fields`` vs. ``TOPAZOcean.check_fields``.

So, I agree that there's too much repeated code, but I don't see how to reduce it meaningfully.

einola:
Some solutions and a DRY-fail. Can you have a look @timspainNERSC ?

einola:
Ping @timspainNERSC

timspainNERSC:
I think that everything that is not directly to do with the error checking should be in a separate PR (or possibly several), including adding the return codes and the exception handling.

timspainNERSC:
The more I look at the design of the error checking part, the less I like it. The bounds are all defined in a single file, mixing together dynamics and column physics fields. I still regret putting all the ModelComponent field names in one header. Let's not compound that mistake further.

I have a suggestion for an alternative design that would allow the individual ModelComponents to define their own limits, meaning that new implementations could use the error checking without having to expose their implementation specific data to a single big file.

- Add an upper and lower limit to `ModelArray`. The limit values default to ±inf. Add a `setLimits(double lower, double upper)` function to `ModelArray`.
- Add a `check()` function to `ModelArray` and `ModelArrayRef`, which checks the array for NaN and that the values are within the defined limits. This should use the `Eigen` functions available on `ModelArray::DataType`.
- The new class `CheckingModelComponent` could provide a pure virtual `checkFields()` function that checks the fields relevant to its domain of the model.
- A common way of configuring which fields to check could also be provided here, at the `CheckingModelComponent` level.
- The fields that will be checked or are available to check should be listed in the help text for that component, along with the limits and if they are configurable (hice shouldn't be, but perhaps ice stress could be).

einola:
Working on this branch, I found some problems with the way we pass evaporation and precipitation around. I'm relegating this to a separate PR.

einola:
Depends on #853

core/src/include/ModelArray.hpp
```diff
@@ -113,6 +114,18 @@ class ModelArray {
...
+     *
+     * @param t The ModelArray::Type for the new object.
+     * @param bounds The physical lower and upper bounds for the new object. See lowerPhysicalLimit
+     * and upperPhysicalLimit.
+     */
+    ModelArray(const Type t, const std::pair<double, double>& bounds)
+        : ModelArray(t)
+    {
+        setLimits(bounds.first, bounds.second);
+    }
```
timspainNERSC: Suggest keeping a constructor with parameters `ModelArray(const Type t)`, which delegates to this constructor by supplying default limit arguments of `{-std::numeric_limits<double>::max(), std::numeric_limits<double>::max()}` (not infinity).

core/src/include/ModelArray.hpp
```diff
@@ -620,6 +634,20 @@ class ModelArray {
     //! discontinuous Galerkin components.
     static bool hasDoF(const Type type);
 
+private:
+    double lowerPhysicalLimit = -std::numeric_limits<double>::infinity();
+    double upperPhysicalLimit = std::numeric_limits<double>::infinity();
```
timspainNERSC: Possibly `std::numeric_limits<double>::max()`, rather than `std::numeric_limits<double>::infinity()`.

core/src/ModelArray.cpp
```diff
@@ -276,6 +274,40 @@ ModelArray::MultiDim ModelArray::locationFromIndex(Type type, size_t index)
...
+    if (masked.isNaN().any())
+        throw std::runtime_error("Field contains NaN.");
+
+    // Now we check the bounds and set the array index (i) and value if we're out of bounds
+    size_t i;
+    double value;
+    if (masked.minCoeff() < lowerPhysicalLimit) {
+        value = masked.col(0).minCoeff(&i);
+    } else if (masked.maxCoeff() > upperPhysicalLimit) {
+        value = masked.col(0).maxCoeff(&i);
```
timspainNERSC, line 307: I would suggest changing the limits to a `<=` test. For example, if one were to apply strict limits to ice concentration, then the allowable values would be 0≤c_ice≤1, with both 0 and 1 as allowed values. To set this limit with the current implementation would require setting limits of `{-std::numeric_limits<double>::min(), 1.0+std::numeric_limits<double>::epsilon()}`. This has the knock-on effect of requiring the default limits to be `{-std::numeric_limits<double>::max(), std::numeric_limits<double>::max()}`, rather than using `std::numeric_limits<double>::infinity()`.

core/src/ModelArray.cpp
```diff
@@ -276,6 +274,40 @@ ModelArray::MultiDim ModelArray::locationFromIndex(Type type, size_t index)
...
+    std::string locStr = "[";
+    for (const size_t& l : loc)
+        locStr += std::to_string(l) + ",";
+    locStr.pop_back();
+    locStr.push_back(']');
+
+    throw std::runtime_error("Field contains out-of-bounds value(s), " + std::to_string(value)
+        + " not in [" + std::to_string(lowerPhysicalLimit) + ","
+        + std::to_string(upperPhysicalLimit) + "]. Error at index " + locStr + ".");
+}
```
timspainNERSC, line 325: Further to above comment, this error string implies that the acceptable range is inclusive of the limit values.

core/src/PrognosticData.cpp
```diff
@@ -15,16 +15,23 @@
 
 namespace Nextsim {
 
+static const std::string checkFieldsKey = "debug.check_fields";
+static const std::string checkFieldsFastKey = "debug.check_fields_fast";
+static const std::map<int, std::string> keyMap
+    = { { PrognosticData::CHECKFIELDS_KEY, checkFieldsKey },
+          { PrognosticData::CHECKFIELDSFAST_KEY, checkFieldsFastKey } };
+static constexpr bool checkFieldsDefault = false;
+static constexpr bool checkFieldsFastDefault = true;
```
timspainNERSC, line 22: This should be moved to the `CheckingModelComponent` class. There would need to be a way for subclasses to provide their own configuration prefix and a call to `CheckingModelComponent::configure()` in the `configure()` function of each subclass.
`CheckingModelComponent`:
```
std::string configurationPrefix() virtual = 0;
…
configure()
{
…
    boolCheckAll() = Configured::getConfiguration( configurationPrefix() + checkAllKey, checkFieldsDefault);
    boolCheckFields
        = Configured::getConfiguration(configurationPrefix() + checkFieldsKey, checkFieldsFastDefault);
```
Also, common text for the `help-config` test could be provided by `CheckingModelComponent`, to be used by all subcalsses.

core/src/PrognosticData.cpp
```diff
@@ -51,6 +58,20 @@ void PrognosticData::configure()
...
+        = Configured::getConfiguration(keyMap.at(CHECKFIELDS_KEY), checkFieldsFastDefault);
+    if (boolCheckAll()) {
+        for (const auto& field : getStore().getAllData()) {
+            fieldsToCheck.emplace_back(field.first, field.second);
+        }
+    } else if (boolCheckFields) {
+        fieldsToCheck.emplace_back("thickness", &hiceAdvection);
+        fieldsToCheck.emplace_back("concentration", &ciceAdvection);
+        fieldsToCheck.emplace_back("u_velocity", getStore().getArrayRef(Protected::ICE_U));
+        fieldsToCheck.emplace_back("v_velocity", getStore().getArrayRef(Protected::ICE_V));
```
timspainNERSC: Consider adding a function to the superclass that takes a `std::map<std::string, ModelArrayReference>` argument, rather than repeating `emplace_back` so many times in the derived class code.

timspainNERSC (changes requested):
A lot of the common code could be moved into the `CheckingModelComponent` class, allowing the interface for checking the fields to be simpler in each of the derived classes and reducing the amount of boilerplate that needs to be introduced when a class is converted to become a `CheckingModelComponent`.

core/src/include/ModelArray.hpp
einola: I'm not sure how that would work, because this constructor calls the ``ModelArray(const Type type)`` constructor. But using a default value is a good idea, so I removed this constructor and added a bounds argument to the existing ``ModelArray(const Type type)`` constructor, with ``{-std::numeric_limits<double>::max(), std::numeric_limits<double>::max()}`` as default.

core/src/include/ModelArray.hpp
einola: Done

core/src/ModelArray.cpp
einola: No, but I'll concede that the code is not very intuitive.

The ``if`` returns true if the value is _outside_ the allowed range, in which case we ask Eigen to find the offending value and its location and we proceed to throw an error.

So, with the current code, an initialisation of cice (for example) would be

```
cice(ModelArray::Type::DG, {0, 1})
```

This gives an allowed range of ``0 ≤ cice ≤ 1``, because as long as this is true, neither the ``if`` on line 299 nor the ``else if`` on line 301 is triggered, which brings us to a ``return`` on line 304.

core/src/ModelArray.cpp
einola: See above

core/src/PrognosticData.cpp
einola: Nice suggestion!

core/src/PrognosticData.cpp
einola: Yes, but I decided that the individual modules don't need to have a fine-grained control over their own checks. So, the user has two options now:

debug.check_fields: If true, all the fields registered in the ModelArray system are checked, as well as all the fields individual checking modules register. Default is false.
deebug.check_fields_fast: If true, a select few fields are checked by PrognosticData. Default true.

So, now only PrognosticData needs to have config keys and help.

einola:
Should be ok now, after removing functionality that I don't think is necessary! Have a look on Monday @timspainNERSC

core/src/ModelArray.cpp
timspainNERSC: Probably should expand the comment at line 296 to explain this.

core/src/ModelArray.cpp
timspainNERSC: Review the comments to make sure this is explained.
(Also 'returned' is a perfectly reasonable non-jargon word that doesn't need an apostrophe.)

core/src/include/ModelArray.hpp
```diff
@@ -111,12 +111,15 @@ class ModelArray {
      */
     ModelArray();
     /*!
-     * @brief Construct a ModelArray of the given type and name
+     * @brief Construct a ModelArray of the given type
```
timspainNERSC, line 114: Thanks for spotting and fixing this :)

core/src/PrognosticData.cpp
```diff
@@ -50,6 +57,22 @@ void PrognosticData::configure()
...
+        for (const auto& field : getStore().getAllData()) {
+            addChecks({ { field.first, field.second } });
+        }
+    } else if (checkFast) {
+        addChecks({
+            { "thickness", &hice },
+            { "concentration", &cice },
+            { "u_velocity", getStore().getArrayRef(Protected::ICE_U) },
+            { "v_velocity", getStore().getArrayRef(Protected::ICE_V) },
+        });
```
timspainNERSC: The ice velocity components are not owned by `PrognosticData` and should not be set to be checked here. The call to `addChecks` should be called in `IDynamics` instead.

core/src/include/ModelArrayReferenceStore.hpp
```diff
@@ -69,6 +69,20 @@ class ModelArrayReferenceStore {
...
+            return storeRW.at(field);
+        } catch (std::out_of_range&) {
+            try {
+                return storeRO.at(field);
+            } catch (std::out_of_range&) {
+                throw;
+            }
+        }
+    }
+
```
timspainNERSC: Due to the way `ModelArrayReferenceStore` allows deferred registration, this is SIGSEGV bait. If ice u,v is check in their owning module (that is, the dynamics module), then this function is unnecessary.

timspainNERSC (changes requested):
Bypassing the designed method of getting data from `ModelArrayReferenceStore` is asking for trouble.

core/src/PrognosticData.cpp
einola: Yes, I see that, but this call is dependent upon the option ``debug.check_fields_fast`` being ``true``. So, I decided it would be better to do all of this in one place, where we have access to ``debug.check_fields_fast`` in the first place. Since ``IDynamics`` is not a ``Configured`` class, then I don't think it can have access to ``debug.check_fields_fast``, and I don't think making ``IDynamics`` ``Configured`` is a good idea.

So, my preferred solution is only to check thickness and concentration if ``debug.check_fields_fast`` is true. It's sufficient, because the idea is only to stop the model gracefully if everything goes wrong and errors in the velocity field will propagate to the thickness or concentration fields very quickly.

core/src/include/ModelArrayReferenceStore.hpp
einola: Yes, that's a fair point. I'm removing this function.

timspainNERSC (approved):
I've only reviewed the changes since I last viewed the PR, but I think I am happy for this to be merged now.

PR #787, opened 2025-02-05
Advection data storage

### LINKED ISSUE #782: Variable data redesign
It is increasingly clear that there should be only one copy of variables in nextSIM-DG, and that should be the copy with the full set of DG components. This will require redesigning how data is held in the model and how it is accessed by the thermodynamics and dynamics components.

The design should met the following requirements:
- The use of `DGVector` in the dynamics is paramount. The speed of the dynamics core is crucial to the execution speed of nextSIM-DG.
- There should be minimal changes to the API of ModelArray.

timspainNERSC:
# Advection data storage

Fixes #782

---
# Change Description

Share data between the column physics and dynamics parts of the module without copying. Tidy up some related classes and code that are no longer used and became a problem as these changes were implemented.

- Create a distinction between `ModelArray::Components` (all the DG components at a particular physical location) and a `ModelArray::Component` (a particular DG component, especially DG0, at all grid locations).
  - A set of `Components` map to an `Eigen::Array::row`.
  - A `Component` maps to a an `Eigen::Array::col`.
- Add functions to access a set of `Components`.
- Add functions to access a specific `Component`.
- Allow a `ModelArray` to be cast to and from `ModelArray::DataType`, that is `Eigen::Array`.
- Make the `ModelComponent` ocean mask into an array held in a static function.
- Move the thickness and concentration arrays with all DG components from the dynamics kernel to `PrognosticData`.
- Add a function to `PrognosticData` to copy all components of one `ModelArray` to the correct components of another.
- Update `PrognosticData::updatePrognosticData` to work correctly with the new full DG arrays.
- The output from `PrognositcData` is now all DG components in all cases.
- Add a complementary function to update the arrays after the dynamics runs.
  - Eventually this function will do nothing and should be removed.
- Create a map between a type with multiple components and the equivalent type with only one component.
  - For example `ModelArray::Type::DG` maps to `ModelArray::Type::H`.
- Add a `const` pseudo-`Type` that is the type that should be used for advected fields. Defined in the relevant `ModelArrayDetails.hpp`.
- Add a `using` typedef that defines an `AdvectionField`, which is just `ModelArray`.
- Remove the function `ModelArrayRef::data()`. The underlying data array is no longer accessible (this way).
- Add the four basic arithmetic operators to `ModelArrayRef` for `double` and `ModelArray` arguments.
- Make `ModelArrayRef` work only on the 0 component of the array it references.
  - Except for the `allComponents()` member function which returns a reference to the full array, with all components (also known as the second `Eigen` dimension).
- Add `TextTag`s for read-write access to the cell-averaged DG fields.
- Adjust the dynamics implementations (`core/src/modules/DynamicsModule/*.cpp`) to work with an external source of data for thickness and concentration.
- It's not necessary to use `ModelArrayRef::data()` when passing `ModelArrayRef`s to the dynamics kernels' `setData()` functions.
- Create `DGVectorHolder` class which can point to a DG `ModelArray` (or at least its underlying `Eigen::Array`) and can decay to a `DGVector&`.
  - Use `DGVectorHolder`s for ice thickness and concentration in the dynamics.
  - Add exception errors if the `setData()` function is used to try to set thickness or concentration.
  - Add a function that set the `DGVectorHolder` data.
- Added a `setVelocites` function to `DGTransport`
- Clean up module functions and features that have never really been used.
  - The `Module` registry in `ModelComponent`.
  - The `advectionStore` in `ModelComponent`. The pre-existing store works just fine.
  - The `hFields` (&c.) lists of field names from `ModelComponent` and derived classes.
  - Deleted from the dynamics source tree:
    - `cgParametricMomentum.hpp`
    - `cgParametricMomentum.cpp`
    - `BBM.hpp`
    - `MEB.hpp`
    - `mevp.hpp`

---
# Test Description

- Added a test for the new `DGVectorHolder` class.
- Test the access of `ModelArrayRef` to only the 0 component.
- Test the access of `ModelArrayRef::allComponents` to all components.
- Test the component(s) functions of `ModelArray`.
- Test the direct access of a `ModelArray` to the underlying data type.
- Reconfigure the `ModelComponent` test since there is no longer a `Module` registry.
- Add a `ModelArray::Type` with components to the test `ModelArrayDetails`.
- Adjusted the advection tests to use the `setVelocities` function of `DGTransport`.
- Test creating a `DGVector` directly from `Eigen::Matrix` and `Eigen::Array`.
- The `ThermoIntegration` test now only checks the DG0 component of the read ice thickness.

---
# Documentation Impact

Added docs/data_storage.rst, which documents how data is shared (after applying this PR) between the column physics and dynamics parts of the model.


timspainNERSC:
@winzerle I've added you as a reviewer as this PR changes how the data is passed through the dynamics, and I would like a check that it doesn't break anything. The output of the tests and usual example runs looks fine, at least to me.

winzerle (approved):
Looks good although it's a bit difficult since many files are touched. But interface to transport seems ok. 

It might be good to do a simple speed-check before and after applying such spaces to see, if there are any efficiency issues.

core/src/include/ModelArrayRef.hpp
```diff
@@ -98,36 +98,86 @@ template <const TextTag& fieldName, bool isReadWrite = RO> class ModelArrayRef {
         return dataReference->zIndexAndLayer(hIndex, layer);
     }
 
-    //! Direct access top the underlying data array.
-    const ModelArray& data() const { return *dataReference; }
-    //! Cast the reference class to a real reference to the referenced ModelArray.
-    operator const ModelArray&() const { return data(); }
+    //! Copies component 0 to an appropriate ModelArray type.
```
einola, line 107: As far as I understand, then (for example) ModelArray's A + B will result in A.component(0) + B.component(0) and the higher order components of A, right? This is fine if B doesn't have higher-order components. But if both A and B have higher-order components, then A + B (or A * B) should be found by projecting onto the Gauss points, doing the operation there, and project back. The way the code stands now the higher-order components of B are simply lost.

These projections are too expensive and complicated to do here, I think, but can we simply prevent people from trying to do A + B when both are higher order?

core/src/include/PrognosticData.hpp
```diff
@@ -41,47 +41,29 @@ class PrognosticData : public ModelComponent, public Configured<PrognosticData>
...
-    HField m_thick;
-    HField m_conc;
     ZField m_tice;
     HField m_snow;
     HField m_damage;
     double m_dt;
 
+    // Full DG component arrays of thickness and concentration
+    AdvectionField hiceAdvection;
+    AdvectionField ciceAdvection;
```
einola: My impulse was to write AdvectedField instead of AdvectionField. I suppose I'd call u and v AdvectionField, since it's the field along wich everything is advected.

core/src/include/PrognosticData.hpp
```diff
@@ -41,47 +41,29 @@ class PrognosticData : public ModelComponent, public Configured<PrognosticData>
...
      * @param filePath the file path to write the restart file to.
      */
     void writeRestartFile(const std::string& filePath, const ModelMetadata& metadata) const;
 
 private:
-    HField m_thick;
-    HField m_conc;
     ZField m_tice;
     HField m_snow;
     HField m_damage;
```
einola, line 60: These should also become AdvectionField(s) and be advected, but that may be for a different PR.

core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -108,26 +108,28 @@ void BBMDynamics::setData(const ModelState::DataMap& ms)
...
     std::cout << tst.start << std::endl;
 
     // Fill the updated damage array with the initial value
-    damage = damage0.data();
+    damage = damage0;
 
     // set the updated ice thickness, concentration and damage
-    kernel.setData(hiceName, hice.data());
-    kernel.setData(ciceName, cice.data());
     kernel.setData(damageName, damage);
```
einola, line 125: We need `setData` here (and `getData` below) because m_damage is HField and not AdvectionField in PrognosticData - right?

Then that's a good reason to change the type of m_damage, because using damage should be the default.

core/src/modules/DynamicsModule/MEVPDynamics.cpp
```diff
@@ -86,23 +86,23 @@ void MEVPDynamics::update(const TimestepTime& tst)
     std::cout << tst.start << std::endl;
 
     // set the updated ice thickness and concentration
-    kernel.setData(hiceName, hice.data());
-    kernel.setData(ciceName, cice.data());
+    kernel.setData(hiceName, hice);
+    kernel.setData(ciceName, cice);
```
einola: Why `kernel.setData` and `getData` still here?

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
```diff
@@ -41,17 +41,17 @@ class FreeDriftDynamics : public IDynamics {
         std::cout << tst.start << std::endl;
 
         // set the updated ice thickness and concentration
-        kernel.setData(hiceName, hice.data());
-        kernel.setData(ciceName, cice.data());
+        kernel.setData(hiceName, hice);
```
einola: Why `kernel.setData` and `getData` still here?

einola (approved):
Not an in-depth review, but this looks good to me.

I suspect you forgot to update the VP and freedrift modules, and we need to do something about the advection of snow, temperature and damage.

timspainNERSC:
> I suspect you forgot to update the VP and freedrift modules

Yes, I did!

> and we need to do something about the advection of snow, temperature and damage.

Coming soon, as a separate issue and PR.

timspainNERSC:
> I suspect you forgot to update the VP and freedrift modules

Yes, I did!

> and we need to do something about the advection of snow, temperature and damage.

Coming soon, as a separate issue and PR.

core/src/include/ModelArrayRef.hpp
timspainNERSC: Since `ModelArrayRef` is exclusively used in the thermo/column physics part of the model, I have made the decision that all access through a `ModelArrayRef` should be to the DG0 component only, except when using the `allComponents()` member function. Currently we don't care about the higher orders in the column physics. If this fact ever changes, then yes, this will have to be changed in the future.

core/src/include/PrognosticData.hpp
timspainNERSC: Agreed, `AdvectedField` is a better name.

core/src/include/PrognosticData.hpp
timspainNERSC: Indeed, a different PR.

core/src/modules/DynamicsModule/BBMDynamics.cpp
timspainNERSC: Perhaps more accurately, the damage is not yet a `DGVectorHolder` in the dynamics. But it amounts to the same thing. Other advected fields are my next target.

core/src/modules/DynamicsModule/MEVPDynamics.cpp
timspainNERSC: The other dynamics implementations were forgotten. Updated to the new system.

core/src/modules/DynamicsModule/include/FreeDriftDynamics.hpp
timspainNERSC: The other dynamics implementations were forgotten. Updated to the new system.

timspainNERSC:
> Looks good although it's a bit difficult since many files are touched. But interface to transport seems ok.
> 
> It might be good to do a simple speed-check before and after applying such spaces to see, if there are any efficiency issues.

It should be (slightly) faster, as there is no longer a copy in each direction when going to and from the dynamics.

core/src/include/ModelArrayRef.hpp
timspainNERSC: Added a comment to that effect to the DOxygen documentation of ModelArrayRef.

PR #794, opened 2025-02-13
fixed inconsistent nSteps usage (#792)

### LINKED ISSUE #792: inconsistent number of sub-steps
There are currently two different variables in the dynamics which define the number of sub-steps for mEVP and BBM. The loops use `VPParameters::nSteps` and `BBMParameters::nSteps` respectively, but the sub time-step and average velocity computation in BBM uses `DynamicsKernel::nSteps`. Per default their values are different, at 120 for the param structs and 100 for the DynamicsKernel. This should lead to an overestimate of `avgU` and `avgV` in BBM. 

I think `DynamicsKernel::nSteps` can simply be removed at this point.

Thanduriel:
# fixed inconsistent nSteps usage

Fixes #792

---
# Change Description

Removed `DynamicsKernel::nSteps` and replaced it with `nSteps` from the parameter structs where it was still used in `BrittleCGDynamicsKernel`. This changes the behavior off BBM slightly since `avgU` and `avgV` are weighted differently now.

### DIFF
--- a/dynamics/src/include/BBMDynamicsKernel.hpp
+++ b/dynamics/src/include/BBMDynamicsKernel.hpp
@@ -17,7 +17,6 @@ namespace Nextsim {
 
 template <int DGadvection> class BBMDynamicsKernel : public BrittleCGDynamicsKernel<DGadvection> {
 public:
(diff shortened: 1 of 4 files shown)


einola (approved):
Good catch!

PR #797, opened 2025-02-26
A polynya test case

### LINKED ISSUE #375: Polynya test case
A simple polynya setup from Bjornsson et al. (2001). It is good in that it tests both dynamics and thermodynamics at the same time.

The setup is a 50 times 100 km box, with one of the 50 km boundaries open and the other closed (coast). A wind (10 m/s) blows at an angle so that the ice drifts both out of the open boundary and onto the coast. Ice piles up at the coast and then moves out of the domain, making room for new ice that is created in the "polynya" that forms.

Image
Figure: The setup from Bjornsson et al. (2001).

Image
Figure: The results of one of the runs from Bjornsson et al. (2001).

einola:
# A polynya test case

Fixes #375 

### Task List

---
# Change Description

This PR provides a polynya test case (as per issue #375). For this to work, we also had to update the handling of open boundaries. The ice now flows freely out of any open boundaries. Regarding code changes, the most important changes are to the boundary handling, while the polynya test-case setup is valuable as a test for the new boundaries. This PR also includes minor fixes and upgrades to the thermodynamics.

---
# Test Description

The configuration in ``config_polynya.cfg`` returns results visually similar to those already published (see issue #375).

---
# Documentation Impact

None.

---
# Other Details

None

---
### Pre-Request Checklist


physics/src/IceGrowth.cpp
timspainNERSC, line 1: Since the bulk of this PR covers dynamics and boundary conditions, these changes should be a separate PR, especially as they don't change the numerical outcome of the `IceGrowth` calculations (I think).

timspainNERSC: And if they do change the numerical output, that value should be tested in the `IceGrowth` test.

core/test/PrognosticData_test.cpp
timspainNERSC, line 1: These changes do not directly impact the polynya test, and should be a separate PR

physics/src/modules/AtmosphereBoundaryModule/FluxConfiguredAtmosphere.cpp
```diff
@@ -108,4 +108,11 @@ void FluxConfiguredAtmosphere::setData(const ModelState::DataMap& dm)
     vwind = v0;
 }
 
+void FluxConfiguredAtmosphere::update(const TimestepTime& tst)
+{
+    /* The open water heat flux is reset by the thermodynamics, so that the (slab) ocean doesn't
+     * super cool. Therefore, we have to reset it here on every update */
+    qow = qow0;
+}
+
```
timspainNERSC: Is this required for the polynya test? It should go in the PR with the `IceGrowth` changes if it is not required for the polynya test to pass.

physics/src/modules/AtmosphereBoundaryModule/ConfiguredAtmosphere.cpp
timspainNERSC, line 1: If these changes aren't required by the polynya test, they should go in a separate PR. (I think in this case they might be required to force the polynya dynamics correctly).

physics/src/modules/AtmosphereBoundaryModule/include/ConfiguredAtmosphere.hpp
timspainNERSC, line 1: See comment on the corresponding `.cpp` file.

physics/src/modules/AtmosphereBoundaryModule/include/FluxConfiguredAtmosphere.hpp
```diff
@@ -42,8 +42,8 @@ class FluxConfiguredAtmosphere : public IAtmosphereBoundary,
     ConfigMap getConfiguration() const override;
 
 protected:
-    //! Performs the implementation specific updates. Does nothing.
-    void update(const TimestepTime&) override { }
+    //! Performs the implementation specific updates.
+    void update(const TimestepTime&) override;
```
timspainNERSC: See comment on the corresponding .cpp file.

timspainNERSC (changes requested):
Some irrelevant changes coming along for the ride.

einola:
Merging develop into this branch borked the whole thing - we now have instability on the boundary again, and I don't understand why.

physics/src/modules/AtmosphereBoundaryModule/ConfiguredAtmosphere.cpp
einola: Yes, these are not relevant or necessary, so I'm reverting them.

physics/src/modules/AtmosphereBoundaryModule/FluxConfiguredAtmosphere.cpp
einola: The configuration can use either ``ConfiguredAtmosphere`` or ``FluxConfiguredAtmosphere``. So far, I've used ``FluxConfiguredAtmosphere`` because it's simpler, but in the literature, people have used the equivalence of ``ConfiguredAtmosphere``, so I've been switching between the two.

The bottom line is that both need these minor fixes for the configuration to work, so I think it's okay to include both in the PR, even though only one will be used in the "official" setup.

einola:
Now contains only directly polynya-related changes (i.e. config and init files, and an integrtion test)

run/make_init_polynya.py
```diff
@@ -0,0 +1,50 @@
...
+x = 100
+y = 50
+res = 2
+
+
+nfirst = int(y / res)
+nsecond = int(x / res)
+nLayers = 3
+
+fname = f"init_polynya.nc"
```
joewallwork: Redundant f-string.
```suggestion
fname = "init_polynya.nc"
```

run/make_init_polynya.py
```diff
@@ -0,0 +1,50 @@
...
+# Creates initial conditions for the Bjornsson et al. (2001) polynya case
+
+# Domain size [km]
+x = 100
+y = 50
+res = 2
+
+
+nfirst = int(y / res)
+nsecond = int(x / res)
```
joewallwork: How about integer division specifically, to avoid the type conversion?
```suggestion
nfirst = y // res/
nsecond = x // res
```

test/PolynyaIntegration_test.py
```diff
@@ -0,0 +1,185 @@
+import os
+import subprocess
+import sys
+import unittest
+
+import netCDF4
+import numpy as np
+
+
+class Polynya(unittest.TestCase):
```
joewallwork, line 10: It'd be useful to have a docstring that gives an overview of what the test case is. It doesn't necessarily need to go into details, but could provide a link to where to get more info. e.g., the paper referenced elsewhere.

test/PolynyaIntegration_test.py
```diff
@@ -0,0 +1,185 @@
...
+        # Creates initial conditions for the Bjornsson et al. (2001) polynya case
+
+        # Domain size [km]
+        x = 100
+        y = 50
+        res = 4
+
+
+        nfirst = int(y / res)
+        nsecond = int(x / res)
```
joewallwork:
```suggestion
        nfirst = y // res
        nsecond = x // res
```

test/PolynyaIntegration_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+        fname = f"init_polynya.nc"
```
joewallwork:
```suggestion
        fname = "init_polynya.nc"
```

test/PolynyaIntegration_test.py
```diff
@@ -0,0 +1,185 @@
...
+        ice_salinity = 5  # should match Ice::s in constants.hpp
+        mu: float = -0.055  # should match Water::mu in constants.hpp
+        ocean_temperature = -1.54
+        ocean_salinity = ocean_temperature / mu
+
+        initializer.sss[:, :] = ocean_salinity
+        initializer.sst[:, :] = ocean_temperature
+        initializer.tsurf[:, :] = ice_salinity * mu
+        initializer.tbott = initializer.tsurf
+        initializer.tintr = initializer.tsurf
```
joewallwork: This code seems to be duplicated in `make_init_polynya.py`. It'd be nice to avoid that, if possible.

joewallwork (changes requested):
Thanks, this mostly looks good but I have some minor suggestions and a slightly bigger one to avoid duplicated code.

Note that the changes I'm making in #889 will rework the approach taken for scripts in the `run` directory, but I'm happy to account for this new addition later in my changes.

run/make_init_polynya.py
einola: Done

run/make_init_polynya.py
einola: Done

test/PolynyaIntegration_test.py
einola: Done

test/PolynyaIntegration_test.py
einola: Done

test/PolynyaIntegration_test.py
einola: Done

test/PolynyaIntegration_test.py
einola: Yes, but I'm not sure how to do that sensibly. The test in ``test/PolynyaIntegration_test.py``should not be touched, while the script in ``run/make_init_polynya.py`` should be an example of how to run one of the cases in the Bjornsson et al paper. So, I don't see how one could reuse code from the other.

einola:
Funnily enough, the test now fails because of changes from PR #884. So, if this current PR had been merged first, then we wouldn't have merged #884 the way it was! I've asked Thomas to take a look!

einola:
The tests should pass again after #893 is merged into develop.

test/PolynyaIntegration_test.py
joewallwork: That's true, I suppose it'd need to be a more significant refactor than would be worth doing here.

joewallwork (approved):
Thanks for the updates to this @einola, I'm happy with it now.

einola:
@timspainNERSC: Can you have a look at this one? I can't merge without your approval, because you asked for changes a while ago.

run/make_init_polynya.py
```diff
@@ -0,0 +1,49 @@
...
+# Creates initial conditions for the Bjornsson et al. (2001) polynya case
+
+# Domain size [km]
+x = 100
+y = 50
+res = 2
+
+nfirst = y // res
+nsecond = x // res
+nLayers = 3
```
timspainNERSC, line 12: Don't need this any more!

timspainNERSC (approved):
Looks good, but the `nLayers` variable in the python script that creates the initial file is no longer needed.

PR #804, opened 2025-03-19
Halo integration

TomMelt:
# WIP

### Task List
- [x] integrates Halo exchange logic into the main code base
- [x] re-work the halo test
- [ ] doesn't work for Z fields but these are being removed shortly

### DIFF
--- a/core/src/include/ModelArray.hpp
+++ b/core/src/include/ModelArray.hpp
@@ -1,7 +1,7 @@
 /*!
  * @file   ModelArray.hpp
  *
- * @date   31 Oct 2024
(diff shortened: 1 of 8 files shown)


core/src/include/ModelArray.hpp
```diff
@@ -638,8 +658,11 @@ class ModelArray {
         std::map<Type, MultiDim> m_dimensions;
     };
     static DimensionMap m_dims;
+
+public:
```
TomMelt, line 723: this will be removed and is for testing purposes only

PR #805, opened 2025-03-27
Ice ocean stress computation

### LINKED ISSUE #791: ice ocean stress interface
The current interface for the ice ocean stress computation, i.e. `CGDynamicsKernel::getIceOceanStressElement` and ``CGDynamicsKernel::getIceOceanStress` is not ideal. The combination of a virtual function + branch based on string comparisons in the inner loop makes it very inefficient. Furthermore, the abstraction on the level of per-element computations makes it impossible to perform these computations on the GPU.

Both problems could be solved by having the virtual function that is implemented by each kernel compute the whole field. It should still be possible to implement shared parts once in `CGDynamicsKernel` . Looking at the concrete implementations for Brittle and VP dynamics they only differ by the velocity used. How much flexibility  is really needed @einola ? Can the formula change significantly? Also, is there a reason why x and y fields are computed separately? I guess it is because of the getData interface?

Thanduriel:
# Ice ocean stress computation

Fixes #791

---
# Change Description

Reworked the ice ocean stress computation to facilitate greater reuse. The implementation is similar to the [previous one](https://github.com/nextsimhub/nextsimdg/pull/664) by @timspainNERSC, with some important differences:

1. **Persistent buffers** Previously, the ice ocean stresses where computed just-in-time in `getDG0Data`. This is not ideal since most of the computations are shared between the u and v fields. Now they are computed as part of  the kernel `update` and stored as `CGDynamicsKernel::uIceOceanStress` and `CGDynamicsKernel::vIceOceanStress`. If this is not desired, I think we should change the kernel-interface to get these fields instead. Alternatively, we could also do the interpolation right away and only store the `DGModelArray`. Is the full `CGVector` every needed?
2. **No virtual function** The only function I introduced is `CGDynamicsKernel::updateIceOceanStress` which contains the implementation and is called by the concrete kernels with the correct velocities. The only reason for introducing a virtual function for this computation is, in my mind, if we want to introduce additional logic around it to only compute ice-ocean stresses some of the time. This could then be implemented in the base class `DynamicsKernel`.
3. **Parameter initialization** There is a pitfall with the initialization of the parameter structs. They are being passed to the kernel constructors before being properly initialized by `configure`. I changed the initialization order in the modules so that the param structs are at least default initialized before being passed to the kernel constructors, elliminating the undefined behavior. Still, since the proper parameter configuration happens later, computations based on the parameters should be be done in `initialise()`. Where I noticed the incorrect usage (e.g. `FreeDriftDynamicsKernel`) I delayed the initialization and set the defaults to NaN so that improper usage is detected quickly. However, the overall design still has the potential for subtle bugs later on, when there are actual user provided parameters that differ from the default values. Ideally, initialization of the struct would be finished by the point it becomes accessible to the kernel but I guess this is not possible with the overall structure. Alternatively, such inconsistencies could be prevented by passing the parameter structs to `initialise` instead of the constructors which in turn would necessitate the use of pointers which have their own shortcomings. So I guess, one just has to be careful.

Furthermore, I did some minor cosmetic changes in the files I touched.

---
# Test Description

I checked that `taux` and `tauy` are the same for `BBMDynamics`. For `MEVPDynamics` the results are not expected to be the same since the ocean angle is used now, which is not yet properly implemented in the other computations.

### DIFF
--- a/core/src/modules/DynamicsModule/include/BBMDynamics.hpp
+++ b/core/src/modules/DynamicsModule/include/BBMDynamics.hpp
@@ -1,7 +1,7 @@
 /*!
  * @file BBMDynamics.hpp
  *
- * @date 19 Nov 2024
+ * @date 27 Mar 2025
  * @author Tim Spain <timothy.spain@nersc.no>
  */
 
@@ -56,8 +56,8 @@ public:
     static HelpMap& getHelpRecursive(HelpMap&, bool getAll);
 
 private:
-    BBMDynamicsKernel<DGCOMP> kernel;
     BBMParameters params;
+    BBMDynamicsKernel<DGCOMP> kernel;
 };
 
 } /* namespace Nextsim */
--- a/core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
+++ b/core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
@@ -1,7 +1,7 @@
 /*!
  * @file MEVPDynamics.hpp
  *
- * @date 19 Nov 2024
+ * @date 27 Mar 2025
  * @author Tim Spain <timothy.spain@nersc.no>
  * @author Piotr Minakowski <piotr.minakowski@ovgu.de>
  * @author Einar Ólason <einar.olason@nersc.no>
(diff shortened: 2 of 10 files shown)


einola:
This looks quite good to me.

~~It's a good point about ``CGVector``. The full ``CGVector`` is never needed; the ice-ocean stress is only used when coupling to the ocean or in the diagnostic output. So, as you suggest, we could do the interpolation right away and only store the ``DGModel Array``. Can you do that?~~ -- This was a bad idea. I prefer it the way it is now!

einola (approved):

PR #808, opened 2025-04-09
bugfix: call Xios::finalize() explicitly

TomMelt:
(This is related to the static data issue raised in #614)

The tests for PR #763 were failing due to the following problem: (explanation adjusted with feedback from @Thanduriel)

* In the previous version, the pointer to the `Xios` object was static and the `Xios` object was never destroyed i.e., the destructor was not called.
* In the new version, the `Xios` destructor (`Xios::~Xios() { finalize(); }` is called when the program terminates
* `finalize()` calls `cxios_finalize()` which eventually calls `MPI_Comm_rank`
This leads to the error:
```
*** The MPI_Comm_rank() function was called after MPI_FINALIZE was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```

**attempted solution**

By manually calling `Xios::finalize()` at the end, we can safely terminate XIOS and also set `isEnabled=False` to stop the code trying to call `finalize()->cxios_finalize()->MPI_Comm_rank` again.

**better solution**

Whilst this PR works, a better solution would be to not use a `static` object to store the singleton (see #674).


joewallwork (approved):
Amazing, thanks @TomMelt! Given that this is a special interaction between XIOS and doctest, I personally think it's fine to have these in for the tests that are set up with doctest. Users wouldn't need to explicitly call the finalizer in their code (unless they also use doctest or have some other special MPI finalization, of course).

TomMelt:
> Amazing, thanks @TomMelt! Given that this is a special interaction between XIOS and doctest, I personally think it's fine to have these in for the tests that are set up with doctest. Users wouldn't need to explicitly call the finalizer in their code (unless they also use doctest or have some other special MPI finalization, of course).

Hmmm I don't think it's just for `doctest` but I could be wrong. I think it will apply to `nextsim` as well. Because we will call `MPI_Finalize()` at the end of `nextsim` but the destructor won't get called until the very end of the program leading to the same issue.

Thanduriel:
This is once again the lifetime problem with static data, see [https://github.com/nextsimhub/nextsimdg/issues/614](https://github.com/nextsimhub/nextsimdg/issues/614). So it will be an issue in the main program as well. I believe the `Finalizer` functionality was introduced to solve this ([https://github.com/nextsimhub/nextsimdg/pull/674](Finalizer)), and should be used here as well, @timspainNERSC ?

While your analysis of the behavior is largely correct, I must address this gross misuse of technical terms :grinning: . First of all there is no GC, the concept of object lifetimes in C++ is called RAII (resource acquisition is initialization) and it is compile time. Destructors are simply called when a variable goes out of scope, which for static lifetimes means on program end. Furthermore, static variables are stored in neither stack nor heap. They are stored in the DATA segment which is part of the loaded program. 

Finally, in the previous version the `Xios` object was dynamically allocated and thus lived on the heap. However, the instance pointer was static like the whole object is now, and while any copies of the pointer acquired through `getInstance` would likely be on the stack, the lifetime of the `Xios` object would be unaffected. Raw pointers have no concept of ownership and the use of smart pointers in their place would have caused the same issue as we have now. In the old version the problem was probably not visible because the object was never destroyed in the first place.

TomMelt:
> This is once again the lifetime problem with static data, see #614. So it will be an issue in the main program as well. I believe the `Finalizer` functionality was introduced to solve this ([https://github.com/nextsimhub/nextsimdg/pull/674](Finalizer)), and should be used here as well, @timspainNERSC ?
> 
> While your analysis of the behavior is largely correct, I must address this gross misuse of technical terms 😀 . First of all there is no GC, the concept of object lifetimes in C++ is called RAII (resource acquisition is initialization) and it is compile time. Destructors are simply called when a variable goes out of scope, which for static lifetimes means on program end. Furthermore, static variables are stored in neither stack nor heap. They are stored in the DATA segment which is part of the loaded program.
> 
> Finally, in the previous version the `Xios` object was dynamically allocated and thus lived on the heap. However, the instance pointer was static like the whole object is now, and while any copies of the pointer acquired through `getInstance` would likely be on the stack, the lifetime of the `Xios` object would be unaffected. Raw pointers have no concept of ownership and the use of smart pointers in their place would have caused the same issue as we have now. In the old version the problem was probably not visible because the object was never destroyed in the first place.

thanks I will edit the PR to reflect your comments

TomMelt:
superseded by #822

PR #815, opened 2025-04-25
The land mask is wrong in the restart file

### LINKED ISSUE #814: Land mask all ones in restart file
The restart files written at the end of the run and the periodically written restart files contain a land mask (called "mask"), but this is always all ones.

### LINKED ISSUE #819: Consistent use of overElements
Some of the physics routines are simple enough so that we don't have to use the overElements function - we can simply use Eigen array arithmetic. The problem is that if we do this, we end up doing work on land points and may introduce spurious values there. We should, therefore, always use overElements in the physics calculations.

einola:
# The land mask is wrong in the restart file

Fixes #814 

### Task List

---
# Change Description

I fixed this by removing the call to ``noLandMask()`` in the ``ModelComponent()`` constructor (and making it a ``default`` constructor).

---
# Test Description

I added a test of the land mask to the realistic integration tests. That test now fails if two pre-defined points are not defined as land and ocean.

---
# Documentation Impact

N/A

---
# Other Details

None
---
### Pre-Request Checklist

### DIFF
--- a/core/src/ModelComponent.cpp
+++ b/core/src/ModelComponent.cpp
@@ -11,10 +11,15 @@
 
 namespace Nextsim {
 
-size_t ModelComponent::nOcean;
+size_t ModelComponent::nOcean = 0;
 std::vector<size_t> ModelComponent::oceanIndex;
 
-ModelComponent::ModelComponent() = default;
+ModelComponent::ModelComponent()
+{
+    // We only set no land mask if the mask hasn't been set by someone else.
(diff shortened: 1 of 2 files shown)


einola:
Update: Using ``default`` constructor caused some problems with the tests. So, I changed the approach and call ``noLandMask()`` in the constructor if ``nOcean`` hasn't been set.

timspainNERSC (approved):

PR #820, opened 2025-04-29
Use overElements in IOceanBoundary and derived classes

einola:
# Use overElements in IOceanBoundary and derived classesFixes #819 

### Task List

---
# Change Description

This changes IOceanBoundary::mergeFluxes and SlabOcean::update to use overElements, instead of Eigen array arithmetics.

---
# Test Description

No additional tests needed.

---
# Documentation Impact

N/A

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/physics/src/SlabOcean.cpp
+++ b/physics/src/SlabOcean.cpp
@@ -1,7 +1,7 @@
 /*!
  * @file SlabOcean.cpp
  *
- * @date 10 Feb 2025
+ * @date 29 Apr 2025
  * @author Tim Spain <timothy.spain@nersc.no>
  */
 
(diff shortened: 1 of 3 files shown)


timspainNERSC:
What is the advantage of using `overElements()` in the `update()` methods of these classes?
I only came up with the `overElements` method as handling conditional logic in the algorithms using `Eigen::Array` native methods made my head spin. If there is no conditional logic in the code (as there is not with these two classes), then the arithmetic that Eigen provides should be more efficient. Using `overElements` certainly doesn't make the code more readable, either, in my opinion.

einola:
The main advantage is that ``overElements`` only does work on wet points, not over land.

In a global setup, ``overElements`` should then be a good deal more efficient.

I also quite like ``overElements``. In theory, it's the only place we loop over the domain in the thermodynamics, so if everything uses ``overElements``, doing an OpenMP parallelisation of that should be a breeze. While I do agree that the changes I made decreased readability, then ``overElements`` is more readable than many other looping solutions I've seen.

einola:
For OpenMP, I also vaguely remember one large loop being preferred over many small ones. That would mean that for OpenMP parallelised code, ``overElements`` would be preferred over ``Eigen::Array`` native methods. But I may be wrong about that, that may be outdated information, or Eigen may handle it internally by turning repeated operations into one big loop.

timspainNERSC (approved):

PR #822, opened 2025-04-29
Set up `Finalizer` for XIOS

joewallwork:
Supersedes #808.

As discussed in the meeting this morning, this PR hooks the `Xios` class into the `Finalizer`. Since no `Model`s are created in the standalone XIOS unit tests, they need to call `Finalizer::finalize` explicitly. It seems the same is also true of the `ConfigOutput` and `ParaGrid` tests. So actually the changeset isn't that different than @TomMelt's one in #808.


joewallwork:
[Rebased on top of `issue756_xios-calendar-increment-singleton`.]

TomMelt (approved):
Thanks @jwallwork23 it's better than my attempt :wink:

PR #824, opened 2025-04-30
Configure XIOS from `.cfg` file

joewallwork:
# Configure XIOS from `.cfg` file

Fixes #747

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

This PR automates XIOS usage so that the timestep, start time, and output frequency are deduced from the configuration files used by nextSIM-DG. As such, the only interaction the typical user should have with XIOS is to specify whether or not to use it in the build.

---
# Test Description

A `xios_tests.cfg` configuration file is provided to account for the XIOS unit tests. The timestep and start time values used in the tests were tweaked so that the same configuration file could be used for all such tests.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] File dates have been updated to reflect modification date
- [x] This change conforms to the conventions described in the README

PR #832, opened 2025-05-07
Xios config output fix

joewallwork:
I noticed that the `ConfigOutput` test failed upon merging #824 into `develop`. (See https://github.com/nextsimhub/nextsimdg/actions/runs/14859539094/job/41720738522.)

I think this PR should fix it, by fully removing XIOS from the `ConfigOutput_test`, as recommended by @timspainNERSC.

I also added in a minor fix for the `XiosRead_test`, which used `HField` where it should've used `ZField`.

### DIFF
--- a/core/test/XiosRead_test.cpp
+++ b/core/test/XiosRead_test.cpp
@@ -107,7 +107,7 @@ MPI_TEST_CASE("TestXiosRead", 2)
     // Create HField instances to read the data into
     HField field_2D(ModelArray::Type::H);
     field_2D.resize();
-    HField field_3D(ModelArray::Type::Z);
+    ZField field_3D(ModelArray::Type::Z);
     field_3D.resize();
 
     // Check calendar step is zero initially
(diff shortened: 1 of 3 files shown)


joewallwork:
Ah, another potential issue was the inconsistency between the new workflow and the other MPI one - sorry about this. See 156ea3847be1fd5511ad06d7fd01acc9eb631d01.

core/test/XiosRead_test.cpp
```diff
@@ -107,7 +107,7 @@ MPI_TEST_CASE("TestXiosRead", 2)
     // Create HField instances to read the data into
     HField field_2D(ModelArray::Type::H);
     field_2D.resize();
-    HField field_3D(ModelArray::Type::Z);
+    ZField field_3D(ModelArray::Type::Z);
```
timspainNERSC, line 110: Technically, this is purely a textual error. In this context, `ZField` and `HField` both resolve to `ModelArray`. Also, don't forget that ZField is being removed entirely Real Soon Now™.

timspainNERSC (approved):
Approved, despite giving me more `ZField`s that need to be removed.

PR #841, opened 2025-05-12
Issue840 Fixing `ConfigOutput_test` to be deterministic

### LINKED ISSUE #840: ConfigOutput_test depends on an uninitialized value.
One of the things tested as part of `ConfigOutput_test` is the number of fields written to the files. Due to the way the `ModelMetadata` class handles the coordinate data, the number of fields provided when `ModelMetadata::affixCoordinates()` is called depends on the structure of the data processed by `ModelMetadata::extractCoordinates()`. Specifically, this method checks whether the data contains a field named `coords`, which indicates that the grid has parameterized coordinates that are not just the x and y array indices. If an array with that name exists in the state the coordinates are extracted from then the flag `ModelMetadata::hasParameters` is set to `true`, else it is set to false. Currently `ConfigOutput_test` does not call `extractCoordinates()` and `hasParameters` does not have a default value. This means that whether `ConfigOutput_test` passes depends on an uninitialized value.

timspainNERSC:
# Fixing `ConfigOutput_test` to be deterministic

Fixes #840

---
# Change Description

Set the `bool` values `ModelMetadata::isCartesian` and `ModelMetadata::hasParameters` to be initially `false`.

---
# Test Description

Added a call to `ModelMetadata::extractCoordinates()` to `ConfigOutput_test.cpp`, so that the model state that is to be written definitely has `coords` and `grid_azimuth` fields, even if they are filled with zeros.

### DIFF
--- a/core/src/include/ModelMetadata.hpp
+++ b/core/src/include/ModelMetadata.hpp
@@ -110,9 +110,9 @@ private:
     // Angle from model reference to grid north (+y for grids) TODO: what for meshes? N/A?
     ModelArray m_gridAzimuth;
     // Are the coordinates Cartesian? x & y versus longitude and latitude
-    bool isCartesian;
+    bool isCartesian = false;
     // Are the more complex coordinates stored?
-    bool hasParameters;
+    bool hasParameters = false;
 #ifdef USE_MPI
     const std::string bboxName = "bounding_boxes";
 #endif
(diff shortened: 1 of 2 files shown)


joewallwork (approved):
Ah okay, good spot! Thanks @timspainNERSC

PR #843, opened 2025-05-12
Issue821: Remove `ZField`

### LINKED ISSUE #821: Removing `ZField`: definitions
With `ZField` being unused in all branches where #801 is fixed, the definition of `ZField` in headers and the special functions for handling them in `ModelArray`(`zIndexAndLayer()`) can be removed.

timspainNERSC:
# Remove `ZField`

Fixes #821 

---
# Change Description

Removes the three-dimensional `ZField` from the list of `ModelArray` types in nextSIM-DG. This simplifies advection and communication between the column physics and dynamics parts of the model. The previously three-dimensional ice temperature field has been replaced by the two-dimensional ice surface temperature field for general thermodynamics models. The thermodynamics implementation based on that of Winton previously had a three vertical level, three-dimensional ice temperature field. This is replaced by the above mentioned surface temperature field and two other two dimensional fields, representing the interior ice temperature and the temperature of the bottom of the ice pack.

This represents the first example of a prognostic field used by a module implementation. The Winton thermodynamics, specifically, requires all three ice temperature fields to be present for a perfect restart. The other thermodynamics implementations can neither read nor write these fields. The restart file must be able to contain these conditional prognostic fields. With these changes, it now can.

As part of this, the `getState()` and `getStateRecursive()` functions were rationalized to accuratelt reflect their use as the functions to get prognostic (now `getStatePrognostic()`) and diagnostic (now `getStateDiagnostic()`) fields. Along with this refactor, the available diagnostic fields have expanded. Previously these had been all fields available through the `ModelArrayRef` data store. However, components of the model can have their own internal state that needn't be available through `ModelArrayRef`, but should be available to be written to the diagnostic output. Almost every field that is a ModelArray member variable is now available through the diagnostic output.

---
# Test Description

- The test of ConfigOutput was edited to reflect the new behaviour of the diagnostic output.
- The test of ThermoWinton was edited to test the two new temperature fields.

---
# Documentation Impact

The names of the newly available diagnostic fields should be documented somewhere other than the source code. Some more thought is needed here. Hopefully the solution will become clearer as more fields are moved out of `PrognosticData`. Solving this is not in the scope of this PR


TomMelt:
Thanks @timspainNERSC  this is great. I will start reviewing it now. Would you be able to resolve the conflicts in the meantime?

core/src/ModelArray.cpp
```diff
@@ -234,6 +231,11 @@ double& ModelArray::operator[](const MultiDim& dims)
 
 ModelArraySlice ModelArray::operator[](const Slice& slice) { return ModelArraySlice(*this, slice); }
 
+ConstModelArraySlice ModelArray::operator[](const Slice& slice) const
+{
+    return ConstModelArraySlice(*this, slice);
+}
+
```
TomMelt, line 238: Are the `ConstModelArraySlice` additions relevant to this PR?

core/src/include/ModelArray.hpp
```diff
@@ -164,6 +165,7 @@ class ModelArray {
      * Creates a ModelArraySlice.
      */
     ModelArraySlice operator[](const Slice&);
+    ConstModelArraySlice operator[](const Slice&) const;
```
TomMelt, line 168: > Are the `ConstModelArraySlice` additions relevant to this PR? 

(Also here)

core/src/include/ModelArraySlice.hpp
```diff
@@ -288,6 +290,27 @@ class ModelArraySlice {
...
+    ConstModelArraySlice(const ModelArray& ma, const Slice& sl)
+        : data(ma)
+        , slice(sl)
+    {
+    }
+
+    operator const ModelArraySlice()
+    {
+        return ModelArraySlice(const_cast<ModelArray&>(data), slice);
+    }
```
TomMelt, line 308: > Are the `ConstModelArraySlice` additions relevant to this PR? 

and here

core/src/modules/include/ProtectedArrayNames.ipp
```diff
@@ -12,6 +12,7 @@
     { "cice", "C_ICE0" }, // Ice concentration
     { "hsnow", "H_SNOW_cell" }, // Snow depth, cell average, m
     { "tice", "T_ICE0" }, // Ice temperature, ˚C
+    { "tsurf", "T_SURF0" }, // Ice surface temperature, ˚C
```
TomMelt, line 15: do the others need adding here i.e., `tbottom` and `tinterior`?

physics/test/ThermoWintonTemperature_test.cpp
```diff
@@ -45,17 +44,17 @@ TEST_CASE("Melting conditions")
     Configurator::addStream(std::move(pcstream));
 
     ThermoWinton twin;
+
+#define TICE -1. // Cannot reference local var in the function definition
```
TomMelt, line 48: can you explain why this is necessary as a Macro? I don't quite understand your explanation here.

TomMelt (approved):
Thanks @timspainNERSC . It looks good to me, just a couple of comments. No changes required (other than the outstanding merge conflict resolution)

core/src/ModelArray.cpp
timspainNERSC: They are, as this code introduces a case where a `const ModelArray` needs to be sliced. It would have been better if I had foreseen this when writing the original `ModelArraySlice` class, but I did not.

This change could be broken out into a separate PR, but I don't think that is worth doing.

core/src/modules/include/ProtectedArrayNames.ipp
timspainNERSC: They do not.

physics/test/ThermoWintonTemperature_test.cpp
timspainNERSC: I'd rather define a `const double tice` to be used both inside and outside the `IceTemperatureData` class definition, but C++ does not allow that directly. Using a macro seemed to be the simplest direct method.

core/src/ModelArray.cpp
TomMelt: no that's fine. I just didn't realise from my review

PR #836, opened 2025-05-07
XIOS: Create files from config upon initialisation

joewallwork:
# XIOS: Create files from config upon initialisation

Fixes #835

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

To quote the issue:
> Following on from #834, we can automate file creation by querying the config to determine input and output files.
> ```
> [XiosInput]
> filename = ...
> field_names = ...
> [XiosOutput]
> filename = ...
> field_names = ...
> ```
> It's probably simplest to just use the `filename` as both the `fileId` and the `fileName` passed to XIOS.
> 
> Further, we can automatically create and add all fields associated with such files based off the `field_names` entries.

---
# Test Description

XIOS read, write, and file tests are updated accordingly.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] File dates have been updated to reflect modification date
- [x] This change conforms to the conventions described in the README

PR #844, opened 2025-05-12
XIOS: Create fields from config upon initialisation

### LINKED ISSUE #842: XIOS: Create fields from config upon initialisation
Following on from #835, we can also automate the creation of XIOS fields at startup.

We have
```
[XiosInput]
filename = ...
field_names = ...
[XiosOutput]
filename = ...
field_names = ...
```
so we can automatically create and add all fields associated with such files based off the `field_names` entries.

joewallwork:
# XIOS: Create fields from config upon initialisation

Fixes #842

Follows up on #836 to automate the creation of fields as well as files. Unfortunately, it reduces the `REQUIRE_THROWS` checks we can use in some of the tests. However, I think overall this is the right way to go.

I later edited the PR to switch up how files are set up to make things more streamlined.

I lumped in some other `HField`->`ZField` changes that I missed in #832.

#### Checklist

- [x] Code changes
- [x] Tests pass
- [x] Provide docs


joewallwork:
Pushed an extra commit that sets the field name to match the fieldId, for simplicity. Now ready for review.

joewallwork:
[Rebased on top of `develop`.]

joewallwork:
Converted to draft because I'm taking a step back to think about the best way to set all this up.

joewallwork:
Okay, this is now ready for review. Added the beginnings of some long-awaited docs, which we should continue to add to.

core/test/XiosFile_test.cpp
```diff
@@ -38,10 +38,14 @@ MPI_TEST_CASE("TestXiosFile", 2)
     config << "[model]" << std::endl;
     config << "start = 2023-03-17T17:11:00Z" << std::endl;
     config << "time_step = P0-0T01:30:00" << std::endl;
+    config << "[XiosInput]" << std::endl;
+    config << "period = P0-0T03:00:00" << std::endl;
+    config << "filename = xios_test_input" << std::endl;
```
joewallwork, line 43: Note that any `filename` found in the `XiosInput` section will now get `setFileMode` with `'read'`. This will open the file provided, so the test needs to be set up such that the file exists and has the correct dimensions, fields, etc. This is why the dimensions, fields, etc. change in this test, too.

joewallwork:
[Rebased on top of `develop`, in which removed `ZField`s were removed.]

core/src/Xios.cpp
```diff
@@ -1496,39 +1493,32 @@ void Xios::createFile(const std::string fileId)
...
             + "' found in both the XiosInput and XiosOutput config sections");
         // TODO: Refactor to allow a field to be both read and written
     }
-    if (readAccess) {
-        setFileMode(fileId, "read");
-    } else {
-        setFileMode(fileId, "write");
+
+    // Terminate early for special unit test cases, for which IDs start with 'unittest'
+    if (fileId.rfind("unittest", 0) == 0) {
```
TomMelt, line 1504: :eyes: what do we think is the probability of someone using `unittest` as their fileID :sweat_smile: 

Is it worth adding a `cout` warning just in case?

TomMelt (changes requested):
LGTM with the exception of adding a warning for the `unittest` case.

core/src/Xios.cpp
joewallwork: Added a warning in 4ce6109c9ec382314580573638831c9f2c6a38d4.

TomMelt: Thanks that's great

TomMelt (approved):
Happy with this now that we have a warning about the `unittest` `fileID` :rocket:

PR #845, opened 2025-05-12
Issue734 add ctest functionality

### LINKED ISSUE #734: add `ctest` functionality to NextSim
In the `decomp` tool we use a CMake function `function(create_test)` to add tests to `ctest` (see [here](https://github.com/nextsimhub/domain_decomp/blob/9efa78b0b843cf510e1037fc120982d75f2e7ac5/test/CMakeLists.txt#L1-L26) for example).

This greatly simplifies local and CI testing.

It would be a good idea to add this functionality to NextSim's tests as well.

TomMelt:
fixes #734

I have added functionality to run the tests with `ctest` instead of running them manually with a script. This matches the approach used in the `decomp` tool

e.g.,

```bash
cd build
ctest
```

Or if you want verbose output:
```bash
cd build
ctest -V
```

And you can use regex's to run a subset if you want (e.g., run only the Xios tests):
```bash
cd build
ctest -R Xios
```

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -66,16 +66,7 @@ jobs:
         apt install -y wget
         cd build
         (cd core/test && wget "ftp://ftp.nersc.no/nextsim/netCDF/partition_metadata_1.nc")
-        for component in core physics dynamics
-        do
-            cd $component/test
-            for file in $(find test* -maxdepth 0 -type f)
-            do
-              echo $file
-              ./$file
-            done
-            cd -
-        done
+        ctest -V
         mv ../test .
         cd test
         python3 ThermoIntegration_test.py
@@ -145,17 +136,8 @@ jobs:
         apt install -y wget
         cd build
         (cd core/test && wget "ftp://ftp.nersc.no/nextsim/netCDF/partition_metadata_1.nc")
-        for component in core physics dynamics
-        do
-            cd $component/test
-            for file in $(find test* -maxdepth 0 -type f)
-            do
-              echo $file
-              nprocs=$(echo $file | sed -r "s/.*MPI([0-9]+)/\1/")
-              mpirun --allow-run-as-root --oversubscribe -n $nprocs ./$file
-            done
-            cd -
-        done
+        ctest -V
+        cd -
 
   test-mac-serial:
 
@@ -195,16 +177,7 @@ jobs:
     - name: run serial tests
       run: |
         cd build
-        for component in core physics dynamics
-        do
-            cd $component/test
-            for file in $(find test* -maxdepth 0 -type f | grep -v "_MPI\d*$")
-            do
-              echo $file
-              ./$file
-            done
-            cd -
-        done
+        ctest -V
         mv ../test .
         cd test
         python ThermoIntegration_test.py
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -92,23 +92,18 @@ jobs:
         make -j 4
 
     - name: run MPI tests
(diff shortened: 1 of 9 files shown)


.github/workflows/test_suite.yml
```diff
@@ -145,17 +136,8 @@ jobs:
...
-            for file in $(find test* -maxdepth 0 -type f)
-            do
-              echo $file
-              nprocs=$(echo $file | sed -r "s/.*MPI([0-9]+)/\1/")
-              mpirun --allow-run-as-root --oversubscribe -n $nprocs ./$file
-            done
-            cd -
-        done
+        ctest -V
+        cd -
```
joewallwork:
```suggestion
```

docs/installation.rst
```diff
@@ -219,3 +219,28 @@ build arguments (``--build-arg`` that can be specified at build time). For examp
...
+
+.. code-block:: console
+
+    ctest -V
+
+If you want to run a subset of tests, you can use the `-R` option with a regular expression to match the test names. For example, to run only the tests related to `XIOS`, use:
+
+.. code-block:: console
+
+    ctest -R XIOS
```
joewallwork: Have you tested this? I would've thought it'd need to be
```suggestion
    ctest -R Xios
```

joewallwork (approved):
This is great, thanks @TomMelt! I have a very minor suggested edit and a question about the docs, but otherwise looks good to go.

joewallwork:
Huh, I'd missed the XIOS test failures. Will take a look at those tomorrow.

docs/installation.rst
TomMelt: good catch. I tested but then wrote all caps by mistake :/

TomMelt:
> Huh, I'd missed the XIOS test failures. Will take a look at those tomorrow.

I fixed them. Does it look ok to you now?

joewallwork:
> > Huh, I'd missed the XIOS test failures. Will take a look at those tomorrow.
> 
> I fixed them. Does it look ok to you now?

Yep, thanks!

core/test/CMakeLists.txt
```diff
@@ -282,6 +313,25 @@ else()
...
+    add_test(NAME testModelArrayRef COMMAND testModelArrayRef)
+    add_test(NAME testPrognosticData COMMAND testPrognosticData)
+    add_test(NAME testPrognosticDataIO COMMAND testPrognosticDataIO)
+    add_test(NAME testMonthlyCubicBSpline COMMAND testMonthlyCubicBSpline)
+    add_test(NAME testFileCallbackCloser COMMAND testFileCallbackCloser)
+    add_test(NAME testFinalizer COMMAND testFinalizer)
+    add_test(NAME testModelArray COMMAND testModelArray)
+    add_test(NAME testSlice COMMAND testSlice)
+    add_test(NAME testModelArraySlice COMMAND testModelArraySlice)
+
```
timspainNERSC: Since these are all of the form `add_test(NAME $1 COMMAND $1)`, is there perhaps a way to glob the available testSomething binaries and generate all of these commands automatically?

TomMelt: I dont think so. AFAIK the tests don't exist until they are compiled. So it would be too late. You could possibly write a list of all the test executable names and iterate over that. Would that be better?

TomMelt: I have refactored the tests in commit 69b6d9e6 to use a list. I have also done a similar thing for the MPI tests and refactored the MPI logic into a separate function

timspainNERSC (approved):
Good work on the regex wrangling.

PR #849, opened 2025-05-19
XIOS: Create domains based off partition metadata

### LINKED ISSUE #732: XIOS: Determine Axes and Domains
Thinking more about what's required for #552, I realised there's a key missing piece for reading files.

Usually, XIOS reads in an XML configuration file along with the NetCDF file that contains the data. The XML file says how to set up the various attributes, e.g., Axes, Domains, Fields. If we want to go ahead with using the XIOS API instead of XML files then we need to be able to read such information from NetCDF files directly. This could be done by borrowing from the existing I/O implementation, for example.

On the other hand, am I right in thinking that the only use case for having XIOS read from file is for restarts? In the case of restarts, you presumably already know how the dimensions and fields are laid out so could reproduce the XIOS attributes without needing to read them.

Any thoughts on this @timspainNERSC @einola?

joewallwork:
# XIOS: Create domains based off partition metadata

Closes #732.
Merges into the branch for #844.

This PR builds on the XIOS config work (which automates the setup of Fields and Files) to automate the setup of the Domain. This can be done straightforwardly upon creating the `ModelMetadata` object.

There's no need to set up XIOS Axis objects for now, now that `ZField` is no longer supported. (The Axis was only ever used for the $z$-component.)


joewallwork:
[Rebased on top of `issue842_xios-create-fields`, which was rebased on top of `develop`.]

core/test/CMakeLists.txt
```diff
@@ -32,12 +32,24 @@ if(ENABLE_MPI)
...
+        add_custom_command(
+            OUTPUT xios_test_partition_metadata_2.nc
+            COMMAND
+                ncgen -b -o xios_test_partition_metadata_2.nc
+                ${CMAKE_CURRENT_SOURCE_DIR}/xios_test_partition_metadata_2.cdl
+            DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/xios_test_partition_metadata_2.cdl
+        )
+        add_custom_target(generate_xios_partition_metadata_2 ALL
+            DEPENDS xios_test_partition_metadata_2.nc
+        )
```
TomMelt, line 52: A note to myself more than anything else, but I am planning to tidy the netcdf generation up a bit like I did for the `ctest` PR, just so we can simplify it a bit. My halo work will add a couple more `.nc` files so I definitely want to refactor it, but I will do that in my own PR :)

core/src/ModelMetadata.cpp
```diff
@@ -62,6 +62,20 @@ void ModelMetadata::getPartitionMetadata(std::string partitionFile)
...
+    const std::string domainId = "xy_domain";
+    xiosHandler.createDomain(domainId);
+    xiosHandler.setDomainType(domainId, "rectilinear");
+    xiosHandler.setDomainGlobalXSize(domainId, globalExtentX);
+    xiosHandler.setDomainGlobalYSize(domainId, globalExtentY);
+    xiosHandler.setDomainLocalXStart(domainId, localCornerX);
+    xiosHandler.setDomainLocalYStart(domainId, localCornerY);
+    xiosHandler.setDomainLocalXSize(domainId, localExtentX);
+    xiosHandler.setDomainLocalYSize(domainId, localExtentY);
+#endif
```
TomMelt, line 78: If you agree, I think we should move this back to XIOS once I have merged my PR which makes the `ModelMetadata` a singleton.

So for now, I would suggest leaving it here, but in my `ModelMetadata` PR (*coming soon*) I will move it back to the `ModelMetadata`.

docs/xios.rst
```diff
@@ -60,3 +60,10 @@ values that are parsed, the handler object will automatically create `Field` and
 `File` data structures for XIOS and associate these as appropriate. If the
 `XiosInput` section is used, an attempt will be made to open the file provided
 by the `filename` entry.
+
+XIOS requires information on the domain decomposition for it to be able to read
+and write data in parallel. This information is held by the `ModelMetadata`
+class, which may be constructed based off a partition metadata file. Upon
+construction of this object, if XIOS is enabled then the XIOS handler will
+automatically create a `Domain` data structure with the appropriate local and
+global sizes and offsets.
```
TomMelt, line 69: same comment here as last PR (https://github.com/nextsimhub/nextsimdg/pull/851) about the inline code sections (`) -> (``)

TomMelt (approved):
LGTM. My comments are for discussion but I dont think anything needs to change in this PR

docs/xios.rst
joewallwork: Done in #851 in comment d6de1337. (That PR merges into this one.)

core/src/ModelMetadata.cpp
joewallwork: I looked into moving this myself but then noticed we'd end up with circular dependencies between `Xios` and `ModelMetadata` because of the XIOS code lower down. I gather your concern here is that if `getPartitionMetadata` becomes static then it won't be able to execute the non-static XIOS code block here?

core/test/CMakeLists.txt
TomMelt: #858 issue here

core/src/ModelMetadata.cpp
TomMelt: I will look into this when I get further along with the `ModelMetadata` changes. I think it's more about sensible grouping of code. To me, this feels like code that should be in the `XIOS` class rather than the `ModelMetadata` class.

Issue #850, opened 2025-05-19
XIOS: Prune the handler class


joewallwork:
While working on integrating XIOS, we provided interfaces to a large subset of XIOS' overall functionality to cover everything that we thought we might need. As the work developed, it became apparent that some of the functions we ported either aren't required or are only ever set to one value. As such, it would make sense to go through the interface and:
* Drop any methods that are never used.
* Privatise any methods that shouldn't be customisable by the user.
* Inline any methods that are only used internally and only at one call site.

joewallwork:
This can be done at the end of the project to tidy up.

PR #851, opened 2025-05-19
XIOS: Inline name setters and remove name getters

joewallwork:
# XIOS: Inline name setters and remove name getters

Towards #850.

The XIOS handler class currently supports setting and getting names for the XIOS File, Field, and Grids. However, it is most convenient to just use the same name as the ID for those objects. (In fact, we already do this in some cases.)

To reduce the amount of code to be maintained in the XIOS interface, the `setXName` methods are called once directly in the `createX` methods and inlined there. The corresponding `getXName` methods are dropped. The docs are updated to mention this.

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -1228,24 +1228,6 @@ void Xios::createField(const std::string fieldId)
 }
 
 /*!
- * Set the name of a field with a given ID
- *
- * @param the field ID
- * @param name to set
- */
-void Xios::setFieldName(const std::string fieldId, const std::string fieldName)
-{
-    xios::CField* field = getField(fieldId);
-    if (cxios_is_defined_field_name(field)) {
-        Logged::warning("Xios: Overwriting name for field '" + fieldId + "'");
-    }
-    cxios_set_field_name(field, fieldName.c_str(), fieldName.length());
-    if (!cxios_is_defined_field_name(field)) {
-        throw std::runtime_error("Xios: Failed to set name for field '" + fieldId + "'");
-    }
-}
-
-/*!
  * Set the operation for a field with a given ID
  *
  * @param the field ID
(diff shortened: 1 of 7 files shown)


core/src/Xios.cpp
```diff
@@ -488,7 +495,8 @@ void Xios::createAxis(const std::string axisId)
         throw std::runtime_error("Xios: Failed to create axis '" + axisId + "'");
     }
     if (axisId == "z_axis") {
-        std::string domainId = "xy_domain";
+        // Create grid_3D associated with a domain called xy_domain and an axis called z_axis
+        const std::string domainId = "xy_domain";
         cxios_domain_valid_id(&exists, domainId.c_str(), domainId.length());
         if (exists) {
             createGrid("grid_3D");
```
TomMelt, line 502: Am I right in thinking this is for Z-fields? Do you want to keep 3D fields in case we add in future? or is it better to remove and save maintenance?

docs/xios.rst
```diff
@@ -59,7 +59,9 @@ handler is instantiated, the config sections above will be parsed. Based on the
 values that are parsed, the handler object will automatically create `Field` and
 `File` data structures for XIOS and associate these as appropriate. If the
 `XiosInput` section is used, an attempt will be made to open the file provided
-by the `filename` entry.
+by the `filename` entry. Note that the nextSIM-DG XIOS integration is set up
+such that the filename and field names coincide with the fileId and fieldId of
+the corresponding `File` and `Field` objects.
```
TomMelt: I think `.rst` requires two backticks (``) to mark inline code (see [here](https://bashtage.github.io/sphinx-material/rst-cheatsheet/rst-cheatsheet.html) for more info). Could you update this file to fix the inline code sections?

TomMelt (changes requested):
Thanks @jwallwork23 , looks great to me other than the minor change requested for the documentation. The Z-field stuff is up to you if you leave it in or not (assuming I understood correctly)

docs/xios.rst
joewallwork: Done in d6de1337.

core/src/Xios.cpp
joewallwork: If we are going to restrict attention to 1D and 2D fields then this would massively simplify the XIOS interface. I've asked about this on Slack. If we're going to go ahead with that then it's probably best if I address it in a different PR.

docs/xios.rst
TomMelt: thanks

TomMelt (approved):
LGTM :rocket:

PR #859, opened 2025-05-27
XIOS: Set up domain from `ModelMetadata`

joewallwork:
# XIOS: Set up domain from `ModelMetadata`

Towards #850
Merges into branch for #860

This PR significantly cuts down the XIOS domain interface.

* The various setters are inlined into `createDomain`, which is made private.
* All domain getters are dropped because they're no longer needed.
* Add `affixModelMetadata` method for the XIOS handler that associates an instance of `ModelMetadata`.
* The `XiosDomain_test.cpp` is dropped. Since we're calling the XIOS C header directly, we can rely on XIOS for the testing.


core/test/CMakeLists.txt
```diff
@@ -90,21 +90,13 @@ if(ENABLE_MPI)
         )
         target_link_libraries(testXiosAxis_MPI3 PRIVATE nextsimlib doctest::doctest)
 
-        add_executable(testXiosDomain_MPI4 "XiosDomain_test.cpp" "MainMPI.cpp")
-        target_compile_definitions(testXiosDomain_MPI4 PRIVATE USE_XIOS)
```
joewallwork, line 94: The reason I switched back to 2 MPI ranks is that the test now uses the test partition metadata file, which is set up for 2 ranks.

joewallwork:
Apologies, I decided to change bases. This PR will now merge into the branch for #860. Doing so meant I could do further inlining because `gridAddDomain` will only be needed in one place when support for 3D and 4D fields is dropped in that PR.

core/src/include/Xios.hpp
```diff
@@ -95,32 +96,12 @@ class Xios : public Configured<Xios> {
...
-    std::string getDomainType(const std::string domainId);
-    size_t getDomainGlobalXSize(const std::string domainId);
-    size_t getDomainGlobalYSize(const std::string domainId);
-    size_t getDomainLocalXSize(const std::string domainId);
-    size_t getDomainLocalYSize(const std::string domainId);
-    size_t getDomainLocalXStart(const std::string domainId);
-    size_t getDomainLocalYStart(const std::string domainId);
-    std::vector<double> getDomainLocalXValues(const std::string domainId);
-    std::vector<double> getDomainLocalYValues(const std::string domainId);
+    void affixModelMetadata(ModelMetadata& metadata);
```
TomMelt, line 97: not for this PR but after my PR (#744 ) is merged we can replace this argument with the `ModelMetadata` singleton instance

TomMelt (approved):
Thanks @jwallwork23 , all looks good to me. Only had 1 comment but it's more of a reminder for me anyway.

joewallwork:
[Rebased on top of `issue766_xios-dumpmodelstate-direct`.]

PR #867, opened 2025-06-04
Move updateWrapper from IceGrowth

### LINKED ISSUE #866: The lateral ice spread module should have a single "update" entry point
Most modules have a single entry point via a function ``update(const TimestepTime& tsTime)``. They then access variables through the ModelArrayRef system. The LateralIceSpreadModule, on the other hand, has two entry functions (``freeze`` and ``melt``) with multiple arguments. It would be more consistent if this module had a single ``update`` function, like all the other modules.

einola:
# Move updateWrapper from IceGrowth

Fixes #866 

### Task List

---
# Change Description

I replaced the overElements call to updateWrapper in IceGrowth with a call to iLateral->update(tst). The idea is to separate out the logic of lateral ice growth and melt out of IceGrowth itself and into ILateralIceSpread modules. This does require ILateralIceSpread.hpp to define and share three variables; for change in ice concentration, ice thickness, and snow thickness.

---
# Test Description

All CI tests pass.

---
# Documentation Impact

N/A

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/physics/src/include/IceGrowth.hpp
+++ b/physics/src/include/IceGrowth.hpp
@@ -1,7 +1,7 @@
 /*!
  * @file IceGrowth.hpp
  *
- * @date Jul 5, 2022
+ * @date 04 Jun 2025
  * @author Tim Spain <timothy.spain@nersc.no>
  * @author Einar Ólason <einar.olason@nersc.no>
  */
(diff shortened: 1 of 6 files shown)


physics/src/modules/include/ILateralIceSpread.hpp
```diff
@@ -20,53 +20,54 @@ class ILateralIceSpread : public ModelComponent {
...
+    {
+        deltaCIce.resize();
+        newice.resize();
+        snowMelt.resize();
+
+        /* Set these to zero, so we don't have uninitialized values floating around.
+         * Should this be done here, or in derived classes? Arguably only in the derived classes,
+         * since having uninitialised values helps with error detection. But as long as we have the
+         * nextsim_thermo.use_thermo_forcing option, all derived classes must initialise these
+         * values to something sensible.*/
```
timspainNERSC: I'd limit this comment to what the code is doing and why. And leave out the uncertainty.
```
         /*
           * Set these to zero, so we don't have uninitialized values floating around.
           * As long as we have the nextsim_thermo.use_thermo_forcing option, all
           * derived classes must initialise these values to something sensible.
           */

physics/src/include/IceGrowth.hpp
timspainNERSC, line 1: I believe the functions

- `newIceFormation`
- `lateralIceSpread`
- `applyLimits`
- `updateWrapper`

are now redundant and can be deleted from `IceGrowth.hpp`.

timspainNERSC (changes requested):
- A suggested change the the wording of a comment.
- Some more lines of code that I am fairly sure can be deleted.

physics/src/modules/include/ILateralIceSpread.hpp
einola: Yes, good point.

physics/src/include/IceGrowth.hpp
einola: Good catch!

timspainNERSC (approved):

Issue #780, opened 2025-01-20
DG components in restarts for all advected fields


einola:
The restart file must include all DG components for all fields advected by the model.

timspainNERSC:
Will all fields that are advected by the model be already included in the restart file? Would it be correct to say that this will not (currently) add any new fields to the restart file? Instead all advected fields are required to be written and read with all DG components.

PR #869, opened 2025-06-06
Automate doxygen frontmatter

joewallwork:
*Don't worry! The vast majority of the lines changed are very minor!*

Since we introduced the pre-commit hook to automatically update date stamps in #660, the date stamps have been updated much more regularly, which is presumably a Good Thing.

However, these trivial changes are becoming real obstacles when rebasing, especially when rebasing over many commits. On several occasions, both @TomMelt and I have found that the vast majority of conflicts that arise are due to differences in when the date stamps were updated. As a result, we feel these frequent date updates are now hampering productivity.

I think we need to reassess how we handle the date stamps. It's not an option to drop them altogether because we require dates and various other pieces of frontmatter in the files to support the Doxygen automatic documentation setup.

In this PR, I propose leaving both the `@date` entry and also the `@file` entry out of all C++ files and provide an automated approach for adding them in when needed. I provide the `scripts/doxygen_prep.sh` script to support these updates, which determines the filename and most recent modification date using `git blame`.

You can run
```sh
for EXTENSION in cpp hpp ipp; do
  find . -name "*.${EXTENSION}" -exec scripts/doxygen_prep.sh {} \;
done
```
from the command line to test it out. Having done so, executing `git diff` should show various changes like
```diff
diff --git a/core/src/CommandLineParser.cpp b/core/src/CommandLineParser.cpp
index ca05fdd9..fa203a8f 100644
--- a/core/src/CommandLineParser.cpp
+++ b/core/src/CommandLineParser.cpp
@@ -1,4 +1,7 @@
 /*!
+ * @file    CommandLineParser.cpp
+ *
+ * @date    05 Jun 2025
  * @author  Tim Spain <timothy.spain@nersc.no>
  * @author  Tim Williams <timothy.williams@nersc.no>
  */
```

Other related changes in this PR:
* Update the Doxygen GitHub Action so that it inserts the `@file` and `@date` information, as appropriate.
* Formatting of frontmatter across all `.cpp`, `.hpp`, and `.ipp` files, for consistency.
* Add `@author` entries to files that were missing them (with the authors chosen determined from who seemed to have contributed most to the `git blame`).
* I propose/clarify in the `README` that `@author` refers to the main author(s) / code owner(s) for the particular file.
* Drop the date stamp update from the `.pre-commit`, as it's no longer relevant.

## Checklist

* [x] Check that the updated Doxgen workflow does indeed work. (It's not triggered by this PR.)

PR #877, opened 2025-06-17
Advect any field

### LINKED ISSUE #99: Advecting arbitrary fields

timspainNERSC:
# Advect any field

Fixes #99 #337 #780 

---
# Change Description

- Add a `FieldAdvection` class to provide access to field advection functionality for `ModelComponents` that require it.
- Add a test of advection that runs the full model on an idealized test case.
- If a single component `ModelArray` (for example, one of `Type::H`) is assigned to a multiple-component `ModelArray` (for example, one of `Type::DG`), then type is not changed and the new data is assigned to the 0 component of the target array.
- The interface class `IDynamics` and the implementations of this module now provide `prepareAdvection()` and `advectField()` functions.
- Moved the damage field to a `DGVectorHolder`, allowing the dynamics to directly use the data held in the `ModelArray` copy of the data.
- Add `advectField()` (for `ModelArray`s) and `advectDGField()` (for `DGVector`s).
  - Overload resolution didn't seem to work well at the compile stage when both functions were called `advectField` with different argument types.
- Add a minimum temperature limit for sea ice (set to -90˚C).

---
# Test Description

The test uses a simple generated restart file on a 50x21 grid with constant wind and current to advect the ice using the BBM dynamics. The testing code reads the final state of the model and ensures that the initial rectangular area of snow has moved towards the +x direction.


core/src/PrognosticData.cpp
```diff
@@ -111,6 +101,9 @@ void PrognosticData::update(const TimestepTime& tst)
 {
     pOcnBdy->updateBefore(tst);
     pAtmBdy->update(tst);
+    pDynamics->prepareAdvection();
+
+    FieldAdvection::advectField(hsnow, tst, 0.);
```
einola: Snow advection should take place in the dynamics routine, probably IDynamics. The reason is that the absolute minimum we can advect is mass and concentration, and the mass is ice and snow. Ergo, we must always advect concentration, ice thickness, and snow thickness, even though we've ignored snow thickness so far.

core/src/PrognosticData.cpp
```diff
@@ -17,21 +18,22 @@ namespace Nextsim {
 
 PrognosticData::PrognosticData()
     : m_dt(1)
-    , m_snow(ModelArray::Type::H)
-    , m_damage(ModelArray::Type::H)
     , hiceAdvection(ModelArray::AdvectionType)
     , ciceAdvection(ModelArray::AdvectionType)
+    , damage(ModelArray::AdvectionType)
+    , hsnow(ModelArray::AdvectionType)
```
einola, line 24: How about ``hsnowAdvection``, to be consistent with ``hiceAdvection`` and ``ciceAdvection``? We could also remove ``Advection`` from the variable name.

core/src/PrognosticData.cpp
```diff
@@ -139,23 +131,10 @@ void PrognosticData::updatePrognosticFields()
...
-    ModelArrayRef<Shared::DAMAGE, RO> damageUpd(getStore());
-
-    // Calculate the cell average thicknesses
-    HField hsnowUpd;
-    hsnowUpd.setData(hsnowTrueUpd.allComponents() * ciceAdvection.component(0));
-
-    m_snow.setData(hsnowUpd);
-    m_damage.setData(damageUpd);
-}
+void PrognosticData::updateDynamicsFields() { }
```
einola: Why not just remove it?

core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -173,6 +165,12 @@ void BBMDynamics::update(const TimestepTime& tst)
     sigmaII = kernel.getDG0Data(sigmaIIName);
 }
 
+void BBMDynamics::advectField(double timestep, ModelArray& field, double lowerLimit, double upperLimit)
```
einola, line 174: Can this be in IDynamics?

core/src/modules/DynamicsModule/include/BBMDynamics.hpp
```diff
@@ -24,8 +24,13 @@ class BBMDynamics : public IDynamics, public Configured<BBMDynamics> {
     BBMDynamics();
 
     std::string getName() const override { return "BBMDynamics"; }
+    void prepareAdvection() override { kernel.prepareAdvection(); }
     void update(const TimestepTime& tst) override;
 
+    void advectField(double timestep, ModelArray& field, double lowerLimit =
+            -std::numeric_limits<double>::infinity(), double upperLimit =
+            std::numeric_limits<double>::infinity()) override;
```
einola, line 32: If we can't move ``advectField`` and ``prepareAdvection`` into ``IDynamics`` (because ``kernel`` needs to be defined, perhaps), then we should write the one line of ``advectField`` out here, like you've done for ``prepareAdvection``.

core/src/modules/DynamicsModule/include/DummyDynamics.hpp
```diff
@@ -20,6 +20,7 @@ class DummyDynamics : public IDynamics {
 
     std::string getName() const override { return "DummyDynamics"; }
     void update(const TimestepTime& tst) override {};
+    void prepareAdvection() override { }
```
einola, line 23: Is that needed?

core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
```diff
@@ -31,6 +31,7 @@ class MEVPDynamics : public IDynamics, public Configured<MEVPDynamics> {
     MEVPDynamics();
 
     std::string getName() const override { return "MEVPDynamics"; }
+    void prepareAdvection() override { kernel.prepareAdvection(); }
```
einola: ``advectField`` implementation missing

core/src/modules/include/IDynamics.hpp
```diff
@@ -124,10 +122,10 @@ class IDynamics : public ModelComponent {
     HField sigmaI;
     HField sigmaII;
     // References to the DG0 finite volume data arrays
-    ModelArrayRef<Shared::H_ICE, RW> hice;
-    ModelArrayRef<Shared::C_ICE, RW> cice;
+//    ModelArrayRef<Shared::H_ICE, RW> hice;
+//    ModelArrayRef<Shared::C_ICE, RW> cice;
```
einola: Remove instead of commenting out.

core/test/FieldAdvection_test.cpp
```diff
@@ -0,0 +1,12 @@
+/*!
+ * @file FieldAdvection_test.cpp
+ *
+ * @date Jun 5, 2025
+ * @author Tim Spain <timothy.spain@nersc.no>
+ */
+
```
einola: This doesn't look like it's testing much. Should there be a specific ``FieldAdvection_test.cpp`` or can we delete this file?

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -181,6 +171,52 @@ template <int DGadvection, int DGstress> class DynamicsKernel {
...
+     * @return A reference to the advected array.
+     */
+    virtual DGVector<DGadvection>& advectDGVField(double timestep, DGVector<DGadvection>& field,
+        double lowerLimit = -std::numeric_limits<double>::infinity(),
+        double upperLimit = std::numeric_limits<double>::infinity())
+        = 0;
+
+    virtual void advectDynamicsFields(double timestep)
+    {
+        advectDGVField(timestep, hice, 0.0);
```
einola, line 189:
```suggestion
        advectDGVField(timestep, hice, 0.0);
        advectDGVField(timestep, hsnow, 0.0);
```

is the way to go. We also need to use the total mass of the ice and snow in the momentum equation, but that's for another PR

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -157,20 +153,14 @@ template <int DGadvection, int DGstress> class DynamicsKernel {
             throw std::runtime_error(
                 std::string("DynamicsKernel::getDG0Data: Use array sharing for ") + name);
         } else {
-            // Use the stored array type to ensure the returned data has the correct type
-            ModelArray::Type type = fieldType.at(name);
-            ModelArray data(type);
-            data.resize();
-            return DGModelArray::dg2ma(advectedFields.at(name), data);
+            return ModelArray(ModelArray::AdvectionType);
```
einola: So, this function now either throws an error or returns an empty ModelArray; did I understand that correctly? Then why don't we remove it?

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -157,20 +153,14 @@ template <int DGadvection, int DGstress> class DynamicsKernel {
...
         }
     }
 
     virtual void update(const TimestepTime& tst) { ++stepNumber; }
 
     void advectionAndLimits(const TimestepTime& tst)
```
einola: Don't we want to stop using this and remove it?

dynamics/src/include/FreeDriftDynamicsKernel.hpp
```diff
@@ -57,12 +62,13 @@ template <int DGadvection> class FreeDriftDynamicsKernel : public CGDynamicsKern
 
     void update(const TimestepTime& tst) override
     {
+        // Advect the ice here to match advection in the rest of the model.
+        advectDGVField(tst.step.seconds(), hice, 0.0);
+        advectDGVField(tst.step.seconds(), cice, 0.0, 1.0);
```
einola: What about calling ``advectDynamicsFields``?

test/ThermoIntegration_test.py
```diff
@@ -201,9 +201,9 @@ def test_iceThickness(self):
         max = 3.3419
         min = 2.9805
         hiceDG0 = self.hice[:, 0]
-        self.assertAlmostEqual(max, hiceDG0.max(), 4, "Max ice thickness not ~= " + str(max) + " m")
-        self.assertAlmostEqual(min, hiceDG0.min(), 4, "Min ice thickness not ~= " + str(min) + " m")
-        self.assertAlmostEqual(mean, hiceDG0.mean(), 4, "Mean ice thickness not ~= " + str(mean) + " m")
+        self.assertAlmostEqual(max, hiceDG0.max(), 4, "Max ice thickness not ~= " + str(max) + " ˚C")
+        self.assertAlmostEqual(min, hiceDG0.min(), 4, "Min ice thickness not ~= " + str(min) + " ˚C")
+        self.assertAlmostEqual(mean, hiceDG0.mean(), 4, "Mean ice thickness not ~= " + str(mean) + " ˚C")
```
einola: Overcorrection! :D

```suggestion
        self.assertAlmostEqual(max, hiceDG0.max(), 4, "Max ice thickness not ~= " + str(max) + " m")
        self.assertAlmostEqual(min, hiceDG0.min(), 4, "Min ice thickness not ~= " + str(min) + " m")
        self.assertAlmostEqual(mean, hiceDG0.mean(), 4, "Mean ice thickness not ~= " + str(mean) + " m")
```

einola (changes requested):
Nice! Small things mostly, but it's a big PR, so it's sometimes difficult to have an overview of it. One thing I'm sure I want to change is the fact that snow should be advected in the same place as the concentration and thickness (as mentioned in the inline comments). I also think that damage should belong to BBMDynamics and not PrognosticData. That would simplify things, but it's a question of whether we want to do that in a different PR.

core/src/ModelArray.cpp
```diff
@@ -37,8 +37,12 @@ ModelArray::ModelArray(const ModelArray& orig)
 
 ModelArray& ModelArray::operator=(const ModelArray& orig)
 {
-    type = orig.type;
-    setData(orig.m_data);
+    if (orig.nComponents() == 1 && nComponents() != 1) {
+        component(0) = orig.data();
```
einola, line 41:
```suggestion
        component(0) = orig.data();
        for ( int n=1; n< nComponents(); n++) {
            component(n).setZero();
```

Like this, assigning the mean component resets the higher-order components. The idea is to avoid ending up with uninitialised values when assigning.

core/src/PrognosticData.cpp
timspainNERSC: Done

core/src/PrognosticData.cpp
timspainNERSC: Done

core/src/PrognosticData.cpp
timspainNERSC: Indeed, no longer needed. Done

core/src/modules/DynamicsModule/BBMDynamics.cpp
timspainNERSC: No, as it needs to communicate with the kernel object. The kernel classes are not related to one another and functions that calls them directly need to be in the individual implementation classes.

core/src/modules/DynamicsModule/include/BBMDynamics.hpp
timspainNERSC: I disagree, and say that we should move the one line of `prepareAdvection` into the source file like `advectField`. But I do agree that they should be consistent.

core/src/modules/DynamicsModule/include/DummyDynamics.hpp
timspainNERSC: Yes, `IDynamics` declares this as a pure virtual function.

core/src/modules/DynamicsModule/include/MEVPDynamics.hpp
timspainNERSC: Implemented.

core/src/modules/include/IDynamics.hpp
timspainNERSC: Done

core/test/FieldAdvection_test.cpp
timspainNERSC: Removed.

dynamics/src/include/DynamicsKernel.hpp
timspainNERSC: Done.

dynamics/src/include/DynamicsKernel.hpp
timspainNERSC: Why indeed. Removed.

dynamics/src/include/DynamicsKernel.hpp
timspainNERSC: Yes we do. Removed.

dynamics/src/include/FreeDriftDynamicsKernel.hpp
timspainNERSC: Much better, yes. Done.

test/ThermoIntegration_test.py
timspainNERSC: Recorrected.

core/src/ModelArray.cpp
timspainNERSC: The use case for this operator is copying back the updated DG0 values after the column physics has run. If the other components were set to 0, as in the suggested change, then the higher DG components would be set to 0. And there is no other copy of the array that the dynamics uses any more.

.github/workflows/test_suite.yml
```diff
@@ -179,3 +180,4 @@ jobs:
         cd test
         python ThermoIntegration_test.py
         ./run_test_jan2010_integration_test.sh python
+        ./run_test_advection_test.sh python
```
joewallwork: Suggest replacing `python` with `python3` here, for consistency with the above.

core/test/FieldAdvection_test.cpp
joewallwork: The test file still seems to be there?

test/run_advection_test.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
+if [ ! -f ../nextsim ]; then
+    echo "Copy or link the nextsim executableinto the parent directory from the build directory"
```
joewallwork:
```suggestion
    echo "Copy or link the nextsim executable into the parent directory from the build directory"
```

test/test_advection_test.py
joewallwork, line 1: Could this not just be renamed to `test_advection.py`?

test/test_advection_test.py
```diff
@@ -0,0 +1,60 @@
...
+if not data_group_name in root.groups:
+    print(f"Error: no group {data_group_name} found in file {file_name}.")
+    sys.exit(2)
+
+data_group = root[data_group_name]
+
+# Check that an hsnow variable exists
+hsnow_name = "hsnow"
+if not hsnow_name in data_group.variables:
+    print(f"Error: no varaible {hsnow_name} found in netCDF path {file_name}/{data_group_name}.")
```
joewallwork:
```suggestion
    print(f"Error: no variable {hsnow_name} found in netCDF path {file_name}/{data_group_name}.")
```

test/run_test_advection_test.sh
joewallwork, line 1: Similarly, couldn't this just be called `run_test_advection.sh` or `run_advection_test`?

joewallwork (changes requested):
This mostly looks good to me - I just have a few fairly minor suggestions and requests.

physics/src/modules/IceThermodynamicsModule/ThermoWinton.cpp
```diff
@@ -162,6 +164,11 @@ void ThermoWinton::setData(const ModelState::DataMap& state)
 
 void ThermoWinton::update(const TimestepTime& tst)
 {
+    // Advect ice temperatures
+    IIceThermodynamics::update(tst);
+    FieldAdvection::advectField(tInternal, tst, IIceThermodynamics::minT, 0.);
+    FieldAdvection::advectField(tBottom, tst, IIceThermodynamics::minT, 0.);
```
einola:
```suggestion
    FieldAdvection::advectField(tInternal, tst, IIceThermodynamics::minT, seaIceTf);
    FieldAdvection::advectField(tBottom, tst, IIceThermodynamics::minT, seaIceTf);
```

I belatedly realised that you need to cap the internal temperatures at ``seaIceTf`` and not 0. This is because:

 - Physics: We don't want advection to cause internal melting of the ice
 - Numerics: There's a division with ``tInternal`` early on in the equations, so a zero value will generate a NaN.

.github/workflows/test_suite.yml
timspainNERSC: ~Done~

test/run_advection_test.sh
timspainNERSC: Done

test/test_advection_test.py
timspainNERSC: Yes, it could. Done

test/test_advection_test.py
timspainNERSC: Done

test/run_test_advection_test.sh
timspainNERSC: Done (also for the integration test).

physics/src/modules/IceThermodynamicsModule/ThermoWinton.cpp
timspainNERSC: Done.

.github/workflows/test_suite.yml
timspainNERSC: Undone. Whatever is the `python3`interpreter in the Mac image does not have access to the `netCDF4` and `numpy` packages.

core/src/modules/DynamicsModule/include/BBMDynamics.hpp
einola: Ok, that's fair enough. I'm happy as long as they are consistent.

core/src/ModelArray.cpp
einola: Yes, that's a good point. I was only thinking about what happens when the initial conditions are set. But in this case, we can probably trust that the ``resize`` operation has set everything to zero (as it currently does).

So, ignore this suggestion.

einola (approved):

joewallwork (approved):
All looks good to me! Once this is merged, I'll fix conflicts and merge #869.

PR #879, opened 2025-06-23
Issue 878 CG ice velocity components

### LINKED ISSUE #878: Ingesting and outputting CG velocities
For bit-exact restarts, the ice velocity components should be stored in the initial/restart files on the CG grid.

Required steps:

- [x] ~The `CGDynamicsKernel` class provides a `setCGArray` function to support passing CG data directly to the appropriate dynamics kernel~ `setData` will accept all `ModelArray::Type`s.
- [ ] Add more intelligence to `ma2cg` to deal with `H`, `DG` and `CG` types, rather than assuming everything is H.
- [ ] Dynamics implementations (`BBMDynamics` & `MEVPDynamics`) should return the CG arrays as part of the return data for `getDataPrognostic()`.

The reading and writing of CG arrays by `ParaGridIO` should already be supported, but this needs to be checked.

timspainNERSC:
# CG ice velocity components

Fixes #878

---
# Change Description

When passing a `ModelArray` to the `ma2cg()` function, the operations that are performed depends on the `ModelArray::Type`. An `HField` is converted to a `DGVector` and then interpolated to a `CGVector`. A `DGField` will only be interpolated to a `CGVector`. A `CGField` will be directly copied into the `CGVector`.

Adds a `getCGData()` function to `CGDynamicsKernel` which will return the full CG array of the ice velocity components. These are then passed to the data maps collected as part of the `getDataPrognostic()` call tree.

Set the size of the `ModelArray::Dimension`s `XCG`and `YCG` to values derived from the `CGdegree` value and the size of a DG0 (`Type::H`) `ModelArray`.

Since the ice velocities are so completely removed from the thermodynamics/physics part of the model (that is where `ModelArray` is used), remove the special `ModelArray` types for U and V.

---
# Test Description

1. Run the model from `init_25km_NH.nc` for 1 TS. Let the resulting restart file be 'file 1'.
2. Run the model from `init_25km_NH.nc` for 1 day. Let the resulting restart file be 'file 2'.
3. Run the model from 'file 2' for 1 TS. Let the resulting restart file be 'file 3'.
4. Compare the values of _u_ and _v_ in 'file 3' with the values in 'file 1' and 'file 2'. Note the velocity in 'file 3' much more closely matches that of 'file 2' than that of 'file 1' which is only one timestep from being initialized as all zero.


TomMelt:
I had a look at the failing test and it looks like one of the MPI tests gets stuck in an infinite loop. I ran it locally and I get the same behaviour.

I will see if I can find out why

core/src/ParaGridIO.cpp
```diff
@@ -272,7 +272,7 @@ void ParaGridIO::dumpModelState(
     for (auto entry : ModelArray::definedDimensions) {
         ModelArray::Dimension dim = entry.first;
         size_t dimSz = (dimCompMap.count(dim)) ? ModelArray::nComponents(dimCompMap.at(dim))
-                                               : dimSz = entry.second.globalLength;
+                                               : dimSz = entry.second.localLength;
```
TomMelt, line 321: found it :sweat: 

`localLength` should be `globalLength`. This part sets the total dimension of the array ready to be written to file. Only the portion of ModelArray will actually be written, but netcdf parallel needs to know the total size so it can write to the correct portion.

```suggestion
                                               : dimSz = entry.second.globalLength;
```

TomMelt: I tested locally and this fixes the issue with the test

timspainNERSC: I made this change because a 1 day example run was failing with the following error message. Making the change to `localLength` let the file writing finish correctly.
```
2010-01-01T23:45:00Z
libc++abi: terminating due to uncaught exception of type netCDF::exceptions::NcEdge: NetCDF: Start+count exceeds dimension bound
file: /tmp/netcdf-cxx-20250303-65682-ahp603/netcdf-cxx4-4.3.1/cxx4/ncVar.cpp  line:1049
Abort trap: 6
```

dynamics/src/CGDynamicsKernel.cpp
```diff
@@ -85,22 +85,22 @@ template <int DGadvection>
...
         return DGModelArray::dg2ma(vtmp, data);
     } else if (name == uIOStressName) {
-        ModelArray data(ModelArray::Type::U);
+        ModelArray data(ModelArray::Type::H);
         DGVector<DGadvection> utmp(*smesh);
         Nextsim::Interpolations::CG2DG(*smesh, utmp, uIceOceanStress);
         return DGModelArray::dg2ma(utmp, data);
     } else if (name == vIOStressName) {
-        ModelArray data(ModelArray::Type::V);
+        ModelArray data(ModelArray::Type::H);
```
joewallwork, line 118: You could now hoist the declaration out of the conditional block to avoid duplication.

core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -103,6 +103,27 @@ void BBMDynamics::setData(const ModelState::DataMap& ms)
     uice = ms.at(uName);
     vice = ms.at(vName);
 
+    // Handle CG data field sizes
+    const ModelArray::MultiDim cgDims = kernel.getCGDimensions();
+    /* If the u is Type::CG and the dimensions read from the restart file do
+     * not match those of the CG dynamics kernel, throw an exception.
+     * TODO: Support interpolation between different CG number of Gauss points.
```
joewallwork, line 110: Is this TODO for this PR or for a future one?

joewallwork (approved):
This looks good to me. I just have a query about TODO notes and a minor suggested refactoring.

TomMelt:
@timspainNERSC I have had a more detailed look and I think the underlying problem is that the `x_cg` and `y_cg` dimensions are wrong in the initial model file. For example, when I run the `config_june23.cfg` it uses input file `init_25km_NH.nc`. In there the dimensions are listed as follows:

```
netcdf init_25km_NH {
dimensions:
	ydim = 121 ;
	xdim = 154 ;
	yvertex = 122 ;
	xvertex = 155 ;
	y_cg = 122 ;
	x_cg = 155 ;
	dg_comp = 1 ;
	dgstress_comp = 3 ;
	ncoords = 2 ;
```

They should be `y_cg = 243` and `x_cg = 309`. If this the intention is not to hardcode these to a specific value because we dont want to make separate input files for each `CGDegree`, then we should just remove these completely and they should be set at run time using `(CGDegree * Dim + 1)`.

These get read in here in `ParaGridIO::getModelState` and then set on line 137.

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/ParaGridIO.cpp#L81-L137

The `localLength` is then overridden later in `BBMDynamics.cpp` on line 125:

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/modules/DynamicsModule/BBMDynamics.cpp#L125

**Note** the function `ModelArray::setDimensions` only sets the `localLength` and not the `globalLength` and that's why you needed to make the changes here:

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/ParaGridIO.cpp#L266-L267

I think this logic needs to be re-written to use the `ModelMetadata` singleton so that it will work for MPI and serial code. Then we need to also modify function `ModelArray::setDimensions` or create a new one that will also set the `globalLength`. 

Let me know what you think. I can try and implement a fix if you want.

einola:
@timspainNERSC What's the status of this?

PR #882, opened 2025-07-14
Expand and restructure gitignore

joewallwork:
I recently built the nextSIM-DG docs and found that the current `.gitignore` does not ignore the various XML, HTML, etc. files that are generated, so I thought it'd be good to add these.

While I was at it, I restructured and sorted the gitignore more generally. I dropped `run/*.cdl` because this seems to me like something we do want to track.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,22 +1,32 @@
-/build*
+# .gitignore
+
+# Directories
 */build
-Results
+/build*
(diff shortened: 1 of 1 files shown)


timspainNERSC (approved):
These look like reasonable files to ignore and a good structure to the file.

PR #885, opened 2025-07-15
TOPAZ4 interpolation

### LINKED ISSUE #880: Interpolate and extrapolate TOPAZ fields more agressively
At the moment, make_initNH.py and era5_topaz4_maker.py will only interpolate TOPAZ4 data onto the model grid if there is no land there in the original TOPAZ4 grid. This has the advantage that we never do any extrapolation, but it has two big disadvantages:

 - We can't run the model with more highly resolved coast line
 - The TOPAZ4 mask isn't consistent over time(!) so points that were dry may become wet, later in the time series and vice versa

In the Lagrangian neXtSIM, we used a triangulation of the TOPAZ4 data to interpolate into deep fjords and the like. We should do something similar here, to have the same functionality.

einola:
# TOPAZ4 interpolation

Fixes #880 

### Task List

---
# Change Description

This PR (unfortunately) does two things:

 1. Address issue #880 by using scipy.interpolate to fill in areas that were land in the TOPAZ4 grid, but may be ocean in a more highly resolved mesh.
 2. Uses the new batch of TOPAZ4 files, which Tim W. has prepared for neXtSIM Lagrangian. I did this because the scipy interpolation routine was much less forgiving of errors that using the old files would introduce. I don't know why it works better with the new set of files, but since we're using them for neXtSIM Lagrangian now, we might as well switch here, too! I'm currently uploading the new files to the SUMMER storage so everyone in SASIP can access them.

---
# Test Description

The June '23 setup runs. I haven't updated the integration test, because Joe's reworking the python scripts needed for that into cdl files.

---
# Documentation Impact

None, but we should document better how to run the June '23 setup.

---
# Other Details

None

---
### Pre-Request Checklist


run/interpolaters.py
joewallwork, line 1: It's usually spelt 'interpolators'.

run/interpolaters.py
```diff
@@ -0,0 +1,75 @@
+import math
+import numpy as np
+from scipy import interpolate
+import pyproj
+
+# Returns TOPAZ data interpolated from the data grid and coordinates to the target grid and coordinates
+def topaz4_interpolate(target_lon, target_lat, data, data_x, data_y, proj_string):
```
joewallwork: Suggest making the preceding comment into a docstring. Similarly for the other functions
```suggestion
def topaz4_interpolate(target_lon, target_lat, data, data_x, data_y, proj_string):
    """
    Returns TOPAZ data interpolated from the data grid and coordinates to the target grid and coordinates
    """
```

run/era5_topaz4_maker.py
```diff
@@ -56,115 +57,9 @@ def era5_source_file_name(field, unix_time, path):
     return f"{path}/ERA5_{field}_y{file_year}.nc"
 
 # Returns the file name that holds the TOPAZ data for a given field at a given time
-def topaz4_source_file_name(field, unix_time, path):
+def topaz4_source_file_name(unix_time, path):
     unix_tm = time.gmtime(unix_time)
```
joewallwork: It'd be useful to have a docstring that says what the two variables mean. For example, it's not totally clear what the path should be because another subdirectory gets joined in the function.

run/interpolaters.py
```diff
@@ -0,0 +1,75 @@
...
+    pole_lat = math.radians(75.0)
+    pole_lon = 320.0
+
+    delta_lon = np.radians(pole_lon - lon)
+
+    rlat = np.radians(lat)
+
+    # The rotation of the basis vector can be computed as the azimuth of the
+    # great circle path from the location to the location of the new pole.
+    return np.arctan2(math.cos(pole_lat) * np.sin(delta_lon), np.cos(rlat) * math.sin(pole_lat) - np.sin(rlat) * math.cos(pole_lat) * np.cos(delta_lon))
```
joewallwork: We could do with applying auto-formatting to Python files, as well as C++. Probably best to do after #889 is merged, though. (Not for this PR.)

run/interpolaters.py
```diff
@@ -0,0 +1,75 @@
...
+    points = np.array([target_x.ravel()[nanmask], target_y.ravel()[nanmask]]).T
+    xi = np.array([target_x.ravel(), target_y.ravel()]).T
+
+    return (interpolate.griddata(points, field.ravel()[nanmask], xi, method='nearest').reshape(target_lon.shape))
+
+# Returns bilinearly interpolated data given array of fractional indices
+# 2023-03-28 Add a wrap-around for the ERA longitude. This is formally
+# incorrect when this function is used for TOPAZ data, but since the target
+# point would have to be out of bounds of the source, it is not so important.
+def bilinear(eyes, jays, data):
```
joewallwork: I haven't seen the `eyes` and `jays` notations before, I like them! However, I feel these argument names do need to be explained in the docstring, otherwise users could get quite confused.

joewallwork (changes requested):
I have a few requests, mainly related to adding docstrings for the functions that make it easier to understand what they are intended for and what arguments they accept. Also, I'd recommend using `interpolators` as the module name.

einola:
Thanks for the review @jwallwork23. Good points about the docstrings, and I've added them to all the functions in ``interpolators.py``. I wasn't too sure about which formatting to go for, but I guess this should be discussed in #891, rather than here.

Thanduriel:
I wanted to run the updated script but I could not find the new TOPAZ4 files. The link in the [readme](https://github.com/sasip-climate/catalog-shared-data-SASIP/blob/main/forcings/TOPAZ4-1d-Arctic.md) gives a 404. There is no TOPAZ directory in the [catalog](https://ige-meom-opendap.univ-grenoble-alpes.fr/thredds/catalog/meomopendap/extract/SASIP/model-forcings/ocean_forcing/catalog.html) besides the old one. Is the upload not completed yet?

einola:
There was a permissions problem with the data I uploaded. It's fixed now, and you should be able to download what you need @Thanduriel

run/era5_topaz4_maker.py
joewallwork: Thanks for adding docstrings for the other functions - they're really useful. Please could you add something here, too, to make it clearer what the arguments mean?

joewallwork (changes requested):
> The June '23 setup runs. I haven't updated the integration test, because Joe's reworking the python scripts needed for that into cdl files.

Given that it's no longer clear that moving to CDL files is the best option (and is unlikely to be done soon in either case), could you update the integration test in this PR?

einola:
> > The June '23 setup runs. I haven't updated the integration test, because Joe's reworking the python scripts needed for that into cdl files.
> 
> Given that it's no longer clear that moving to CDL files is the best option (and is unlikely to be done soon in either case), could you update the integration test in this PR?

Well, I can ... but I end up with 80k deletions and 80k additions. I don't know how @timspainNERSC made these files, but I wrote a shell script with ``sed``, ``ncdump``, and ``ncks`` doing the heavy lifting. In the end, I get slightly different values, resulting in a huge amount of deletions and additions. So, I don't think it's worth it. Strictly speaking, it doesn't need updating, anyway.

joewallwork (approved):
> Well, I can ... but I end up with 80k deletions and 80k additions. I don't know how @timspainNERSC made these files, but I wrote a shell script with `sed`, `ncdump`, and `ncks` doing the heavy lifting. In the end, I get slightly different values, resulting in a huge amount of deletions and additions. So, I don't think it's worth it. Strictly speaking, it doesn't need updating, anyway.

Fair enough! Indeed, it's probably not worth it then.

Thanks for the further docstring additions. I think this is good to go.

PR #890, opened 2025-07-17
Issue881 init temp

### LINKED ISSUE #881: Initialisation of temperatures for Winton model
The Winton model crashes if T1 (``tinterior``) is zero at the start of the integration, because of a division by zero. However, at the moment, the default initial value in ``make_init_base.py`` is zero. A related problem is that if there is no ``tinterior`` field in the restart file, the model copies the surface value to those points. This could also cause ``tinterior`` to be initialised as zero and cause a crash.

I also don't think it's good for the model to start with T1 and T2 that are larger than the freezing point of ice, but this is more difficult to demonstrate. It is at least not physically reasonable, although the model may recover.

einola:
# Initialisation of ice temperatures

Fixes #881

### Task List

---
# Change Description

This PR simplifies the initialisation of temperature; we now only care about two cases:

 1. If we find a temperature field in the restart file, we use it.
 2. If we don't find a temperature field in the restart file:
    1. If it's the surface, we set it to 0°C
    2. If it's T1 or T2 of Winton's scheme, we set it to the freezing point of ice.

Reproducing neXtSIM Lagrangian behaviour, we don't need or expect the user to supply initial temperatures. I've also changed the init scripts and base class, so that temperature is not initialised in those. Very advanced users can still do it, but I can't imagine there will ever be a need for it.

---
# Test Description

All unit tests and integration tests work. Also tested with the ``make_initNH.py`` script in a realistic run.

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist


einola:
Still need to test it a bit better.

einola:
Sufficiently tested and ready for review @timspainNERSC, @jwallwork23, @TomMelt

physics/src/modules/IceThermodynamicsModule/ThermoWinton.cpp
```diff
@@ -111,51 +110,22 @@ void ThermoWinton::setData(const ModelState::DataMap& state)
...
-
-    if (state.count(tInteriorName) > 0) {
+    /* If the internal temperature is not in the restart file, then we simply set it to the freezing
+     * point of seawater. It's a safe approximation, and it seems the user doesn't really care! */
+    try {
         tInternal = state.at(tInteriorName);
-        setInterior = true;
+    } catch (const std::out_of_range& e) {
+        Logged::info("No " + tInteriorName
+            + " field in restart file. Setting it to the freezing point of ice.\n");
```
joewallwork: A couple of comments that might be naive due to my lack of knowledge about sea ice dynamics.

1. Out of interest, what's the difference between the freezing points of ice (here) and sea water (below). Is sea ice formed of fresh (or less saline) water, giving it a different freezing point?
2. The term "freezing point of ice" confuses me because ice is already a solid and I understood that the freezing point was the temperature at which a liquid transitions to being a solid.

physics/src/modules/include/IIceThermodynamics.hpp
```diff
@@ -27,15 +27,18 @@ class IIceThermodynamics : public ModelComponent {
...
         deltaHi.resize();
         snowToIce.resize();
+
+        /* If the surface temperature is not in the restart file, then we simply set it to the
+         * zero. It's a safe approximation, and it seems the user doesn't
+         * really care! */
+        try {
+            tsurf = ms.at(tsurfName);
+        } catch (const std::out_of_range& e) {
+            Logged::info("No " + tsurfName + " field in restart file. Setting it to 0°C.\n");
```
joewallwork, line 37: The warning message is different again here. Could they be made consistent?

joewallwork (changes requested):
This mostly looks good to me but I have a couple of (possibly naive) questions about the warning messages.

physics/src/modules/IceThermodynamicsModule/ThermoWinton.cpp
einola: I guess you're right that I should use "melting point" instead of "freezing point" in this case, so I'll change that.

So, sea ice forms when salty water freezes, at something like -1.8°C (depending on the ocean salinity). But there isn't room for all the ocean salt in the crystalline structure of the ice, so that gets ejected into the ocean when the ice forms. The end result is that the ocean has a salinity of c.a. 34 PSU (roughly parts per million), while newly formed ice has a salinity of about 12 PSU. This then drops down to about 5 PSU as the ice ages. This drop in salinity raises the melting point to almost, but not quite, 0°C

The reason I use the melting point as a reference and not 0°C is stated in the PR itself.

physics/src/modules/include/IIceThermodynamics.hpp
einola: Well, I don't have access to the melting point of sea ice here. And strictly speaking, I don't need it either. 0°C is an ok initial condition for the ice surface temperature.

joewallwork (approved):
Thanks for the explanation and edits, this is clearer now.

PR #896, opened 2025-07-28
Don't heal damage where there's no ice

### LINKED ISSUE #895: Minor bug in ConstantHealing
The code in constant healing has a division by concentration, but no escape if the concentration is zero. So, we end up with NaNs in the damage field where there's no ice. This trips up the code once ice grows or is advected into grid cells that previously had no ice in them.

The solution is to put an ``if ( cice[i] == 0.) continue;`` statement at the top.

einola:
# Don't heal damage where there's no ice

Fixes #895 

### Task List

---
# Change Description

I put an ``if ( cice[i] == 0.) continue;`` statement at the top of ConstantHealing::updateElement to skip ice-free points. This makes sense, but also averts inserting NaNs into the field, as the code does a division by ``cice[i]`` at some point.

---
# Test Description

All tests run.

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/physics/src/modules/DamageHealingModule/ConstantHealing.cpp
+++ b/physics/src/modules/DamageHealingModule/ConstantHealing.cpp
@@ -52,8 +52,10 @@ void ConstantHealing::update(const TimestepTime& tstep)
 void ConstantHealing::updateElement(size_t i, const TimestepTime& tstep)
 {
     // No ice, no healing
-    if (cice[i] == 0.)
+    if (cice[i] == 0.) {
+        damage[i] = oldDamage[i];
         return;
+    }
 
     // Only lateral growth contributes to healing, not melt(!)
     double const lateralGrowth = std::max(0., deltaCi[i]);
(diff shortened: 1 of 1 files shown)


physics/src/modules/DamageHealingModule/ConstantHealing.cpp
```diff
@@ -51,6 +51,12 @@ void ConstantHealing::update(const TimestepTime& tstep)
 
 void ConstantHealing::updateElement(size_t i, const TimestepTime& tstep)
 {
+    // No ice, no healing
+    if (cice[i] == 0.) {
```
joewallwork: Might it be worth doing an approximate check? e.g.,
```suggestion
    double tol = 1e-10;
    if (abs(cice[i]) < tol) {
```
Otherwise unexpected behaviour might result when `cice` is effectively (but not quite) zero.

joewallwork (changes requested):

physics/src/modules/DamageHealingModule/ConstantHealing.cpp
einola: Yes, good point. I'll use IceMinima::c(), which should be used for this purpose throughout the model. In particular, the thermodynamics melts all ice in the grid cell when the concentration drops below this limit.

joewallwork (approved):
Great!

PR #900, opened 2025-07-30
Simplify thermodynamics and address instability

### LINKED ISSUE #899: Instability in realistic setup
We still (again) have an instability in the realistic setup. The model blows up after about one day, with very high ice thickness near the boundary. This problem is caused by an error in the thermodynamics routines. When using ``nextsim_thorm.use_thermo_forcing = false``, the model runs fine.

The figure shows the time step when the ice grows from almost zero thickness to 4749.6 m (encircled in red). The model crashes in the following time step.

Image

einola:
# Simplify thermodynamics and address instability

Fixes #899 

### Task List

---
# Change Description

I had problems following the program flow when trying to address #899. To address this, I reduced the number of variables carried in the ModelArrayRef system by throwing out any and all concentration, thickness, snow, and damage variables, save one of each. In particular:

 - Don't save a separate "slab thickness" variable. The thermodynamics routines are easily modified to deal with this.
 - Don't save the state from the start of the time step. Any module that needs this information can easily deduce it on its own.

The model is much more stable after doing this; I haven't directly discovered what was causing the problem, but the model is stable now.

---
# Test Description

All standard tests run. I had to modify the test for the column-only unit test. It failed with errors of the order of a mm after 100 years. I don't think this is a problem.

I have also run it for 5 months already with full dynamics and thermodynamics using the 25 km resolution setup. A year-long run is running now.

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/core/src/PrognosticData.cpp
+++ b/core/src/PrognosticData.cpp
@@ -24,9 +24,6 @@ PrognosticData::PrognosticData()
     , pDynamics(0)
 
 {
-    getStore().registerArray(Protected::H_ICE, &hice, RO);
-    getStore().registerArray(Protected::C_ICE, &cice, RO);
-    getStore().registerArray(Protected::H_SNOW, &hsnow, RO);
     getStore().registerArray(Shared::DAMAGE, &damage, RW);
     getStore().registerArray(Shared::H_ICE_DG, &hice, RW);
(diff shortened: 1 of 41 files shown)


timspainNERSC (approved):
I'm surprised you can delete so many `ModelArrayRef`s and still have the model work. But if it does then great.

PR #908, opened 2025-08-11
Add missing Python requirements

joewallwork:
@niravshah241 pointed out that `netCDF4` is missing from the list of Python requirements. I noticed a couple of others that should be added, too, and applied `sort`. I also added a section to the docs on installing Python dependencies.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -111,7 +111,7 @@ Python dependencies
 -------------------
 
 To install the Python dependencies, first ensure that you have a Python
-environment set up. You can use `conda` or `virtualenv` for this purpose. Once
+environment set up. You can use ``conda`` or ``virtualenv`` for this purpose. Once
 you have created and activated your Python environment, you can install the
 required Python packages with
 .. code::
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -110,6 +107,17 @@ To unload an environment you can use the handy (and very fun) alias of ``despack
 information on using ``spack`` environments please see `using environments
 <https://spack.readthedocs.io/en/latest/environments.html#using-environments>`_.
 
+Python dependencies
+-------------------
+
+To install the Python dependencies, first ensure that you have a Python
+environment set up. You can use `conda` or `virtualenv` for this purpose. Once
```
TomMelt: small comment but I think you need double back ticks for this e.g., 

```suggestion
environment set up. You can use ``conda`` or ``virtualenv`` for this purpose. Once
```

TomMelt (approved):
Thanks @jwallwork23 . LGTM other than the backticks

docs/installation.rst
joewallwork: Good catch!

PR #909, opened 2025-08-11
Python linting with Ruff

### LINKED ISSUE #405: format python codes
I thought it would be a good idea to standardize the formatting of the python scripts using a tool such as [`black`](https://black.readthedocs.io/en/stable/).

This will need to be done in one go and quickly pushed back into master. Otherwise it could create annoying diffs.

### LINKED ISSUE #891: Applying auto-formatting to Python files, as well as C++
We could do with applying auto-formatting to Python files, as well as C++. Probably best to do after #889 is merged, though. (Not for this PR.)

_Originally posted by @jwallwork23 in https://github.com/nextsimhub/nextsimdg/pull/885#discussion_r2213131114_

joewallwork:
# Python linting with Ruff

Fixes #891

This PR adds [ruff](https://github.com/astral-sh/ruff) for Python linting, which is used widely by the community.

#### Useful info

* With `ruff check` we can see the current issues.
* With `ruff check --fix` we can get ruff to automatically fix many issues that are guaranteed not to break things.
* There is also a `ruff check --fix --unsafe-fixes` option but it should be used with care.
* There is also `ruff format` but it's quite opinionated so would likely completely reformat everything.
* `ruff.toml` contains the configuration, including any rules to ignore.
* The vast majority of the lines changed are in `test/era5_topaz4_test_data.py` because this is an enormous file.
* At some point in working on this PR, running `ruff check --statistics` gave the following:

```
58	D103   	undocumented-public-function
52	PLR2004	magic-value-comparison
27	D100   	undocumented-public-module
11	S101   	assert
 7	PLR0911	too-many-return-statements
 4	PLR0912	too-many-branches
 3	ARG001 	unused-function-argument
 3	PLR0915	too-many-statements
 2	E722   	bare-except
 2	PLR0913	too-many-arguments
 2	S602   	subprocess-popen-with-shell-equals-true
 1	A001   	builtin-variable-shadowing
 1	B007   	unused-loop-control-variable
 1	S607   	start-process-with-partial-path
Found 174 errors.
```

To get things to work, I applied ruff fixes, made code changes, and ignored rules that'd require more work to satisfy. The ignored rules are:
* Magic value comparison
* Too many statements/branches/arguments/return statements
* Subprocess call with `shell=True`
* Start process with partial path
* Missing docstrings for public modules/functions

I also applied per-file ignores to some of the code generation code for use of assertions as they'd need some refactoring to remove them and retain checks.

#### Checklist

* [x] Make `ruff check` pass by fixing issues and ignoring rules, as appropriate.


joewallwork:
[Rebased on top of `develop`, excluding large data generation scripts from linting.]

core/src/modules/module_maker.py
```diff
@@ -1,39 +1,40 @@
...
+
 def denamespace(nname):
-    """Returns the last element of the name, without any of the qualifying
-    namespaces."""
+    """Return the last element of the name, without any of the qualifying namespaces."""
     return nname.split(":")[-1]
 
 def generator(module_class_name, fq_interface_name, fq_impl_names, help_names):
-    """Generates the text for a module support file."""
+    """Generate the text for a module support file."""
```
timspainNERSC, line 8: Is there a reason the prescriptive second person form (generate), rather than the descriptive third person form (generates)? Admittedly I got this from writing javadoc comments many years ago, rather than any Python standard, but I feel it makes more sense to me.

timspainNERSC (approved):
Big, but not complex. I can only apologize for the amount of trailing whitespace.

core/src/modules/module_maker.py
joewallwork: It comes from `D401 First line of docstring should be in imperative mood`. We could ignore this rule if preferred?

timspainNERSC: No, I'll try to follow that rule in future.

PR #920, opened 2025-08-29
Python formatting with Ruff

### LINKED ISSUE #405: format python codes

joewallwork:
# Python formatting with Ruff

Fixes #405
Follows #909

Having introduced linting for Python files with `ruff check`, this PR goes on to apply formatting with `ruff format`. This is more opinionated than `ruff check` as you'll see that there were a lot of changes, despite the codebase already passing the Python lint check.

The only option I've configured so far is to have a indent length of 2, which I'd seen in various places.

I've added `# fmt: off` and `# fmt: on` comments in a few places to avoid formatting of 2D arrays, as requested in #405.

## Questions for reviewer

* Indent of 4 would be more standard, should we switch to that?
* Are there any other options to be configured?


joewallwork:
[Rebased on top of `issue891_ruff`, which had been rebased on top of `develop`.]

.github/workflows/linting.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       run: |
         cd ${{ github.workspace }}
         pip install ruff
+        ruff format --diff ./
```
einola, line 41: I guess yo ucan remove ``ruff check`` if you've already done a ``ruff format``, right?

.cmake-lint-config.py
einola, line 1: Why are you changing this file? Is it a mistake?

einola (changes requested):
This is so big that it's essentially impossible to review - so I think we'll just assume that ruff format hasn't messed anything up (and the tests should catch this, if it's serious).

I've commented on two files to check, otherwise it's all good.

.cmake-lint-config.py
joewallwork: It was formatted with Ruff. If you look closely at the changes, it's actually just moved things around.

.github/workflows/linting.yml
joewallwork: `ruff format` doesn't raise an error if it makes any changes, so the subsequent `ruff check` will do that.

.cmake-lint-config.py
einola: Ah, yes. I was a bit quick here. I didn't expect the cmake lint config file to be a python script :)

einola (approved):
I forgot to say that I would be happy to have an indent 4 spaces and not 2. That way it's more standard. So if you just change that and rerun ruff on everything, then it's good for me.

joewallwork:
> I forgot to say that I would be happy to have an indent 4 spaces and not 2. That way it's more standard. So if you just change that and rerun ruff on everything, then it's good for me.

Thanks @einola! Hm my comment must be out of date because looking at the Python files they're all using the default indentation of 4 spaces, not 2. Sorry for the confusion.

I spotted an unwanted formatting change in a0ff730f3d3a41a5db22bfb04db8dcce418b6320 and fixed it with a `# fmt: off` protection. I'll merge this once tests pass.

Issue #247, opened 2023-02-16
Create differential tests on output data prior to merging MPI and OpenMP solutions


dorchard:
(This issue is to start a discussion on testing strategies)

Once we get to the point of merging MPI and OpenMP parts of the code base it would seem useful to have end-to-end tests that can compare output runs to some 'golden outputs' so that we can catch any problems, and assess with any changes are within acceptable error bounds (e.g., due to re-association of float point operations coming from parallelisation). Calculating RMSE (or something similar) between the output and the golden output would seem a useful measure here (we can also just do bitwise comparison as well - although I suspect that meeting that would be optimistic). This would generally help I think with code optimisations. Sometimes the golden outputs would need to change (due to scientific reasons) but other times its good to have them to assess changes that we want to be semantics preserving (e.g., optimisations).

dorchard:
This issue depends on https://github.com/nextsimdg/nextsimdg/issues/334.

PR #924, opened 2025-09-09
Add Halo Exchange to Dynamics

### LINKED ISSUE #120: MPI parallelization of Dynamics

TomMelt:
# Add Halo Exchange to Dynamics

Fixes #120 

### Task List
- [ ] ~Defined the tests that specify a complete and functioning change~ will be done in separate follow up PR (see issue #1071 

---
# Change Description

This PR adds halo exchange functionality to the dynamics and for advected fields.

After adding halo exchange into the dynamics we fix fields like `hice` for example.

image

**Figure 1** `hice` (left) before halo exchange is added to dyanmics and (right) after.

After adding halo exchange into the `advectField` function we also fix it for advected fields like `tbottom`, `tsurf` and `tinterior`, for example.

image

**Figure 2** `tbottom` (left) before halo exchange is added to `advectField` and (right) after.

---
# Test Description

None currently

---
# Documentation Impact

None currently

---
# Other Details

### :warning: Open Problems:
* AFAIK for `Interpolations::DG2CG(smesh, cgVector, DGVector)` values inside the "inner" domain (i.e., not halo cells) depends on vertices set outside the inner cells i.e., from neighbouring points on the lattice. This means that you have to halo exchange the DGVector before you halo exchange the CGVector. It is not equivalent to first interpolate from DGVector onto CGVector and then halo exchange the CGVector.
*  We still don't have a solution for how to set vertices outside the main domain (so inner halo boundaries are fine, but points along the edge are not (also periodic boundaries would be an issue)
* When should halo exchange take place?
  * @Thanduriel raised point that we perhaps need to exchange after dynamics rather than before
  * @timspainNERSC think it should be before and possibly after
  * @TomMelt thinks just before should be enough
  * @einola either should work

### Todo

- [x] add halo exchange automatically for advected fields
- [ ] ~Need some scientific/statistical/sense tests to ensure correctness (raised in issue #247) -- perhaps an idealized test case?~ will be done in separate follow up PR (see issue #1071

### DIFF
--- a/dynamics/src/CGDynamicsKernel.cpp
+++ b/dynamics/src/CGDynamicsKernel.cpp
@@ -19,6 +19,10 @@
 
 #include <limits>
 
+#ifdef USE_MPI
+#include "include/Halo.hpp"
+#endif
+
 namespace Nextsim {
 
 template <int DGadvection>
@@ -272,6 +276,14 @@ template <int DGadvection> void CGDynamicsKernel<DGadvection>::prepareIteration(
     Interpolations::DG2CG(*smesh, cgA, data.at(ciceName));
     VectorManipulations::CGAveragePeriodic(*smesh, cgA);
 
+#ifdef USE_MPI
+    // Halo object only depends on shape of the array, so we can share it for all CGVectors of the
+    // same shape
+    Halo halo(cgH);
+    halo.exchangeHalos(cgH);
+    halo.exchangeHalos(cgA);
+#endif
+
     // Reinit the gradient of the sea surface height. Not done by
     // DataMap as seaSurfaceHeight is always dG(0)
     computeGradientOfSeaSurfaceHeight(DynamicsKernel<DGadvection, DGstressComp>::seaSurfaceHeight);
@@ -431,6 +443,12 @@ template <int DGadvection>
 DGVector<DGadvection>& CGDynamicsKernel<DGadvection>::advectDGVField(
     double timestep, DGVector<DGadvection>& field, double lowerLimit, double upperLimit)
 {
+
+#ifdef USE_MPI
+    Halo halo(field);
+    halo.exchangeHalos(field);
+#endif
+
     dgtransport->step(timestep, field);
 
     //! Slope Limiting
--- a/dynamics/src/include/BrittleCGDynamicsKernel.hpp
+++ b/dynamics/src/include/BrittleCGDynamicsKernel.hpp
@@ -15,6 +15,10 @@
 #include "include/constants.hpp"
 #include <cmath>
 
+#ifdef USE_MPI
+#include "include/Halo.hpp"
+#endif
+
 namespace Nextsim {
 
 // The brittle momentum solver for CG velocity fields
@@ -124,6 +128,12 @@ public:
         avgU.zero();
         avgV.zero();
 
+#ifdef USE_MPI
+        // Note: we only need to create one halo object per array type,
+        // dimensionality and size.
+        Halo halo(u);
+#endif
+
         for (size_t subStep = 0; subStep < params.nSteps; ++subStep) {
 
             projectVelocityToStrain();
@@ -139,6 +149,11 @@ public:
             updateMomentum(tst);
 
             applyBoundaries();
+
+#ifdef USE_MPI
+            halo.exchangeHalos(u);
+            halo.exchangeHalos(v);
+#endif
 
             // Land mask
         }


Thanduriel:
After looking at the code again I now believe that the update is correct as is. The exchange always happens right after a field is updated, ensuring that the correct results are available for subsequent computations. 

The only exception are the advected fields in the dynamics that are also modified by the thermodynamics. Here we need to ensure that the halo cells contain the correct values before the advection, either by doing the column physics on the halo cells or by doing an exchange beforehand. Another exchange afterwards should not be strictly necessary but if we do another exchange, I think it might be possible to skip the exchange of stresses (s11,s12,s22), since that computation works purely local on individual cells .

dynamics/src/CGDynamicsKernel.cpp
```diff
@@ -272,6 +276,14 @@ template <int DGadvection> void CGDynamicsKernel<DGadvection>::prepareIteration(
     Interpolations::DG2CG(*smesh, cgA, data.at(ciceName));
     VectorManipulations::CGAveragePeriodic(*smesh, cgA);
 
+#ifdef USE_MPI
+    // Halo object only depends on shape of the array, so we can share it for all DGVectors of the
```
Thanduriel: In this case it should read CGVectors, not DGVectors

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -116,6 +120,20 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
...
+        // halo exchange
+        //  - damage, hice, cice
+        //  - s11, s12, s22
+        //  - e11, e12, e22
+        //  - dStressY, dStressX
+        //  - u, v
+        // Note: only need to create one halo object per array type, dimensionality and size.
+#ifdef USE_MPI
+        Halo halo(hice);
+        halo.exchangeHalos(static_cast<DGVector<DGadvection>&>(hice));
```
Thanduriel: I think `hsnow` is supposed to be treated like `hice`, even if is not (yet) used in the dynamics. So perhaps it should be exchanged as well @einola ?

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -128,18 +146,42 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
...
             std::array<std::reference_wrapper<DGVector<DGstressComp>>, N_TENSOR_ELEMENTS> stress
                 = { s11, s12, s22 }; // Call the step function on the StressUpdateStep class
             // Call the step function on the StressUpdateStep class
             stressStep.stressUpdateHighOrder(
                 params, *smesh, stress, { e11, e12, e22 }, hice, cice, deltaT);
 
+#ifdef USE_MPI
+            haloDGVector.exchangeHalos(s11);
+            haloDGVector.exchangeHalos(s12);
+            haloDGVector.exchangeHalos(s22);
```
Thanduriel: I think this exchange is redundant. If all the inputs are correct on the halo cells the computed stress should be as well. Maybe you can try this? I expect the results to be bit-wise identical.

Thanduriel (changes requested):
Looks correct to me as-is but I have a few thoughts on possible changes.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
timspainNERSC: Yes, `hsnow` should be exchanged here.

einola: If we modify ``advectDynamicsFields`` to call ``advectField`` instead of ``advectDGVField``, then the halo exchange is included there and not needed here.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -128,18 +146,42 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
 
             projectVelocityToStrain();
 
+#ifdef USE_MPI
+            Halo haloDGVector(e11);
+            haloDGVector.exchangeHalos(e11);
+            haloDGVector.exchangeHalos(e12);
+            haloDGVector.exchangeHalos(e22);
```
einola: Is this also redundant @Thanduriel ? u and v are correct, so it seems to me it should be, but I'm not sure.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
```diff
@@ -128,18 +146,42 @@ template <int DGadvection> class BrittleCGDynamicsKernel : public CGDynamicsKern
...
+            haloDGVector.exchangeHalos(s12);
+            haloDGVector.exchangeHalos(s22);
+#endif
+
             stressDivergence(); // Compute divergence of stress tensor
 
+#ifdef USE_MPI
+            Halo haloCGVector(dStressX);
+            haloCGVector.exchangeHalos(dStressX);
+            haloCGVector.exchangeHalos(dStressY);
```
einola: And I also wonder if this one is redundant.

dynamics/src/CGDynamicsKernel.cpp
einola:
```suggestion
    // Halo object only depends on shape of the array, so we can share it for all CGVectors of the
```

That's all that needs correcting - right?

einola (changes requested):
We should modify ``advectDynamicsField`` as per my comment on line R132.

There are also questions about redundant exchanges.

einola:
A possible test would be to run ``config_benchmark.cfg`` with one of the high-resolution setups (e.g., ``init_benchmark_256x256.nc``) and simply check for conservation of volume. That is to say, to check if the total sum of ``hice`` in the domain is fixed.

This does not check the thermodynamics.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
Thanduriel: Yes, I think so too.

dynamics/src/include/BrittleCGDynamicsKernel.hpp
Thanduriel: I think the outermost nodes (in the halo) of these cG fields will be wrong after the update but the error should not propagate further in the momentum update and is then overwritten.

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -162,6 +166,10 @@ template <int DGadvection, int DGstress> class DynamicsKernel {
         double upperLimit = std::numeric_limits<double>::infinity())
     {
         DGVectorHolder<DGadvection> holder(field);
+#ifdef USE_MPI
+        Halo halo(holder);
+        halo.exchangeHalos(static_cast<DGVector<DGadvection>&>(holder));
+#endif
```
TomMelt: this can be removed

Thanduriel (approved):
Looking good now!

What is the plan for the VP dynamics? Applying the changes from the `BrittleCGDynamicsKernel` shouldn't be much work now. I approve this regardless but if we don't do it right away, maybe create an issue?

dynamics/src/CGDynamicsKernel.cpp
TomMelt: thanks this has been fixed

dynamics/src/include/BrittleCGDynamicsKernel.hpp
TomMelt: Thanks the halo exchange is only done in `advectDGVField` now.

dynamics/src/CGDynamicsKernel.cpp
TomMelt: thanks this has also been done

TomMelt:
> Looking good now!
> 
> What is the plan for the VP dynamics? Applying the changes from the `BrittleCGDynamicsKernel` shouldn't be much work now. I approve this regardless but if we don't do it right away, maybe create an issue?

Thanks see issue #1072

einola (approved):

PR #932, opened 2025-10-01
Introduce static `TimePartition` struct for `Iterator`

joewallwork:
# Introduce static `TimePartition` struct for `Iterator`

Fixes #925

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

This PR follows the recommendation in #925 to reduce duplication of configuration code by introducing a `TimePartition` struct and a static instance of it associated with the `Iterator` class. This gets set up during `Model.configure` and may be accessed from other classes via the `Iterator::getTimePartition` interface.

---
# Test Description

The XIOS tests are updated to be configured by `Model`, so all of them now create an instance of this at the start. However, they currently use a reduced `configureTime` because setting up restart and partition files seems unnecessary for most of them.

---
# Documentation Impact

The XIOS doc pages are updated to reflect these changes.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] This change conforms to the conventions described in the README

PR #936, opened 2025-10-02
Issue 932: Metadata time

timspainNERSC:
# Add time limits to `ModelMetadata`

Fixes #932 (partially)

---
# Change Description

Remove the time limit calculations (such as they are) from `Iterator` to `ModelMetadata`. Also, store the time limits in `ModelMetadata` and output them to the restart file through `CommonRestartMetadata`. As an added advantage, these values are now actually written to the restart file.

---
# Test Description

Ran the standard model configuration by hand to make sure the run length was still correct. The integration tests in the CI suite will provide automated testing.

---
# Documentation Impact

None: purely an internal change.

---
# Other Details

As it stands, this may currently break XIOS, as no configuration related to start time, stop time or time stepping is now passed to XIOS.


core/src/Model.cpp
```diff
@@ -63,20 +62,25 @@ void Model::configure()
...
+
+    if (runLengthStr.empty()) {
+        if (stopTimeStr.empty()) {
+            throw std::invalid_argument(std::string("At least one of ") + keyMap.at(STOPTIME_KEY)
+                + " or " + keyMap.at(RUNLENGTH_KEY) + " must be set");
+        } else {
+            m_etadata.setTimes(startTimeStr, TimePoint(stopTimeStr), stepStr);
+        }
+    } else {
+        m_etadata.setTimes(startTimeStr, Duration(runLengthStr), stepStr);
```
joewallwork, line 81: I guess when strings get passed to the first and third arguments these get converted into `TimePoint`s? I hadn't realised this would work.

core/src/Xios.cpp
```diff
@@ -112,11 +112,7 @@ bool Xios::doOnce()
 }
 
 //! Get the configuration for the XIOS handler
-ConfigMap Xios::getConfig() const
-{
-    ConfigMap cMap = ModelConfig::getConfig();
-    return cMap;
-}
+ConfigMap Xios::getConfig() const { return ConfigMap(); }
```
joewallwork, line 115: I was confused why this didn't break anything but then realised `Xios::getConfig` isn't actually used anywhere. Will fix that in a follow-up PR.

joewallwork (approved):
Thanks @timspainNERSC, this looks good to me - much better approach than what I was thinking of in #932. Will close that PR now.

core/src/Model.cpp
timspainNERSC: If type `A` has a constructor that takes type `B` as an argument, then an instance of type `B` can be implicitly converted to a new `A`.

It's not quite like Python, but C++ does allow some non-explicit type conversions.

PR #928, opened 2025-09-18
XIOS implementation of `readForcingTimeStatic`

joewallwork:
# XIOS implementation of `readForcingTimeStatic`

Fixes #916

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

This PR provides an XIOS implementation of the `readForcingTimeStatic` function. Doing so requires introducing a new `XiosForcing` category to account for forcings as well as restarts.

This PR works under the assumption that the forcings data are found in the same input file used for restarts. I plan to account for separate forcings and restart files when addressing #929.

Other changes to facilitate the implementation:
* Switch `getCurrentDate` to return a `TimePoint` rather than string.
* Separate out `configGetFieldNames` for the different categories (input, output, forcing).

---
# Test Description

Tests for reading forcings have been added into the existing `XiosRead_test`.

---
# Documentation Impact

The XIOS section of the documentation is updated to mention the new `XiosForcing` section.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] This change conforms to the conventions described in the README

PR #937, opened 2025-10-02
Issue 931: Old-style restart files

### LINKED ISSUE #931: ParaGridIO no longer reads old-style restart files.
Following the twin principles of being liberal what you accept and not annoying me excessively, `ParaGridIO` should still read old-style restart files containing the `data`group.

timspainNERSC:
# Read old-style restart files

Fixes #931

---
# Change Description

Adds the necessary code to allow the old-style restart files with netCDF groups to still be read. This involves making the ParaGridIO code generic so that dimensions and variables can be searched for both in the file root and in the `data` group. All newly written files are in the flat new style.

---

# Test Description

The model successfully runs using my old style restart and forcing files.

---
# Documentation Impact

Since no-one else should be generating the old-style files, this does not really need to be documented. Plus, I should eventually replace my old-style restart files over time.


core/src/include/StructureFactory.hpp
joewallwork, line 1: Note that this will conflict with @TomMelt's changes in #744, although it'll be possible to rectify the two.

joewallwork (changes requested):
Nice to have backward compatibility for reading files.

I have a couple of requests:
1. Perhaps add a `Logged` warning when reading files through the legacy option to say that this is being done? Could even include a deprecation warning, if you think that'd be appropriate.
2. In the `ParaGridIO_Xios` implementation of `getModelState`, perhaps we should raise an error if groups are found, with a message saying that the legacy format is not supported in the XIOS implementation? (I'm not sure how readable the XIOS errors would be.) I can then do the same for `readForcingTimeStatic` in #928.

Also note that some of the other code changes that align the MPI and non-MPI cases will conflict with similar changes in #744.

timspainNERSC:
It might be easier to start from scratch.
> **This branch has conflicts that must be resolved**
> * core/src/Model.cpp
> * core/src/ParaGridIO.cpp
> * core/src/RectGridIO.cpp
> * core/src/StructureFactory.cpp
> * core/src/include/ParaGridIO.hpp
> * core/src/include/RectGridIO.hpp
> * core/src/include/StructureFactory.hpp
> * core/src/modules/StructureModule/include/ParametricGrid.hpp
> * core/src/modules/StructureModule/include/RectangularGrid.hpp
> * core/src/modules/include/IStructure.hpp
> * core/test/ParaGrid_test.cpp
> * core/test/RectGrid_test.cpp
>

timspainNERSC:
>  2. In the `ParaGridIO_Xios` implementation of `getModelState`, perhaps we should raise an error if groups are found, with a message saying that the legacy format is not supported in the XIOS implementation? (I'm not sure how readable the XIOS errors would be.) I can then do the same for `readForcingTimeStatic` in #928.

What would be a sensible way of doing this? As far as I can tell, `getModelState` no longer does anything with the `filename` parameter, it (presumably) having been passed to the XIOS set up beforehand. It rather seems overkill to re-open the netCDF file from the passed `filename`, check for groups and then close it again.

joewallwork:
> What would be a sensible way of doing this? As far as I can tell, `getModelState` no longer does anything with the `filename` parameter, it (presumably) having been passed to the XIOS set up beforehand. It rather seems overkill to re-open the netCDF file from the passed `filename`, check for groups and then close it again.

Ah, good point. There's already an initial read of the NetCDF file in `Xios::affixModelMetadata` within the conditional that checks if an input file is provided so that would be the best place to add it.
https://github.com/nextsimhub/nextsimdg/blob/af75f0793cd4f8fddd513250180e7ada6905582d/core/src/Xios.cpp#L673-L680

Issue #436, opened 2023-11-13
Add information about updating the documentation


MarionBWeinzierl:
Major code changes may require updates or changes of the [documentation on readthedocs](https://nextsim-dg.readthedocs.io/en/latest/index.html), so that this stays in line with the code. 

Therefore, we need information in the README on how update the docs, and also when to update it (change on main? change on develop?). 

@auraoupa , are you able to provide this information, and potentially access to readthedocs, if needed?

auraoupa:
readthedocs documentation is managed on the docs branch and in the docs repo, that is why when you click on "Edit on github" it will bring you there. Modifying the docs branch will then trigger the building, and you are able to see the outcome here : https://readthedocs.org/projects/nextsim-dg/builds/

Also there is an automatic documentation built with doxygen that is managed with a [github workflow](.github/workflows/doxygen.yml) and is visible here : https://nextsimhub.github.io/nextsimdg/index.html

I will modify soon the README to mention all of this in the front page

MarionBWeinzierl:
Done in #445

TomMelt:
@auraoupa , do you have any instructions on how to build the readthedocs docs locally? Or even the doxygen?

auraoupa:
Theoretically it should be possible but I've never done it, because of all the automation that is hidden in readthedoc and the automatic doxygen run that we do in this [github workflow](https://github.com/nextsimhub/nextsimdg/blob/develop/.github/workflows/doxygen.yml) 
The documentation is developped in the docs branch so it is easy to reverse a wrong manipulation, but the output will change live which is not great for sure. 
@stephanieleroux do you have more insights on this matter ?

TomMelt:
Also is it possible to build the documentation for two branches (via `readthedocs`) i.e., `main` and `develop` instead of just `main`.

auraoupa:
Actually it is synchronized with the ```docs``` branch not ```main``` and I could not find a way to synchronize with multiple branches when I looked for it. That is why I rebase the ```docs``` branch from ```develop``` from time to time

PR #942, opened 2025-10-03
Local docs

TomMelt:
# Build docs locally

Add `Dockerfile` to support building the docs locally. This is extremely helpful for debugging changes to the documentation without having to push to the `docs` branch

- adds `Dockerfiles/Dockerfile.sphinx` to create a docker image
- adds explanation of how to build and run docker image in `docs/installation.rst`

relates to closed issue #436

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,18 @@
 /docs/html/
 /docs/latex/
 /docs/xml/
+/docs/_build/
+
+# ignore all files in docs/api except for manually maintained rst files
+/docs/api/*
+!/docs/api/class.rst
+!/docs/api/counttimercpp.rst
+!/docs/api/function.rst
+!/docs/api/index.rst
+!/docs/api/num.rst
+!/docs/api/structure.rst
+!/docs/api/typedef.rst
+
 /.project
 __pycache__/
 /.pydevproject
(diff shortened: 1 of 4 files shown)


joewallwork (commented):
The docs build works, although I get a "Access to the file was denied" error when I try to open any of the HTML files. Is that expected as far as this PR goes?

TomMelt:
> The docs build works, although I get a "Access to the file was denied" error when I try to open any of the HTML files. Is that expected as far as this PR goes?

Yes, unfortunately this is a `docker` thing. You probably run docker as root? If you check the permissions of the `_build` folder it will probs say `root:root`. Could you try something like: `chown -R $USER:$USER _build`  and see if you can open `index.html`?

TomMelt:
If you have any suggestions for a nice portable way to run docker not as root, I would definitely appreciate it. My next best option is to instead make a copy of the docs inside the docker container, build them and serve them with `nginx` or something, and then port forward back to the host, so you could use something like `http:://localhost:8000` to view the docs. But that felt like more work :sweat_smile:

joewallwork (approved):
> Yes, unfortunately this is a docker thing. You probably run docker as root? If you check the permissions of the _build folder it will probs say root:root. Could you try something like: chown -R $USER:$USER _build and see if you can open index.html?

Ah, yes - I can confirm that that works. Should be sufficient for now I think.

PR #951, opened 2025-10-14
XIOS: Handle `CGField`

### LINKED ISSUE #911: XIOS: Handle `CGField` and `DGSField`
Related to #831.

In addition to `DGField` and `VertexField`, we will need to account for `CGField` for handing $u$- and $v$-velocities in restart files, as well as DG stress `DGSField`.

joewallwork:
# XIOS: Handle `CGField`

Fixes #911
Merges into #952 

### Task List

---
# Change Description

Currently the XIOS implementation accounts for `HField`, `VertexField`, and `DGField`. #952 extends to cover `DGSField`. This PR extends further to cover `CGField`.

This involved checking the global and local dimensions and start index are set up correctly, as well as the XIOS domain, grid, and field.

Several error messages were also made clearer as part of this change.

---
# Test Description

The XIOS read and write tests are updated to account for the CG field type.

A minor looping issue is also fixed in the value checking - it was missing the final value for `VertexField`. As such, the value checking code in `XiosRead_test` was reworked to loop over fields first and then dimensions. To make this easier to check, I changed the coordinate field to align with the coordinates of the grid (assuming unit spacing in each direction).

Note that the file reading code accounts for halos out-of-the-box! :angel:

---
# Documentation Impact

`CGField` is mentioned in the XIOS docs page.

---
### Pre-Request Checklist


joewallwork:
[Rebased on top of [issue911_xios-dgs](https://github.com/nextsimhub/nextsimdg/tree/issue911_xios-dgs)]

timspainNERSC (approved):
One small, non-essential suggestion: rather than using `cice` as the name of the CG variable, might I suggest `uice`, as ice velocity is written by the real model as a CG field, whereas ice concentration never will be.

joewallwork:
> One small, non-essential suggestion: rather than using `cice` as the name of the CG variable, might I suggest `uice`, as ice velocity is written by the real model as a CG field, whereas ice concentration never will be.

Thanks @timspainNERSC. I made this change in https://github.com/nextsimhub/nextsimdg/pull/951/commits/8f40681c3cc1a9f443603dd10d6a8c6a8f723d20. Given that that variable was being used for diagnostics, I switched to output `hsnow` as a diagnostic. Let me know if you'd prefer to use something different.

core/src/Xios.cpp
```diff
@@ -695,8 +695,15 @@ void Xios::affixModelMetadata()
                 } else if (dimType == ModelArray::Dimension::YVERTEX) {
                     localLength = metadata.getLocalExtentY() + 1;
                     start = metadata.getLocalCornerY();
+                } else if (dimType == ModelArray::Dimension::XCG) {
+                    localLength = CGDEGREE * metadata.getLocalExtentX() + 1;
+                    start = CGDEGREE * metadata.getLocalCornerX();
+                } else if (dimType == ModelArray::Dimension::YCG) {
+                    localLength = CGDEGREE * metadata.getLocalExtentY() + 1;
+                    start = CGDEGREE * metadata.getLocalCornerY();
```
TomMelt, line 697: FWIW this is also how I am currently determining the CG dims.... but I would like to hear Tim's comments on PR #969 first. I am worried we might end up in a situation where CG dims is set in several different places. I don't think you need to change this, just raising it as an FYI.

core/src/Xios.cpp
```diff
@@ -706,15 +713,21 @@ void Xios::affixModelMetadata()
...
+                        throw std::runtime_error("Xios: inconsistent local dimensions for "
+                            + dimName + ": " + std::to_string(localLength) + " vs. "
+                            + std::to_string(
+                                ModelArray::definedDimensions.at(dimType).localLength));
                     }
                     if (start != ModelArray::definedDimensions.at(dimType).start) {
-                        throw std::runtime_error("Xios: inconsistent start index for " + dimName);
+                        throw std::runtime_error("Xios: inconsistent start index for " + dimName
+                            + ": " + std::to_string(start) + " vs. "
+                            + std::to_string(ModelArray::definedDimensions.at(dimType).start));
```
TomMelt: I think this alleviates my prev. concern about different sizes for CG dims in the code. However, I might need to understand how this works a bit better. Where will this be called in NextSim? Is it only called once?

TomMelt (changes requested):
Thanks @jwallwork23 looks good to me. I don't think you need to change anything but I do want to make sure I understand the plan for CG Vectors before we merge this (and my changes #969). I will mark as request changes for now, but after we discuss next meeting I can probs just approve this.

core/src/Xios.cpp
joewallwork: My subsequent PR #968 will make it a bit clearer, but it's only called once yeah. In that follow-up PR I have it happening when the `Model` gets configured, which in turn calls `ModelMetadata.setTime`.

TomMelt (approved):
Thanks @jwallwork23. It's all good for now

PR #954, opened 2025-10-16
Flush netCDF buffers on write

### LINKED ISSUE #953: Flush netCDF buffers on write
At the moment, the netCDF write buffer is not flushed before the end of the simulation. This may cause an increase in memory usage, although the OS should probably flush the buffer before that becomes a problem. A more significant problem is that if the model crashes in a way that is still not caught (mostly through signals like SIGTERM), the diagnostic file will only have one readable time step, making diagnostic of the model run impossible.

The solution is to call the ``sync()`` function of the relevant ``netCDF::NcFile`` object after writing data to the file. There's a potential performance impact of this, but right now, we must have full outputs after a crash.

einola:
# Flush netCDF buffers on write

Fixes #953

### Task List

---
# Change Description

Call netCDF::NcFile's sync in ParaGridIO::writeDiagnosticTime. This flushes the netCDF buffers, as requested in issue #953

---
# Test Description

I ran the model, using ConfigOutput and frequent outputs (every-other time step). I checked that the time axis of the output file was being updated so that I could monitor the simulation in real time. I then stopped the simulation with ctrl-c and confirmed that the diagnostics file was readable and contained all the output time steps expected.

---
# Documentation Impact

None

---
# Other Details

N/A

---
### Pre-Request Checklist

### DIFF
--- a/core/src/ParaGridIO.cpp
+++ b/core/src/ParaGridIO.cpp
@@ -570,6 +570,9 @@ void ParaGridIO::writeDiagnosticTime(const ModelState& state, const std::string&
 #endif
         }
     }
+
+    // Flush buffer to disc for monitoring and diagnostic purposes.
+    ncFile.sync();
 }
 
 void ParaGridIO::close(const std::string& filePath)


joewallwork (approved):
Good catch!

PR #959, opened 2025-10-17
Add Ruff to pre-commit hook

### LINKED ISSUE #956: Include ruff format in pre-commit hook example
The pre-commit hook example is very useful, and so is using ruff format ... if only the two were merged somehow!

joewallwork:
Closes #956.

This PR adds `ruff format` to the `.pre-commit` hook, as requested by @einola.


.pre-commit
```diff
@@ -10,16 +10,23 @@
...
 
-# Apply clang-format for linting
+# Apply clang-format for C++ linting
 for FILE in $(git diff --cached --name-only | grep -iE '\.(cpp|cc|h|hpp)$'); do
   clang-format -i ${FILE} --verbose
 done
 
-# Apply cmake-lint for linting
+# Apply ruff format for Python linting
+ruff format
```
einola: How about

```suggestion
for FILE in $(git diff --cached --name-only | grep -iE '\.py$'); do
    ruff format ${FILE}
done
```

? It's consistent with the rest.

einola (approved):
Thanks! One suggestion, but it's up to you.

.pre-commit
joewallwork: Ah yes, good point! Avoids reformatting unrelated code.

joewallwork:
I think the Mac test fail is an intermittent one - I've also seen it crop up in https://github.com/nextsimhub/nextsimdg/pull/961.

PR #946, opened 2025-10-08
Improved scripts to create initial conditions files

einola:
# Improved scripts to create initial conditions files

Fixes #945

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

Adds the functionality to make_init_base.py to make generic geographic grids, in addition to idealised ones. The user must supply lat/lon coordinates of the grid and may provide a mask and initial fields. The program can create nextsim grids from a grid of lat/lon coordinates or lat/lon coordinates of a NEMO C-grid (which also provides coordinates of the upper right corner of the grid box).

    Usage:
     0. Import make_init_base
      >> from make_init_base import initMaker
     1. Create an initialiser object with a given filename, e.g.
      >> init = initMaker("test")
         2a. Create a cartesian grid from dimensions and resolution, e.g.
          >> init.make_cartesian_grid(128, 128, 3e3)
         2b. Create a geographic grid from a file containing coordinates. We need to define type of grid:
            * "p": only centre points of the grid are given
            * "ur": the centre and upper right corner of the grid is given
            * "ll": the centre and lower left corner of the grid is given
          init.make_geographic_grid("coordinates.nc", "ur", plat_name="gphit", plon_name="glamt", qlat_name="gphif", qlon_name="glamf")
     3. Modify any variables needed, e.g.
      >> init.cice = 1
      >> init.hice = 3
      >> init.hice[10:20, 10:20] = 4
     4. The init file is written when the initialiser object goes out of scope (e.g. when the program ends or at the end of a loop).

I also modified and renamed make_initNH.py to use make_init_base.py

---
# Test Description

I used the script to recreate the grid of the integration test, as well as the grid from our NEMO-coupled setup and a few others. All of which run successfully. 

---
# Documentation Impact

The functionality of make_init_base.py should be documented in the usage section of the manual, but it is not.

---
# Other Details

N/A

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
[...]

PR #960, opened 2025-10-17
Change double underscore to single

einola:
From PR #946. An error that slipped by me and the CI.


joewallwork (approved):
Ah yes. Good to go provided tests pass.

PR #961, opened 2025-10-17
Overhaul workflows

joewallwork:
I've been thinking about green software engineering practices recently and also made my first contribution to https://github.com/zizmorcore/zizmor so I had some ideas for ways to overhaul the GitHub Actions workflows to make them more efficient, secure, and understandable.

#### Efficiency improvements

* Cancel running jobs if new commits are pushed by setting `concurrency` and `cancel-in-progress: true`.
* Always limit the time limit of jobs with `timeout-minutes`.
* ~~Don't trigger `pull_request_review` if the workflow would already have been triggered.~~
* Switch to a manual `workflow_dispatch` trigger rather than the wasteful `pull_request_review` one.

#### Security improvements

* Hook up Zizmor for security-oriented linting of GitHub Actions workflows.
* Restrict permissions except where write-access is needed.
* Always use `persist-credentials: false` when checking out the repo.
* Use hashes for external GitHub Actions rather than version numbers.

#### Clarity improvements

* Rename `doxygen.yml` as `build_docs.yml` and actually build the docs there as well as running Doxygen.
* Clearer comments throughout.
* More consistent format for the workflows.
* Separate the docs build workflow into a test build while review is ongoing and then a deployment once the PR is merged.


joewallwork:
~~There seems to be an unrelated(?) issue with the Mac build.~~

~~Edit: must have been an intermittent thing.~~

Edit: oh no, it's back...

.github/workflows/build_docs.yml
TomMelt, line 1: Perhaps something we can discuss in the team meeting and maybe for @auraoupa , but why do we have a `doxygen.yml`/`build_docs.yml` anyway? Is it not already handled by ReadTheDocs. Why don't we just link to the readthedocs version?

.github/workflows/docker.yml
```diff
@@ -17,43 +18,52 @@ on:
...
       packages: write
     steps:
       - name: Checkout the repo
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Setup Docker buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435  # v3
```
TomMelt, line 45: is this the new best practice for github actions? to supply commits? or are we doing this because it no longer works with more recent versions?

.github/workflows/test_suite.yml
```diff
@@ -174,14 +193,15 @@ jobs:
...
   test-mac-serial:
-
     runs-on: macos-latest
-
     timeout-minutes: 10
-
     steps:
     - uses: actions/checkout@v2
+      with:
+        persist-credentials: false
```
TomMelt, line 192: I haven't seen this before. What's the benefit? (I mean it sounds more secure...)

TomMelt: I took a look at https://github.com/actions/checkout/issues/485 and I must admit I dont fully understand but it certainly better than the default :melting_face:

scripts/build-docs.sh
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 
 for EXTENSION in cpp hpp ipp; do
-    find . -name "*.${EXTENSION}" -exec /doxygen_prep.sh {} \;
+  find . -name "*.${EXTENSION}" -exec scripts/doxygen_prep.sh {} \;
```
TomMelt: @jwallwork23 I can see how I have caused confusion, but this script is used by the Docker container where doxygen_prep.sh has already been copied to the root directory. I should either change the Dockerfile to account for this, or we should leave this as it was. What do you think is best?

```suggestion
  find . -name "*.${EXTENSION}" -exec /doxygen_prep.sh {} \;
```

TomMelt (changes requested):
Thanks @jwallwork23 , this is much better. I am glad we now cancel running jobs when a new one is launched and we also dont duplicate on pull request when it's already been run. I must admit that had been bothering me, so I am really glad you have fixed it :partying_face: 

I have a few comments for discussion more than anything. The only one that needs resolving before merge is to do with the `build-docs.sh` script

TomMelt:
> ~There seems to be an unrelated(?) issue with the Mac build.~
> 
> ~Edit: must have been an intermittent thing.~
> 
> Edit: oh no, it's back...

@jwallwork23 I took the liberty of rebasing on top of `develop` to incorporate the latest mac ci bugfix #966.

.github/workflows/docker.yml
joewallwork: It's best practice for any external/non-core actions (see https://docs.zizmor.sh/audits/#unpinned-uses).
> This can be a security risk:
>
>  *  Completely unpinned actions can be changed at any time by the upstream repository.
>  *  Tag- or branch-pinned actions can be changed by the upstream repository, either by force-pushing over the tag or updating the branch.

.github/workflows/test_suite.yml
joewallwork: This is a pattern that's recommended whenever checking out the repo. See https://docs.zizmor.sh/audits/#artipacked.

scripts/build-docs.sh
joewallwork: Ah okay I see. I should've modified the `build_docs` workflow to use the same approach as you did in the Dockerfile.

.github/workflows/build_docs.yml
joewallwork: Oh okay, I didn't realise ReadTheDocs already handled this.

scripts/build-docs.sh
joewallwork: I think I fixed it in https://github.com/nextsimhub/nextsimdg/pull/961/commits/7351464f6164d90a87beb7bf67fb1bef7d31e5ee?

joewallwork:
Hmmm the `paths-ignore` logic appears to have been broken for some time: https://github.com/actions/runner/issues/2324. So the `pull_request_review` trigger was still happening.

Instead, I've switched to a `workflow_dispatch` trigger. This will avoid the annoying and wasteful review triggers and instead allow us to run the test suite and linting branches manually as required for merging of the PR. Once this PR is merged, we'll be able to do this from the [Actions](https://github.com/nextsimhub/nextsimdg/actions) tab. We'll be able to click the `Test suite` and/or `Linting` workflow, click **Run workflow** on the right and then select the branch corresponding to the PR.

scripts/build-docs.sh
TomMelt: Yep, that looks good to me :ok_hand:

.github/workflows/build_docs.yml
TomMelt: well I think it does. Hopefully @auraoupa can confirm?

TomMelt (approved):
Thanks @jwallwork23 LGTM now :+1:

.github/workflows/build_docs.yml
auraoupa: Hi, about the documentation there are two as described [here](https://github.com/nextsimhub/nextsimdg?tab=readme-ov-file#documentation) : one is via doxygen (exhaustive description of the code) the other via readthedocs (some API description + hand written pages), they are both triggered by changes in master (before a release for instance) end when the docs branch is updated. Hope it helps ?

auraoupa (approved):

PR #965, opened 2025-10-20
Avoid land nodes when calculating ice-ocean stress

### LINKED ISSUE #964: Land mask in ice-ocean stress calculation
At the moment, the ice-ocean stress is calculated everywhere, both on land and where there's no ice. To save on compute time, neither should be done, but we'll start with not calculating values on land. Calculating values on land nodes may also cause a spillover from uninitialised values, so it's better to set everything to zero at start and then avoid land nodes in all calculations.

einola:
# Avoid land nodes when calculating ice-ocean stress

Fixes #964 

### Task List

---

I set all the CGDynamicsKernel CG variables to zero in the initialise call and then added an "if on land continue" clause to ``updateIceOceanStress``.

---
# Test Description

This has little effect in my setup, since the CG variables are set to all zero on resize by the compiler. There is, however, no guaranty that this happens on different platforms - it's a compiler specific behaviour. I suspect that it may be causing problems on our HPC, but testing is diffucult. I can only do it when Guillaume is back from holidays! I'm putting this in anyway, because it's a reasonable change to make.

---
# Documentation Impact

None

---
# Other Details

None

---
### Pre-Request Checklist

### DIFF
--- a/dynamics/src/CGDynamicsKernel.cpp
+++ b/dynamics/src/CGDynamicsKernel.cpp
@@ -17,7 +17,6 @@
 #include "include/VectorManipulations.hpp"
 #include "include/cgVector.hpp"
 
-#include <cmath>
 #include <limits>
 
 namespace Nextsim {
--- a/dynamics/src/CGDynamicsKernel.cpp
+++ b/dynamics/src/CGDynamicsKernel.cpp
@@ -17,6 +17,7 @@
 #include "include/VectorManipulations.hpp"
 #include "include/cgVector.hpp"
 
+#include <cmath>
 #include <limits>
 
 namespace Nextsim {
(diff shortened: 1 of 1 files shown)


dynamics/src/CGDynamicsKernel.cpp
```diff
@@ -17,6 +17,7 @@
 #include "include/VectorManipulations.hpp"
 #include "include/cgVector.hpp"
 
+#include <cmath>
```
joewallwork: I don't think this is needed?

joewallwork (approved):
This looks good but I think the additional include isn't needed?

dynamics/src/CGDynamicsKernel.cpp
einola: You're right, it isn't, but my IDE added it anyway. I've removed it now.

PR #966, opened 2025-10-20
fix failing mac ci

TomMelt:
This PR fixes the failing mac CI build

- adds a virtual env for `numpy` and `netCDF4`
- uses venv in CMake build and running of tests

Original workflow was failing because CMake autodetects the latest version of python `3.14` but default `pip` installs into `3.13`


einola (approved):
A nice solution.

PR #968, opened 2025-10-24
Automation and pruning for the XIOS handler class

joewallwork:
# Automation and pruning for the XIOS handler class

Towards #850
Merges into #951

### Task List

---
# Change Description

This PR contains refactoring work to (a) automate the XIOS I/O implementation to streamline the user experience and (b) remove member functions that it turns out aren't needed.

#### Automation

* Separate `affixModelMetadata` into `parseInputFiles`, `setupDomains`, `setupAxes`, and `setupGrids`. Call `parseInputFiles` in the `Xios` constructor and the rest in the `ParaGridIO` constructor.
* Automate setting `nCoords` for `VERTEX` in `setupDomains`.
* Automate setting `nComponents` for `DG` in `setupDomains`.
* Automate setting `nComponents` for `DGSTRESS` in `setupDomains`.
* Automate `enableXios` by inlining it at the start of `Model.configureTime`.

#### Pruning

* Inline `enableXios` in `Model.configureTime` as mentioned above.
* Remove `getClientMPISize` and `getClientMPIRank`.
* Remove `setAxisValues` and `getAxisValues`.
* No need to call `ModelMetadata.setTime` because it already gets called in `Model.configureTime`.
* Inline `isInitialized` at the end of the `Xios` constructor, throwing an error if not.
* Drop `getCalendarType` since it's hard-coded to `"Gregorian"`.
* Drop `getCalendarTimestep` because this is given by `ModelMetadata.stepLength`.

---
# Test Description

The XIOS handler tests are reworked to account for the automation. This involved some reordering of how things are done.

---
# Documentation Impact

The XIOS docs page has been updated.

---
# Other information

I changed the default `DGCOMP` value in `discontinuousgalerkin/ModelArrayDetails.cpp` to be consistent with the default DG degree that gets set in the root `CMakeLists.txt`.

---
### Pre-Request Checklist


core/test/XiosCalendar_test.cpp
```diff
@@ -48,13 +46,10 @@ MPI_TEST_CASE("TestXiosCalendar", 1)
     model.configureTime();
 
     // Get the Xios singleton instance and check it's initialized
+    // NOTE: The singleton is created during configureTime
```
timspainNERSC: Because `Xios::getInstance()` is called as part of setting the time, and this runs the code that creates the singleton? Rather than it being explicitly coded as part of `Model::configureTime()`? If the first is the case, then I think this comment could be better worded.

timspainNERSC (approved):
I'd like it if the repeated comment in the test were more clearly worded, but it is technically correct as it stands. It just makes it sound like creating the XIOS singleton was specifically coded as part of the function `configureTime()`, when (as I understand it) that just happens to be the first place that requests the singleton.

core/test/XiosCalendar_test.cpp
joewallwork: Good point! Done in https://github.com/nextsimhub/nextsimdg/pull/968/commits/79c6eec8ade1d31dfdadc4fb0f93332d6240c456.

PR #969, opened 2025-10-27
handle CGDims properly in serial and MPI

TomMelt:
# Handle CGDims properly in serial and MPI

**TODO**

- [x] ~rebase on latest develop to fix failing mac ci once PR #970 is merged~ 

---

Fix failing tests for PR #879 

- Modify `ModelArray::setDimensions` to support MPI local and global dims
- add support for new `setDimensions` function into the tests and code

---
### Background

The underlying problem is that the `x_cg` and `y_cg` dimensions are wrong in the initial model file. For example, when I run the `config_june23.cfg` it uses input file `init_25km_NH.nc`. In there the dimensions are listed as follows:

```
netcdf init_25km_NH {
dimensions:
	ydim = 121 ;
	xdim = 154 ;
	yvertex = 122 ;
	xvertex = 155 ;
	y_cg = 122 ;
	x_cg = 155 ;
	dg_comp = 1 ;
	dgstress_comp = 3 ;
	ncoords = 2 ;
```

They should be `y_cg = 243` and `x_cg = 309`, for `CGDegree=2`. 

These get read in here in `ParaGridIO::getModelState` and then set on line 137.

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/ParaGridIO.cpp#L81-L137

The `localLength` is then overridden later in `BBMDynamics.cpp` on line 125:

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/modules/DynamicsModule/BBMDynamics.cpp#L125

**Note** the function `ModelArray::setDimensions` only sets the `localLength` and not the `globalLength` and that's why you needed to make the changes here:

https://github.com/nextsimhub/nextsimdg/blob/ca3dddb53ae7c7e9a239d2953204a4c1806486c3/core/src/ParaGridIO.cpp#L266-L267

I think this logic needs to be re-written to use the `ModelMetadata` singleton so that it will work for MPI and serial code. Then we need to also modify function `ModelArray::setDimensions` or create a new one that will also set the `globalLength`.


core/src/include/ModelArray.hpp
```diff
@@ -330,7 +330,12 @@ class ModelArray {
      * @param type The type of array the dimensions are to be specified for.
      * @param dim The per-dimension size to be set.
      */
+#ifdef USE_MPI
+    static void setDimensions(Type, const MultiDim&, const MultiDim&);
```
timspainNERSC, line 334: I must admit that I don't like the idea of having a different function signature serial and MPI versions of the code. I wonder if a better solution might be changing the result of `ModelArray::dimensions` from a plain `MultiDim` to a vector of not scalars but local-global pairs in the MPI case.

core/src/modules/DynamicsModule/MEVPDynamics.cpp
```diff
@@ -98,17 +100,27 @@ void MEVPDynamics::setData(const ModelState::DataMap& ms)
...
     }
     // Set the dimensions of CG arrays
+#ifdef USE_MPI
+    auto& metadata = ModelMetadata::getInstance();
+    ModelArray::MultiDim globalDims, localDims;
+    globalDims[0] = metadata.getGlobalExtentX();
+    globalDims[1] = metadata.getGlobalExtentY();
+    localDims[0] = metadata.getLocalExtentX();
+    localDims[1] = metadata.getLocalExtentY();
+    ModelArray::setDimensions(ModelArray::Type::CG, globalDims, localDims);
```
timspainNERSC, line 120: Also, the metadata class could have a function to generate the correct form of the combined local/global dims.

core/test/RectGrid_test.cpp
```diff
@@ -146,7 +146,11 @@ TEST_CASE("Write and read a ModelState-based RectGrid restart file")
 
     // Reset dimensions so it is possible to check if they
     // are read correctly from refeence file
+#ifdef USE_MPI
+    ModelArray::setDimensions(ModelArray::Type::H, { 1, 1 }, { 1, 1 });
```
timspainNERSC, line 150: This would become ```
ModelArray::setDimensions(ModelArray::Type::H, { { 1, 1 }, { 1, 1 } });
```

timspainNERSC (commented):
I think a smarter structure for the dimensions might be useful, rather than changing the function signature of `ModelArray::dimensions` depending on how the model is compiled.

PR #970, opened 2025-10-27
ci: fix mac ci workflow to detect venv

TomMelt:
## Fix failing mac CI

- previous CI was failing because it could not detect `netCDF4` package in the python venv
- This caused the integration test to fail when they couldn't load the auto-generated `.nc` files

This PR fixes that by correctly locating the venv in the CMake build stage

### Notes

`-DPython_ROOT_DIR=/Library/Frameworks/Python.framework/Versions/Current/bin` is no longer required. This was used when we were trying to use the system python. But venv's will be prioritized by default.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -208,8 +208,7 @@ jobs:
         . $HOME/nextsim/bin/activate
         mkdir -p build
         cd build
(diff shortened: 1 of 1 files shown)


joewallwork (approved):
Ah okay, makes sense. Thanks @TomMelt!

PR #974, opened 2025-11-04
Read input file to determine dimensions

### LINKED ISSUE #972: Initial read of input file to get dimensions
Currently, there are differences in how the serial, MPI-parallel, and XIOS configurations read dimensions. For greater consistency, we should do an initial read of the NetCDF dimensions from the input file for all cases. If running in serial or with XIOS, this can be used as needed. If running with MPI, it can be used for verification purposes (checking partition file is consistent).

joewallwork:
# Read input file to determine dimensions

Fixes #972
Merges into #968

### Task List

---
# Change Description

Currently, there are differences in how the serial, MPI-parallel, and XIOS configurations read dimensions. For greater consistency, this PR implements a `setDimensionsFromFile` member function for `ModelMetadata`. It gets called in `Model.configureRestarts`. It also gets called in `ParaGridIO.getModelState` (for consistency with the existing implementation).

---
# Test Description

XIOS tests are updated to account for the fact that dimensions are set in `Model.configureRestarts`.

---
# Documentation Impact

None

---
### Pre-Request Checklist


joewallwork:
[Rebased on top of updated base branch]

joewallwork:
[Rebased on top of `develop` again.]

core/src/ParaGridIO.cpp
```diff
@@ -83,6 +83,10 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
 {
     ModelState state;
 
+    // TODO: Do this once rather than with every read
+    ModelMetadata& metadata = ModelMetadata::getInstance();
+    metadata.setDimensionsFromFile(filePath);
```
joewallwork: Question for reviewer: Do we really want to set dimensions from file in every call to `getModelState`, as in the existing implementation? Would it not be better to do the following?
```suggestion
    ModelMetadata& metadata = ModelMetadata::getInstance();
    if (filePath /= metadata.initialFileName) {
      metadata.setDimensionsFromFile(filePath);
    }
```

timspainNERSC: How many calls to `getModelState()` are expected? In each model run, there should only be one call to `SomethingGrid::getModelState()` and within that one call to `SomethingGridIO::getModelState()` (where `SomethingGrid` is the name of the `IStructure`implementing class). There are a few more examples around the code base for testing, but I believe that is all.

timspainNERSC (approved):
If we were calling `getModelState()` in the main model loop, the extra change would be worth adding. As it should only be called once per run, I'm not sure even the small amount of added complexity is worthwhile.

core/src/ParaGridIO.cpp
joewallwork: Fair point, thanks. I removed the TODO in https://github.com/nextsimhub/nextsimdg/pull/974/commits/17ea449adb17310a1cb1f631a3806b67204d8953 and will merge once tests pass.

joewallwork:
Actually, before merging I should check with @Thanduriel that this PR does what was requested and will be of use in the GPU port?

core/src/Xios.cpp
```diff
@@ -487,83 +487,34 @@ void Xios::parseInputFiles()
 
     // Initial read of the NetCDF file to deduce the dimensions
     for (std::string filename : { inputFilename, forcingFilename }) {
```
Thanduriel: Unnecessary copy.
```
for(const std::string& filename : ...)
```

core/src/ModelMetadata.cpp
```diff
@@ -150,6 +165,85 @@ ModelMetadata::ModelMetadata()
 
 #endif
 
+void ModelMetadata::setDimensionsFromFile(const std::string filename)
```
Thanduriel: Unnecessary copy, could be a reference.
```
void ModelMetadata::setDimensionsFromFile(const std::string& filename)
```

core/src/include/ModelMetadata.hpp
```diff
@@ -71,6 +71,16 @@ class ModelMetadata {
...
 
+    /*!
+     * @brief   Set dimensions information based on the contents of an input file.
+     *
+     * @param   filename the name of the file
+     * @details If an input file hasn't been read yet, the dimensions are read from the file and
+     *          set. Otherwise, a consistency check is made against the dimensions read from file
+     *          and already set.
+     */
+    void setDimensionsFromFile(const std::string filename);
```
Thanduriel: .

Thanduriel (approved):
I solved my buffer initialization problem by doing it lazily on GPU, so as long as the CPU initialization still works it should be fine for me.

I noticed that you sometimes pass large objects (like std::string) by value. This is considered bad practice in general for performance reasons (see the [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-in)). It does not really matter in the places I commented on, so you decide if you want to change it, but I think it is best to simply always follow the linked rule.

core/src/Xios.cpp
joewallwork: Thanks @Thanduriel, fixed in 7c27843a.

core/src/ModelMetadata.cpp
joewallwork: Fixed in 7c27843a.

PR #975, opened 2025-11-04
chore: only print progress on main proc

TomMelt:
This is a mostly stylistic change rather than functional. When you run in parallel e.g., `mpirun -n N` the progress is output `N` times. This PR makes it only print on the main process (proc 0)


core/src/modules/DynamicsModule/BBMDynamics.cpp
```diff
@@ -134,7 +137,14 @@ void BBMDynamics::setData(const ModelState::DataMap& ms)
 
 void BBMDynamics::update(const TimestepTime& tst)
 {
+#ifdef USE_MPI
+    auto& modelMPI = ModelMPI::getInstance();
+    if (modelMPI.getRank() == 0) {
+        std::cout << tst.start << std::endl;
```
TomMelt, line 143: we should probably use a logger here instead. (Can we see if boost log can handle MPI? or can we make it handle MPI?)

PR #983, opened 2025-11-10
Build docs on pushing to develop

joewallwork:
I noticed that the docs only ever get deployed upon pushing to `main` or `docs`, but we agreed that this should also happen upon pushing to `develop`, i.e., when a PR gets merged into `develop`.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -2,9 +2,9 @@
 name: Build docs
 
 on:
-  # Trigger the workflow upon merging into main or docs
+  # Trigger the workflow upon merging into main, develop, or docs
   push:
-    branches: [ main, docs ]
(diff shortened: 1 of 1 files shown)


auraoupa:
Hi @jwallwork23, actually that is not where the readthedocs documentation is built, this workflow concerns the doxygen documentation that we are trying to get rid of ... The trigger for the readthedoc documentation is defined inside readthedocs and has been updated to develop instead of docs.

joewallwork:
> Hi @jwallwork23, actually that is not where the readthedocs documentation is built, this workflow concerns the doxygen documentation that we are trying to get rid of ... The trigger for the readthedoc documentation is defined inside readthedocs and has been updated to develop instead of docs.

Oh okay, I misunderstood. Thanks for explaining this - refreshing the website I can see that it has updated.

PR #996, opened 2025-11-18
XIOS handler restructure

### LINKED ISSUE #850: XIOS: Prune the handler class

### LINKED ISSUE #1000: Avoid unnecessary copies in XIOS handler class
As pointed out in the review of #974, the XIOS handler class includes some unnecessary copies, e.g., of strings. We should go through the code and tidy this up.

While we're at it, the `const` qualifier could also be used more.

joewallwork:
# XIOS handler restructure

Fixes #850
Fixes #1000
~~Merges into #974~~

This will hopefully be the last major restructuring of the XIOS handler class.

The PR includes some pruning:
* Remove unused `enableXios` function.
* Drop errors from the `ParaGridIO_Xios` member functions already thrown in `Xios.read` and `Xios.write`.
* Cut down the `configGetX` member functions.
* Remove the `contextId` and `calendarType` arguments from the constructor because they're now set in `iodef.xml` as of #993.

It also includes better modularisation:
* Extract a new `setupFiles` member function out of the constructor.
* Rename `parseInputFiles` as `setupFields`.
* Split the configuration into `setupClient`, `setupContext`, and `setupCalendar` and have all of them called from `configure` alongside `setupFiles` and `setupFields`.

Other misc changes:
* Additional checking of the calendar setup.
* Reordering of member functions to better group XIOS concepts together.
* Better use of references to avoid unnecessary copies.
* Use of `const` qualifier in more places.

### DIFF
--- a/core/src/ParaGridIO_Xios.cpp
+++ b/core/src/ParaGridIO_Xios.cpp
@@ -84,10 +84,6 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
     std::set<std::string> restartFieldIds = xiosHandler.configGetInputRestartFieldNames();
     for (auto& entry : state.data) {
         const std::string fieldId = entry.first;
-        if (!xiosHandler.getFieldReadAccess(fieldId)) {
-            throw std::runtime_error("ParaGridIO::getModelState: field " + fieldId
-                + " is not configured for reading, but is being read from file.");
-        };
         if (restartFieldIds.count(fieldId) == 0) {
             throw std::runtime_error(
                 "ParaGridIO::getModelState: field " + fieldId + " is not configured as a restart.");
@@ -121,10 +117,6 @@ ModelState ParaGridIO::readForcingTimeStatic(
     // Get all forcings and load them into a new ModelState
     std::set<std::string> forcingFieldIds = xiosHandler.configGetForcingFieldNames();
     for (const std::string& fieldId : forcings) {
-        if (!xiosHandler.getFieldReadAccess(fieldId)) {
-            throw std::runtime_error("ParaGridIO::readForcingTimeStatic: forcing " + fieldId
-                + " is not configured for reading, but is being read from file.");
-        }
         if (forcingFieldIds.count(fieldId) == 0) {
             throw std::runtime_error("ParaGridIO::readForcingTimeStatic: field " + fieldId
                 + " is not configured as a forcing.");
@@ -160,10 +152,6 @@ void ParaGridIO::dumpModelState(const ModelState& state, const std::string& file
     std::set<std::string> restartFieldIds = xiosHandler.configGetOutputRestartFieldNames();
     for (auto entry : state.data) {
         const std::string fieldId = entry.first;
-        if (xiosHandler.getFieldReadAccess(fieldId)) {
-            throw std::runtime_error("ParaGridIO::dumpModelState: field " + fieldId
-                + " is not configured for writing, but is being written to file.");
-        };
         if (restartFieldIds.count(fieldId) == 0) {
             throw std::runtime_error("ParaGridIO::dumpModelState: field " + fieldId
                 + " is not configured as a restart.");
@@ -186,10 +174,6 @@ void ParaGridIO::writeDiagnosticTime(const ModelState& state, const std::string&
     std::set<std::string> diagnosticFieldIds = xiosHandler.configGetDiagnosticFieldNames();
     for (auto entry : state.data) {
         const std::string fieldId = entry.first;
-        if (xiosHandler.getFieldReadAccess(fieldId)) {
-            throw std::runtime_error("ParaGridIO::writeDiagnosticTime: field " + fieldId
-                + " is not configured for writing, but is being written to file.");
-        };
         if (diagnosticFieldIds.count(fieldId) == 0) {
             throw std::runtime_error("ParaGridIO::writeDiagnosticTime: field " + fieldId
                 + " is not configured as a diagnostic.");
--- a/core/src/ParaGridIO_Xios.cpp
+++ b/core/src/ParaGridIO_Xios.cpp
@@ -52,8 +52,8 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
     }
 
     // Get all variables in the file and load them into a new ModelState
-    for (std::string fieldId : xiosHandler.configGetInputRestartFieldNames()) {
-        ModelArray::Type type = xiosHandler.getFieldType(fieldId);
+    for (const std::string& fieldId : xiosHandler.configGetInputRestartFieldNames()) {
+        const ModelArray::Type& type = xiosHandler.getFieldType(fieldId);
         if (type == ModelArray::Type::H) {
             HField field(ModelArray::Type::H);
             field.resize();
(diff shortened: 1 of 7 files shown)


joewallwork:
Okay I added some commits that address #1000, i.e., make better use of references (to avoid unnecessary copies) and use the `const` qualifier in all the places where arguments and loop variables should be `const`.

core/src/include/Xios.hpp
```diff
@@ -68,44 +60,40 @@ class Xios : public Configured<Xios> {
 
     /* Configuration */
     void configure() override;
-    void configureServer();
 
     /* Calendar, date and duration */
-    void setCalendarStart(const TimePoint start);
-    void setCalendarStep(const int stepNumber);
+    void setCalendarStart(const TimePoint& start);
+    void setCalendarStep(const int& stepNumber);
```
Thanduriel: Went a little far here, `int` should still be passed as values.

core/test/XiosRead_test.cpp
```diff
@@ -93,52 +93,54 @@ MPI_TEST_CASE("TestXiosRead", 2)
     REQUIRE(ModelArray::size(ModelArray::Dimension::DG) == DGCOMP);
 
     // Read restarts from file and check they take the expected values
-    ModelState restarts = grid.getModelState(restartFilename);
+    const ModelState& restarts = grid.getModelState(restartFilename);
     int rank;
     MPI_Comm_rank(test_comm, &rank);
     for (auto& entry : restarts.data) {
```
Thanduriel: You could use a structured binding here:
```for (const auto& [fieldName, modelarray] : restarts.data()) {```

File: core/src/Xios.cpp [hunk 3]
```diff
@@ -270,7 +254,7 @@ void Xios::configureServer()
  * @param isoFormat as bool
  * @return corresponding string representation
  */
-std::string Xios::convertXiosDatetimeToString(const cxios_date datetime, const bool isoFormat)
+std::string Xios::convertXiosDatetimeToString(const cxios_date& datetime, const bool& isoFormat)
```
Thanduriel: pass bool by value

core/src/Xios.cpp
```diff
@@ -293,14 +277,13 @@ std::string Xios::convertXiosDatetimeToString(const cxios_date datetime, const b
  * @param isoFormat as bool
  * @return corresponding XIOS datetime representation
  */
-cxios_date Xios::convertStringToXiosDatetime(const std::string datetimeStr, const bool isoFormat)
+cxios_date Xios::convertStringToXiosDatetime(std::string datetimeStr, const bool& isoFormat)
```
Thanduriel: pass bool by value

core/src/Xios.cpp
```diff
@@ -359,7 +342,7 @@ void Xios::setCalendarStart(const TimePoint start)
  *
  * @param Step number to update to
  */
-void Xios::setCalendarStep(const int stepNumber) { cxios_update_calendar(stepNumber); }
+void Xios::setCalendarStep(const int& stepNumber) { cxios_update_calendar(stepNumber); }
```
Thanduriel: pass int by value

core/src/Xios.cpp
```diff
@@ -561,9 +470,9 @@ void Xios::setupDomains()
     ModelArray::setNComponents(ModelArray::Type::VERTEX, ModelArray::nCoords);
     ModelArray::setNComponents(ModelArray::Type::DG, getAxisSize("DGAxis"));
     ModelArray::setNComponents(ModelArray::Type::DGSTRESS, getAxisSize("DGSAxis"));
-    for (auto entry : domainIds) {
-        ModelArray::Type type = entry.first;
-        const std::string domainId = entry.second;
+    for (auto& entry : domainIds) {
```
Thanduriel: could be done with a structured binding
```for (const auto& [type, domainId] : domainIds) {```

File: core/src/Xios.cpp [hunk 7]
```diff
@@ -693,11 +602,11 @@ void Xios::setupDomains()
 void Xios::setupGrids()
 {
     // Create XIOS grid associated with domain and possibly axis
-    for (auto entry : gridIds) {
+    for (auto& entry : gridIds) {
```
Thanduriel: structured binding

core/src/ParaGridIO_Xios.cpp
```diff
@@ -80,13 +81,9 @@ ModelState ParaGridIO::getModelState(const std::string& filePath)
     }
 
     // Assume that all fields in the supplied ModelState are necessary, and so read them from file.
-    std::set<std::string> restartFieldIds = xiosHandler.configGetInputRestartFieldNames();
+    const std::set<std::string>& restartFieldIds = xiosHandler.configGetInputRestartFieldNames();
```
Thanduriel: This function returns a value so there is no advantage to using a reference.

core/src/ParaGridIO_Xios.cpp
```diff
@@ -111,19 +108,15 @@ ModelState ParaGridIO::readForcingTimeStatic(
...
-    TimePoint xiosTime = xiosHandler.getCurrentDate();
+    const TimePoint& xiosTime = xiosHandler.getCurrentDate();
     if (xiosTime > time) {
         throw std::runtime_error("ParaGridIO::readForcingTimeStatic: requested time point does"
                                  " not align with the calendar and timestep used by XIOS.");
     }
 
     // Get all forcings and load them into a new ModelState
-    std::set<std::string> forcingFieldIds = xiosHandler.configGetForcingFieldNames();
+    const std::set<std::string>& forcingFieldIds = xiosHandler.configGetForcingFieldNames();
```
Thanduriel: better as value `const std::set<std::string>`

core/test/XiosRead_test.cpp
```diff
@@ -93,52 +93,54 @@ MPI_TEST_CASE("TestXiosRead", 2)
     REQUIRE(ModelArray::size(ModelArray::Dimension::DG) == DGCOMP);
 
     // Read restarts from file and check they take the expected values
-    ModelState restarts = grid.getModelState(restartFilename);
+    const ModelState& restarts = grid.getModelState(restartFilename);
```
Thanduriel: `getModelState` returns a value anyway so no need for a reference

core/test/XiosRead_test.cpp
```diff
@@ -147,19 +149,22 @@ MPI_TEST_CASE("TestXiosRead", 2)
...
+    const Duration& timestep = metadata.stepLength();
     // TODO: Avoid making configGetForcingFieldNames public?
     auto forcingFieldNames = xiosHandler.configGetForcingFieldNames();
     for (int ts = 0; ts <= 4; ts++) {
 
         // Read forcings from file and check they take the expected values
-        TimePoint time = xiosHandler.getCurrentDate();
-        ModelState forcings = pio->readForcingTimeStatic(forcingFieldNames, time, forcingFilename);
+        const TimePoint& time = xiosHandler.getCurrentDate();
+        const ModelState& forcings
```
Thanduriel: remove reference

Thanduriel (approved):
Always happy to see more const!

The pass by value/ reference pendulum has swung a bit far in the other direction. For small arguments (e.g. fundamental types bool, int)  it is generally best to pass them by value. Another area where value semantics are preferred is when assigning return values. There is a decent [stack overflow post](https://stackoverflow.com/questions/37649236/why-not-always-assign-return-values-to-const-reference) on the topic.

In a few places you could probably save a few lines by using [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding.html). However, this is just syntactic sugar.

joewallwork:
Thanks @Thanduriel, these are all nice improvements! I've added them (plus a few extra similar ones) in my latest commits.

PR #1001, opened 2025-12-01
Modularize the IceGrowth class.

### LINKED ISSUE #998: Requirements for ICEPACK
One of the near-future goals for nextSIM-DG is integrating the ICEPACK ice thermodynamics module in addition to the current ice thermodynamics, which is derived from Lagrangian neXtSIM. The modularization is at the level currently implemented as the non-modularized `IceGrowth.cpp`. One of the first tasks will be to modularize this part of the code. An analysis of the precise interface requirements of ICEPACK will be required first. The ICEPACK library is written in Fortran, so some kind of C-based shim will be required to call the ICEPACK library.

One of the key properties is that ICEPACK uses multiple ice types, each of which has its own concentration, thickness and snow depth fields. The dynamics will then operate on the overall concentration, thickness and snow depth fields. These fields will be the ones defined in the interface to the module, which the other categories being internal to the ICEPACK module, only being used outside for input and output. The overall fields will also be the only ones shared via the `ModelArrayRef` mechanism.

The output (ParaGridIO) will handle these fields to convert the individual 2d data arrays into multidimensional arrays in the netCDF output files. The XIOS output implementation should also do the same.

Outstanding questions:
* Can the multiple ice categories and thermodynamic level stay entirely within the Fortran part of the ICEPACK library? With only the overall concentration, thickness and snow depth emerging into the C++ part of the model. There would need to be a way of getting off the variables for all the categories into a `ModelState` object for output, as well as to read those fields in on restart.

timspainNERSC:
# Modularize the IceGrowth class.

Fixes #998 (partially)

---
# Change Description

Creates a new module `IceGrowthModule` with an interface class `IIceGrowth` and a single implementation `NSIceGrowth`. This latter is the old `IceGrowth`class adapted to the new modular interface.

---
# Test Description

The tests of the affected classes (`testIceGrowth`->`testNSIceGrowth`, `testIceMinima` and `testPrognosticData`) all pass. Running the model for 10 days produces plausible output.

---
# Documentation Impact

The name of the class in the online help system is changed as necessary.

### DIFF
--- a/core/src/PrognosticData.cpp
+++ b/core/src/PrognosticData.cpp
@@ -30,7 +30,7 @@ PrognosticData::PrognosticData()
     , pAtmBdy(nullptr)
     , pOcnBdy(nullptr)
     , pDynamics(nullptr)
-    , pIceGrowth(nullptr)
+    , pColumnPhysics(nullptr)
 {
     getStore().registerArray(Shared::DAMAGE, &damage, RW);
     getStore().registerArray(Shared::H_ICE_DG, &hice, RW);
@@ -54,8 +54,8 @@ void PrognosticData::configure()
     pDynamics = &Module::getImplementation<IDynamics>();
     tryConfigure(pDynamics);
 
-    pIceGrowth = &Module::getImplementation<IIceGrowth>();
-    tryConfigure(pIceGrowth);
+    pColumnPhysics = &Module::getImplementation<IColumnPhysics>();
+    tryConfigure(pColumnPhysics);
 
     checkAll() = Configured::getConfiguration(keyMap.at(CHECKFIELDS_KEY), checkFieldsDefault);
     checkFast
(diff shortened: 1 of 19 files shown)


joewallwork (approved):
Looks good to me - nice to make things more consistent with other such modules.

It wasn't totally clear what "NS"IceGrowth refers to, although the `module.cfg` seems to suggest it refers to neXtSIM-Lagrangian?

timspainNERSC:
> Looks good to me - nice to make things more consistent with other such modules.
> 
> It wasn't totally clear what "NS"IceGrowth refers to, although the `module.cfg` seems to suggest it refers to neXtSIM-Lagrangian?

This comment made me remember that I'd been disliking the `IceGrowth` name for a while now. And there's not going to be a better opportunity to change it.

Issue #967, opened 2025-10-24
Create a friend XIOS unit test class


joewallwork:
This would allow us to avoid the `unittest` hack and also to make many of the member functions private.

joewallwork:
This issue is made redundant by #1002.

PR #1002, opened 2025-12-05
File behaviour for restarts and diagnostics (take 2)

### LINKED ISSUE #898: XIOS: Account for filename patterns
Filename patterns are not currently supported in our XIOS implementation. This would likely make use of `setFileSplitFreq` and `getFileSplitFreq`.

_Originally posted by @jwallwork23 in https://github.com/nextsimhub/nextsimdg/pull/846#discussion_r2204889989_

### LINKED ISSUE #933: XIOS doesn't write first timestep
On closer examination of the files produced by the XIOS I/O implementation, I realised that the first timestep never gets written. I did quite a bit of debugging but couldn't get it to happen. I need to go through the documentation to check if this is expected behaviour.

### LINKED ISSUE #935: XIOS: Mask should be treated separately
Once #934 and #910 have been merged, we should account for special treatment of the landmask (and possibly other fields?).

The landmask is fixed and so should only be written once. This can be achieved using the XIOS 'once' field operation setting.

Moreover, it should be a diagnostic output rather than a restart.

### LINKED ISSUE #977: Diagnostic files should default to averaging in XIOS
This can be fixed in a future PR, but the non-XIOS diagnostic files now default to averaging.

_Originally posted by @timspainNERSC in https://github.com/nextsimhub/nextsimdg/pull/949#discussion_r2493689959_

### LINKED ISSUE #997: XIOS: restart files shouldn't include time dimension
Currently, the restart files written by XIOS should just be instantaneous outputs for restarting from. That is, the time dimension shouldn't be included and only a single timestep's worth of data should be in there.

### LINKED ISSUE #1003: XIOS: Case of only one field type
Unfortunately, the way that we now label domains breaks down when there's only one field type. If there's only one field type then there's only one domain, meaning XIOS doesn't append the dimension names `x` and `y` with `_<domainId>`. We need to find either a workaround in case someone only makes use of one field type, or think of a better way to set this up.

joewallwork:
# File behaviour for restarts and diagnostics (take 2)

Fixes #935
Fixes #977
Fixes #997
Fixes #898
Fixes #933
Fixes #1003
Makes #967 redundant

### Task List

---
# Change Description

This PR fixes several things related to how restarts and diagnostics are configured with XIOS:
1. Diagnostics are created by averaging all timesteps included in an output file dump.
2. Restart file output frequency and split frequency are aligned so that each output file dump only includes a single timestep's worth of data.
3. Restart files are only read once.
4. Handle file patterns in filenames.
5. Ensure XIOS writes the first timestep.
6. If only one domain type is used then the output file will use `x` and `y` dimensions, rather than the versions appended with `_<dimName>`. We account for this by assuming that `x` and `y` imply `x_dim` and `y_dim` if encountered in an input file.

Whilst file patterns are made consistent with what's used elsewhere in nextSIM-DG, the output filenames show the range of datetimes separated by a hyphen.

---
# Test Description

Both read and write tests are separated out for reading of different file types. This is what accounts for most of the lines changed.

Tests are updated to account for the new behaviour.

I decided to drop `XiosFile_test` because it is no longer needed. This meant that we could drop the annoying `unittest` special case.

---
# Documentation Impact

The XIOS documentation page has been updated accordingly.

---
### Pre-Request Checklist


joewallwork:
Closing as superseded by #1010 and #1011.

PR #1005, opened 2025-12-16
Halo add corners

TomMelt:
# Add corner neighbours into halo exchange logic

### Task List

---

# Change Description

This PR has the following key changes:

- adds support for corner neighbours to the existing halo exchange logic (which previously just exchanged edges)
- add support for `CGVector` and `DGVectorHolder` exchange (this is required for corners and edges)
- updates metadata tests and closed-boundary (CB) halo test to work with corner neighbours
- removes the Periodic-Boundary (PB) halo exchange test for now, this will be addressed in a subsequent PR #1044

---
# Test Description

### `HaloExchangePB_test.cpp`

`HaloExchangeCB_test.cpp` creates test data for all the array types:
- `HField`
- `VertexField`
- `DGField`
- `DGVector`
- `CGVector`

I modified the previous version, so that the model data are created in such a way that the value of each cell can be calculated based on its indices. Therefore we can check after exchange that all of the cells have been exchanged successfully. This also now allows us to change the number of ranks arbitrarily (assuming you also provide the appropriate `partition_metadata` file. 

### `ModelMetadataCB_test.cpp` and `ModelMetadataPB_test.cpp`

Small modifications have been made to the periodic and closed boundary `ModelMetadata` tests.

The changes now add the additional corner neighbour metadata and check that it is ready correctly.

### `partition_metadata_3_cb.cdl` and `partition_metadata_3_pb.cdl`

These files are now generated at compile time by running the `decomp` tool on a new file `halo_test_mask.cdl`. This is in line with similar changes introduced by @joewallwork on the XIOS tests.

---
# Other Details

* `domain_decop` git commit has now been bumped to the latest version on main. This version of the decomp tool contains the corner neighbour metadata.

### Further work

- Integrating halo exchange to nextsim is handled by PR #924


core/src/include/Halo.hpp
```diff
@@ -52,11 +54,36 @@ class Halo {
 
     /*!
      * @brief Constructs a halo object from DGVector
```
joewallwork:
```suggestion
     * @brief Constructs a halo object from DGVectorHolder
```

core/src/include/Halo.hpp
```diff
@@ -52,11 +54,36 @@ class Halo {
...
+     * @param cgv CGVector object to create halo from
+     */
+    template <int N> Halo(CGVector<N>& cgv)
+    {
+        m_numComps = 1;
+        isCG = true;
+        CGdegree = N;
+        nCells = CGdegree;
         setSpatialDims();
         intializeHaloMetadata();
```
joewallwork: In none of these cases do the constructors depend on their argument, right? The only information that's being used is the value of `N` and the type. So do we actually need to provide `dgv`, `dgvh`, or `cgv` as arguments?

core/src/include/Halo.hpp
joewallwork, line 1: There seems to be quite a lot of implementation in this header file. Might it be better to separate it out into `core/src/Halo.cpp`?

core/src/include/Halo.hpp
```diff
@@ -234,35 +258,164 @@ class Halo {
...
-                ModelArraySlice::copySliceWithIters(source, sourceIter, tempBuffer, tempBufferIter);
+        size_t buffer_len = sendBufferSize;
+        if (isCG) {
+            buffer_len = sendBufferSize / CGdegree;
+        }
+        for (size_t comp = 0; comp < m_numComps; ++comp) {
+            Eigen::Map<Eigen::ArrayXXd, 0, Eigen::InnerStride<Eigen::Dynamic>> source_map(
+                source.col(comp).data(), m_Nx, m_Ny, Eigen::InnerStride<>(m_numComps));
+            Eigen::Map<Eigen::ArrayXXd> buffer_map(send[comp].data(), buffer_len, nCells);
+            for (auto edge : edges) {
```
joewallwork, line 218: This code could benefit from some comments as I'm struggling to understand what it does. numComps is the `N` value in the template constructor, right?

core/src/include/Halo.hpp
```diff
@@ -234,35 +258,164 @@ class Halo {
...
-            MapType map(send[i].data(), m_numHaloCells, 1);
-            // map is connected with the send buffer so the following line sets the data in send
-            map = tempBuffer.col(i);
+    }
+
+    /**
+     * @brief Calculate which edge the send position sendPos would fall under in the send buffer
+     *
+     */
+    Edge edgeFromSendPos(int sendPos, int fromRank)
```
joewallwork: Missing `@params` etc in docstring

core/src/include/Halo.hpp
```diff
@@ -234,35 +258,164 @@ class Halo {
...
+            fromRank = metadata.cornerRanks[corner][0];
+            disp = metadata.cornerHaloSend[corner][0];
+            recvOffset = metadata.cornerHaloRecv[corner][0];
+        }
+        auto sendEdge = edgeFromSendPos(disp, fromRank);
+        if (isVertex) {
+            count = 1;
+            disp = disp + sendEdge;
+            recvOffset = recvOffset + 4;
+            // this is messy
```
joewallwork: Is this a TODO to do it differently?

core/src/include/Halo.hpp
```diff
@@ -374,9 +516,235 @@ class Halo {
...
+
+            if (metadata.cornerRanks[corner].size() > 0) {
+                // non-periodic case
+                hasCorner = true;
+                isPeriodic = false;
+            } else if (metadata.cornerRanksPeriodic[corner].size() > 0) {
+                // periodic case
+                hasCorner = true;
+                isPeriodic = true;
+            }
```
joewallwork: Presumably it's impossible for both conditions to be true at the same time? Or would it be useful to include a failure mode in case it happens due to a bug?

core/src/include/Halo.hpp
```diff
@@ -374,9 +516,235 @@ class Halo {
...
+                        Eigen::Map<Eigen::ArrayXXd> rmap(recv[comp].data(), buffer_len, nCells);
+                        auto offset = buffer_len - (4 - corner) * nCells;
+                        rmap.block(offset, 0, nCells, nCells).transposeInPlace();
+                    }
+                }
+            }
+        }
+    }
+
+    template <typename T> void populateTarget(T& target)
```
joewallwork: Please add a docstring.

core/src/ModelMetadata.cpp
```diff
@@ -148,6 +195,24 @@ void ModelMetadata::getPartitionMetadata(std::string partitionFile)
 
     readNeighbourData(ncFile);
 
+    // cornerHaloRecv doesn't need to be read because it can be easily calculated.
```
joewallwork, line 198: I guess there's an assumption (probably right) that a small amount of computation is cheaper than a small amount of I/O?

core/test/HaloExchangeCB_test.cpp
joewallwork, line 1: Nice to reduce the amount of numbers hard-coded in code here, well done!

joewallwork (commented):
Thanks for this @TomMelt. Here's some initial feedback. I haven't checked all of the logic but let me know if there's anything in particular you want checking.

core/src/include/ModelMetadata.hpp
```diff
@@ -206,6 +222,14 @@ class ModelMetadata {
     neighbourArray neighbourExtentsPeriodic;
     neighbourArray neighbourHaloSendPeriodic;
     neighbourArray neighbourHaloRecvPeriodic;
+
+    typedef std::array<std::vector<int>, N_CORNER> cornerArray;
```
Thanduriel: Types should be in CamelCase.
```
using CornerArray = std::array<std::vector<int>, N_CORNER>;
```

core/src/include/ModelMetadata.hpp
```diff
@@ -191,12 +191,28 @@ class ModelMetadata {
...
     static constexpr std::array<Edge, N_EDGE> edges = { BOTTOM, RIGHT, TOP, LEFT };
     std::array<std::string, N_EDGE> edgeNames = { "bottom", "right", "top", "left" };
 
+    enum Corner { TOP_LEFT, TOP_RIGHT, BOTTOM_RIGHT, BOTTOM_LEFT, N_CORNER };
+    static constexpr std::array<Corner, N_CORNER> corners
+        = { TOP_LEFT, TOP_RIGHT, BOTTOM_RIGHT, BOTTOM_LEFT };
+    std::array<std::string, N_CORNER> cornerNames
+        = { "top_left", "top_right", "bottom_right", "bottom_left" };
+
     typedef std::array<std::vector<int>, N_EDGE> neighbourArray;
```
Thanduriel: Types should be in CamelCase. Prefer `using` over `typedef` (see [CppCoreGuidlines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t43-prefer-using-over-typedef-for-defining-aliases)).
```
using NeighbourArray = std::array<std::vector<int>, N_EDGE>;
```

core/src/include/ModelMetadata.hpp
```diff
@@ -191,12 +191,28 @@ class ModelMetadata {
      * @return The total number of grid points in Y for the global domain.
      */
     int getGlobalExtentY() const;
+    /*!
+     * @brief Gets the extents of the grid in the X direction for all ranks.
+     * @return A vector containing the number of grid points in X for each rank.
+     */
+    std::vector<int> getRankExtentsX() const;
```
Thanduriel: Return by `const&` to prevent unnecessary copies.

core/src/include/ModelMetadata.hpp
```diff
@@ -191,12 +191,28 @@ class ModelMetadata {
...
+    /*!
+     * @brief Gets the extents of the grid in the Y direction for all ranks.
+     * @return A vector containing the number of grid points in Y for each rank.
+     */
+    std::vector<int> getRankExtentsY() const;
```
Thanduriel: Return by `const&` to prevent unnecessary copies.

core/src/include/Halo.hpp
```diff
@@ -374,9 +516,235 @@ class Halo {
...
+     * @brief Transpose corners received from vertical edges
+     *
+     * If the sending edge was from a vertical side (e.g., LEFT or RIGHT) then it has been
+     * transposed when flattened into the 1D send buffer. We need to account for that.
+     */
+    void transposeCorners()
+    {
+        auto& metadata = ModelMetadata::getInstance();
+        int fromRank;
+        size_t disp;
```
Thanduriel: These variables could be defined closer to where they are used, i.e. in the `if (hasCorner)` branch.

core/src/include/Halo.hpp
Thanduriel: Agreed. Function definitions should go in a source file if possible. Exceptions are (public) template functions which have to be defined in the header and maybe one-liners.

core/src/include/Halo.hpp
```diff
@@ -374,9 +516,235 @@ class Halo {
...
+                    disp = metadata.cornerHaloSend[corner][0];
+                }
+
+                auto sendEdge = edgeFromSendPos(disp, fromRank);
+
+                if (isVertical(sendEdge)) {
+                    auto buffer_len = recvBufferSize / CGdegree;
+                    for (size_t comp = 0; comp < m_numComps; ++comp) {
+                        Eigen::Map<Eigen::ArrayXXd> rmap(recv[comp].data(), buffer_len, nCells);
+                        auto offset = buffer_len - (4 - corner) * nCells;
```
Thanduriel: Where does this 4 come from? Is it just `Corner::N_CORNER`?

core/src/ModelMetadata.cpp
```diff
@@ -157,6 +222,8 @@ int ModelMetadata::getLocalExtentX() const { return localExtentX; }
 int ModelMetadata::getLocalExtentY() const { return localExtentY; }
 int ModelMetadata::getGlobalExtentX() const { return globalExtentX; }
 int ModelMetadata::getGlobalExtentY() const { return globalExtentY; }
+std::vector<int> ModelMetadata::getRankExtentsX() const { return rankExtentsX; }
+std::vector<int> ModelMetadata::getRankExtentsY() const { return rankExtentsY; }
```
Thanduriel:
```suggestion
const std::vector<int>& ModelMetadata::getRankExtentsX() const { return rankExtentsX; }
const std::vector<int>& ModelMetadata::getRankExtentsY() const { return rankExtentsY; }
```

core/test/ModelMetadataPB_test.cpp
```diff
@@ -23,13 +23,19 @@ constexpr ModelMetadata::Edge RIGHT = ModelMetadata::Edge::RIGHT;
 constexpr ModelMetadata::Edge TOP = ModelMetadata::Edge::TOP;
 constexpr ModelMetadata::Edge LEFT = ModelMetadata::Edge::LEFT;
 
+constexpr ModelMetadata::Corner TOP_LEFT = ModelMetadata::Corner::TOP_LEFT;
+constexpr ModelMetadata::Corner TOP_RIGHT = ModelMetadata::Corner::TOP_RIGHT;
+constexpr ModelMetadata::Corner BOTTOM_RIGHT = ModelMetadata::Corner::BOTTOM_RIGHT;
+constexpr ModelMetadata::Corner BOTTOM_LEFT = ModelMetadata::Corner::BOTTOM_LEFT;
```
Thanduriel: If the intend here is just to introduce shorter names, `using` would be more clear:
```
using enum ModelMetadata::Corner;
```

core/src/include/Halo.hpp
```diff
@@ -234,35 +258,164 @@ class Halo {
...
+     * @brief Calculate which edge the send position sendPos would fall under in the send buffer
+     *
+     */
+    Edge edgeFromSendPos(int sendPos, int fromRank)
+    {
+        auto& metadata = ModelMetadata::getInstance();
+
+        // extents of sending domain
+        auto extentX = metadata.getRankExtentsX()[fromRank];
+        auto extentY = metadata.getRankExtentsY()[fromRank];
```
Thanduriel: I think it would be better to not use `auto` if the expression has a simple type. Here the type is just `int` while in many other places the standard size type `size_t` is used instead. The distinction between signed and unsigned ints is important because the following conditions would not work with an unsigned type.

core/src/include/Halo.hpp
```diff
@@ -234,35 +258,164 @@ class Halo {
...
+            // recvOffset is the offset in the recv buffer and this belongs to the current rank
+            recvOffset = recvOffset + recvBufferSize / nCells * cell;
+
+            // disp is the offset in the "sending" buffer which belongs to rank "fromRank"
+            // Therefore we need to compute how many halo cells that rank has to work out the offset
+            // for each cell
+            auto extentX = CGdegree * metadata.getRankExtentsX()[fromRank] + 1;
+            auto extentY = CGdegree * metadata.getRankExtentsY()[fromRank] + 1;
+            // TODO this is too big. Should only be 2 x CGdegree * haloWidth (extentX + extentY)
+            auto fromRankSendBufferSize = 2 * haloWidth * CGdegree * (extentX + extentY);
```
Thanduriel: Is this todo outdated? The line below looks just like suggested.

Thanduriel (changes requested):
Here is my feedback.

core/src/include/Halo.hpp
TomMelt: I think we do, because we need to templated `N` from  `dgv` and `dgvh`.

You are right that `Halo(DGVectorHolder<N>& dgvh)` and `Halo(DGVector<N>& dgv)` are essentially the same. But `Halo(ModelArray& ma)` and `Halo(CGVector<N>& cgv)` set different member variables from `dgv`.

What I could do however, is change `Halo(DGVectorHolder<N>& dgvh)` to call `Halo(DGVector<N>& dgv)` with `static_cast<DGVector<DGadvection>&>()` like we do in dynamics:

https://github.com/nextsimhub/nextsimdg/blob/87dadeb5c30e5156015e06761ce3635961f995c9/dynamics/src/include/BrittleCGDynamicsKernel.hpp#L131-L134

core/src/include/Halo.hpp
TomMelt: Thanks done in f1829af1

core/src/include/Halo.hpp
TomMelt: I have simplified this in f1829af1

core/src/include/Halo.hpp
TomMelt: Good point. I have added explanation here 0436038d

Please let me know if this is clearer.

core/src/include/Halo.hpp
TomMelt: Thanks see eb26a814

core/src/include/Halo.hpp
TomMelt: Nope, this is the best way I could think to do it. But I have updated the comment to be more helpful

a8c47722

core/src/include/Halo.hpp
TomMelt: Sure, that's a good point. It shouldn't be possible. So I will add a check in case.

see c030bf38

core/src/include/Halo.hpp
TomMelt: Thanks for spotting. Done in c76f111e

core/src/ModelMetadata.cpp
TomMelt: That's true, but the main reason is it would require a lot of changes in `domain_decomp` to essentially add a fixed constant which is essentially 1 (i.e., the `halo_size`). We would need to write routines to save that metadata in `domain_decomp` and read it in `nextsim`. I thought it wasn't worth the effort :sweat_smile:

core/test/HaloExchangeCB_test.cpp
TomMelt: Thank you :)

core/src/include/ModelMetadata.hpp
TomMelt: Thanks this and CornerArray have been done in a59341f4

core/src/include/ModelMetadata.hpp
TomMelt: Thanks `getRankExtentsX()` and `getRankExtentsY()` have been modified in c4210bbc

core/src/include/Halo.hpp
TomMelt: Thanks see bcfedc7d

core/src/include/Halo.hpp
TomMelt: Yes, I have replaced this and a few other magic numbers with the enums b6938692

core/src/ModelMetadata.cpp
TomMelt: Thanks this was fixed in c4210bbc

core/test/ModelMetadataPB_test.cpp
TomMelt: Thanks that was my intent. It appears that your suggested syntax is only available in `-std=c++20`, see [here](https://en.cppreference.com/w/cpp/language/enum.html).

Are you happy for me to keep it? or do you have any other suggestions?

core/src/include/Halo.hpp
TomMelt: Thanks that's a very good point. And a footgun I shot myself with several times. Changed to `int` here  23e3ca9c

core/src/include/Halo.hpp
TomMelt: Yep thanks for spotting it. It was a reminder that is no longer valid. Removed here eeeecf07

core/src/include/Halo.hpp
TomMelt: Thanks @joewallwork and @Thanduriel . I have split the implementation as suggested and I also tidied up a bit.

- Split implementation out from `Halo.hpp` into `core/src/Halo.cpp`
- Moved implementation of non-templated functions to `core/src/Halo.cpp`
- Made the following functions private:
  - `setSpatialDims()`
  - `intializeHaloMetadata()`
  - `sendBufferPositions()`
  - `sourceToSendBuffer()`
  - `recvBufferPositions()`
  - `recvBufferToTarget()`
  - `recvBufferToTarget()`
  - `transposeCorners()`
  - `populateTarget()`

See commit c2136348

core/test/ModelMetadataPB_test.cpp
Thanduriel: I guess the pre C++20 alternative would be to introduce each name separately which is still a bit more compact than the current version:
```
using ModelMetadata::Corner::TOP_LEFT;
```
But I am fine with the current version.

Thanduriel (approved):
I have one alternative suggestion for the shorter enum types but you decide if you want to make the change. From my side the code is ready for merge.

core/test/ModelMetadataPB_test.cpp
TomMelt: It looks like `using ModelMetadata::Corner::TOP_LEFT;` is also only available in `-std=c++20` but I have done the following which is slightly better:

```suggestion
using Corner = ModelMetadata::Corner;

constexpr auto TOP_LEFT = Corner::TOP_LEFT;
constexpr auto TOP_RIGHT = Corner::TOP_RIGHT;
constexpr auto BOTTOM_RIGHT = Corner::BOTTOM_RIGHT;
constexpr auto BOTTOM_LEFT = Corner::BOTTOM_LEFT;
```

core/src/include/Halo.hpp
joewallwork: This is much clearer, thanks!

core/src/include/Halo.hpp
```diff
@@ -402,47 +465,14 @@ class Halo {
...
-     * │ │x│x│ │
-     * ├─┼─┼─┼─┤
-     * │x│o│o│x│
-     * ├─┼─┼─┼─┤
-     * │x│o│o│x│       o = original data
-     * ├─┼─┼─┼─┤       x = mpi halo data (from recv)
-     * │ │x│x│ │ (empty) = unused data in DGVector
-     * └─┴─┴─┴─┘
-     */
     template <typename T> void exchangeHalos(T& target)
```
joewallwork, line 479: Please could you add a docstring for this function?

core/src/Halo.cpp
```diff
@@ -0,0 +1,390 @@
...
+ * @brief Halo exchange class implementation
+ * @details
+ *
+ * Halo exchange class
+ *
+ * All functionality for halo exchange between MPI ranks is contained in this class.
+ *
+ * Halo supports the main data structures of NextSim e.g., ModelArray, DGVector and CGVector.
+ *
+ * The halos are exchange via one-sided MPI communication using RMA.
```
joewallwork, line 12: Perhaps spell out RMA here for clarity.

core/src/Halo.cpp
```diff
@@ -0,0 +1,390 @@
...
+        if ((sendEdge == Edge::LEFT or sendEdge == Edge::RIGHT)
+            and (corner == Corner::TOP_RIGHT or corner == Corner::TOP_LEFT)) {
+            disp = disp + 1;
+        }
+    }
+}
+
+void Halo::populateRecvBuffers()
+{
+
```
joewallwork:
```suggestion
```

joewallwork (approved):
This is great, thanks @TomMelt. I have a few small requests but I think it's otherwise good to go!

core/src/Halo.cpp
TomMelt: Thanks changed in 876dea49

core/src/Halo.cpp
TomMelt: Thanks changed in 876dea49

core/src/include/Halo.hpp
TomMelt: Thanks changed in 876dea49

PR #1010, opened 2026-01-08
XIOS: Separate out testing of restarts and diagnostics

### LINKED ISSUE #977: Diagnostic files should default to averaging in XIOS

### LINKED ISSUE #1003: XIOS: Case of only one field type

joewallwork:
# XIOS: Separate out testing of restarts and diagnostics

Fixes #977
Fixes #1003
Refinement of #1002

### Task List

---
# Test Description

This PR separates the XIOS reading and writing tests out by type. In addition, we add a read test for the diagnostic outputs and check that they are time-averaged.

---
# Change Description

Code changes are made to account for the case where only one XIOS domain is used when writing. This is required due to an unfortunate consequence of the way XIOS appends the domain name only if there are two or more domains.

In detail: if only one domain type is used then the output file will use `x` and `y` dimensions, rather than the versions appended with `_<dimName>`. We account for this by assuming that `x` and `y` imply `x_dim` and `y_dim` if encountered in an input file.

---
# Documentation Impact

n/a

---
### Pre-Request Checklist

### DIFF
--- a/core/src/ModelMetadata.cpp
+++ b/core/src/ModelMetadata.cpp
@@ -196,6 +196,21 @@ void ModelMetadata::setDimensionsFromFile(const std::string& filename)
             if (dim.isNull()) {
                 dim = ncFile.getDim(dimensionSpec.altName);
             }
+#ifdef USE_XIOS
+            // Account for the fact that XIOS writes dimensions differently if only one
+            // discretisation is written out. If the dimension is still null at this point then we
+            // assume that the only discretisation used is HDomain-based, i.e., HField, DGField, or
+            // DGSField.
+            if (dim.isNull()) {
+                if (dimensionSpec.name == "x_dim") {
+                    dim = ncFile.getDim("x");
+                } else if (dimensionSpec.name == "y_dim") {
+                    dim = ncFile.getDim("y");
+                } else {
+                    continue;
+                }
+            }
+#endif
             // If we didn't find a dimension with the dimensions name or altName, throw.
             if (dim.isNull()) {
                 throw std::out_of_range(
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -794,11 +794,16 @@ void Xios::createField(const std::string& fieldId)
         throw std::runtime_error("Xios: Failed to create field '" + fieldId + "'");
     }
 
-    // Restarts are read "once", everything else is written "instant"ly (no averaging)
+    // Set the operation type
     std::string operation;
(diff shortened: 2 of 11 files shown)


core/src/ModelMetadata.cpp
```diff
@@ -196,6 +196,21 @@ void ModelMetadata::setDimensionsFromFile(const std::string& filename)
...
+            if (dim.isNull()) {
+                if (dimensionSpec.name == "x_dim") {
+                    dim = ncFile.getDim("x");
+                } else if (dimensionSpec.name == "y_dim") {
+                    dim = ncFile.getDim("y");
+                } else {
+                    continue;
+                }
+            }
+#endif
```
joewallwork, line 213: This is fixed by #1015, which ensures that XIOS never writes out files with `x` or `y` as dimensions.

core/src/Xios.cpp
```diff
@@ -957,10 +962,13 @@ void Xios::setupFields()
 
         // Create map for field types
         const std::map<std::string, ModelArray::Type> dimensionKeys = {
+            { "yx", ModelArray::Type::H },
             { "ydimxdim", ModelArray::Type::H },
             { "y_dimx_dim", ModelArray::Type::H },
+            { "yxdg_comp", ModelArray::Type::DG },
             { "ydimxdimdg_comp", ModelArray::Type::DG },
             { "y_dimx_dimdg_comp", ModelArray::Type::DG },
+            { "yxdgstress_comp", ModelArray::Type::DGSTRESS },
```
joewallwork, line 971: These changes will also be reverted in #1015.

timspainNERSC (approved):
As usual with the XIOS tests, I haven't done more than a check that the they look reasonable and logical. Everything else looks good.

joewallwork:
> As usual with the XIOS tests, I haven't done more than a check that the they look reasonable and logical. Everything else looks good.

Thanks @timspainNERSC. I'm going to wait for the current CI jobs on `develop` to finish before merging because the Docker one *should* :crossed_fingers: update the Docker image following #1024.

PR #1011, opened 2026-01-08
XIOS: properly handle filename patterns

### LINKED ISSUE #898: XIOS: Account for filename patterns

### LINKED ISSUE #933: XIOS doesn't write first timestep

### LINKED ISSUE #935: XIOS: Mask should be treated separately

### LINKED ISSUE #997: XIOS: restart files shouldn't include time dimension

joewallwork:
# XIOS: properly handle filename patterns

Fixes #935
Fixes #997
Fixes #898
Fixes #933
Makes #967 redundant
Refinement of #1002
~~Merges into #1010~~
Merges into #1015

### Task List

---
# Change Description

This PR fixes several things related to how restarts and diagnostics are configured with XIOS:
1. Properly handle file patterns in filenames.
2. Restart file output frequency and split frequency are aligned so that each output file dump only includes a single timestep's worth of data.
3. Restart files are only read once.
4. Ensure XIOS writes the first timestep.

Whilst file patterns are made consistent with what's used elsewhere in nextSIM-DG, the output filenames show the range of datetimes separated by a hyphen.

---
# Test Description

Tests are updated to account for the new behaviour.

I decided to drop `XiosFile_test` because it is no longer needed. This meant that we could drop the annoying `unittest` special case. This is what gives the large number of lines removed.

---
# Documentation Impact

The XIOS documentation page has been updated accordingly.

---
### Pre-Request Checklist


joewallwork:
Changed the base of this PR so that it targets #1015 rather than #1010.

core/src/Xios.cpp
```diff
@@ -73,8 +71,8 @@ Xios::HelpMap& Xios::getHelpText(HelpMap& map, bool getAll)
             "-Dxios_DIR=/path/to/xios." },
     };
     map["XiosOutput"] = {
-        { keyMap.at(OUTPUT_FIELD_NAMES_KEY), ConfigType::STRING, {}, "", "",
-            "Comma-separated list of field names to be written to the output file." },
+        { keyMap.at(OUTPUT_FIELD_NAMES_KEY), ConfigType::STRING, {}, "restart%Y-%m-%dT%H:%M:%SZ.nc",
```
TomMelt: Is this string `"restart%Y-%m-%dT%H:%M:%SZ.nc"` meant to be here? Is this mean for output filename? or should this config option be renamed?

TomMelt (changes requested):
Thanks @joewallwork looks good to me. I just had one question on the config help text. I am confused whether the name is wrong or the description.

core/src/Xios.cpp
joewallwork: Oh good spot! This was meant to be used for the default restart file name for consistency with the rest of the model... but now I just pick up the value already extracted elsewhere in the model. So it was a default put in the wrong place that's no longer needed. Fixed in 973ab557.

TomMelt (approved):
Thanks @joewallwork happy with your latest commit. It is ready to go 🚂

PR #1015, opened 2026-01-13
Post-process dims in XIOS output files

joewallwork:
Merges into #1010.

Currently in https://github.com/nextsimhub/nextsimdg/pull/1010 we have the following issue with writing fields with XIOS:
* If there are multiple fields that are (between them) based on two or more XIOS domains then the dimension names get appended with those domain names. By naming the domains `dim`, `vertex`, and `cg`, this gives rise to dimensions `x_dim`, `y_dim`, `x_vertex`, `y_vertex`, `x_cg`, and `y_cg`.
* If the fields being written out are only based on one XIOS domain then the appendment doesn't happen, meaning we just have `x` and `y`.
  
The workaround in that PR handles dimensions `x` or `y` in a file being read by assuming they correspond to `x_dim` and `y_dim`. Ideally, we would avoid having `x` or `y` alone in any files and instead make a post-processing of the output file after it is written with XIOS.

That's achieved in this PR via a post-processing step. Upon finalising the XIOS context, we:
1. Count how many different XIOS domains were used for writing out. If there was exactly one then we proceed to tackle the issue.
2. Loop overall output files that might have been written by XIOS and check they exist.
3. For any such files that exist, rename the dimensions and variables as appropriate and write out.

To achieve this, I had to implement subtraction of a `Duration` from a `TimePoint` because of how XIOS presents the time window suffices.

### DIFF
--- a/core/src/ModelMetadata.cpp
+++ b/core/src/ModelMetadata.cpp
@@ -171,6 +171,10 @@ void ModelMetadata::setDimensionsFromFile(const std::string& filename)
         throw std::runtime_error(
             "ModelMetadata :: setDimensionsFromFile() called without input file.");
     }
+
+    // Set to record the names of dimensions found in the file
+    std::set<std::string> dimNames;
+
     try {
 #ifdef USE_MPI
         auto& modelMPI = ModelMPI::getInstance();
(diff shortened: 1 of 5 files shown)


core/src/ModelMetadata.cpp
joewallwork, line 1: With XIOS, if dimensions are unused then they aren't written out into the file. As such, if only `HField`s get written out (as in `XiosWriteDiagnostic_test`) then only `x_dim` and `y_dim` will be in the file, even if `x_vertex` etc. have been defined. As such, I converted an error to a warning here to allow for only a subset of the dimensions being defined in an input file.  If no dimensions at all are found then an error gets thrown.

timspainNERSC (approved):

joewallwork:
Will merge following #1010.

PR #1016, opened 2026-01-13
Test XIOS reading and writing with different ranks

### LINKED ISSUE #1004: XIOS: Test reading and writing with different numbers of ranks
Following the separation out of reading and writing tests in #1002, it should be fairly straightforward to test this.

joewallwork:
Fixes #1004.
~~Merges into #1011.~~

This PR modifies the diagnostic reading test to use a different number of MPI ranks from the corresponding write test (3 vs. 2) to check this is possible.

We aren't currently able to do the same for the restart tests because the information they encode depends on the rank.


TomMelt (approved):
Thanks @joewallwork this is good to go 🚀

PR #1017, opened 2026-01-14
XIOS: allow reading and writing in the same run

### LINKED ISSUE #1007: XIOS: Allow a field to be both read and written
Currently, a field can only be configured for reading or writing, not both. To progress onto realistic cases, we need to refactor the approach to allow this.

Required by #982.

joewallwork:
# XIOS: allow reading and writing in the same run

Fixes #1007
~~Merges into #1011~~

### Task List

---
# Change Description

In order to allow the reading and writing of a field in the same run, it turned out that I needed to create an additional field for the reading and to associate the two together with a field reference. They point to the same data but it allows the field for reading to have different I/O properties, e.g., `"instant"`, `"averaging"`, `"once"`. It does not appear to be required to create an additional field for writing - you can just use the 'base' field.

To facilitate the creation of the inherited field, this PR adds a `createInheritedField` member function. Helper functions `getFileIOType` and `setFieldOperation` are also introduced.

---
# Test Description

To demonstrate that the functionality does indeed work, I've introduced an additional test `XiosReadWriteRestarts_test`. It reads the `xios_test_input.nc` file in as a restart in `ModelState` format, writes it back out as a restart dump, and then checks the corresponding files exist.

I also removed `XiosField_test` as it's no longer needed.

---
# Documentation Impact

A developer note has been added in the XIOS doc page.

---
### Pre-Request Checklist


joewallwork:
How bizarre. The tests all pass on my machine (even after removing the build directory and rebuilding). I'll debug this next week.

joewallwork:
The issue has been fixed by updating the Docker image (in particular the XIOS version) so this is now ready for review.

core/src/Xios.cpp
```diff
@@ -150,6 +150,23 @@ Xios::~Xios() { finalize(); }
...
     if (isEnabled) {
+
+        // Ensure that base fields have operation type 'instant' if not already defined
+        std::set<std::string> fieldIds = configGetInputRestartFieldNames();
+        std::set<std::string> forcingFieldIds = configGetForcingFieldNames();
+        fieldIds.insert(forcingFieldIds.begin(), forcingFieldIds.end());
+        for (const std::string& fieldId : fieldIds) {
+            xios::CField* field = getField(fieldId);
+            if (cxios_is_defined_field_operation(field)) {
+                Logged::warning("Xios: Overwriting operation for field '" + fieldId + "'");
```
TomMelt: Not a problem, but I am just curious. In what scenarios would we be overwriting? and do you think it would cause any issues?

TomMelt (approved):
Just a small question. But otherwise good to go 🚀

core/src/Xios.cpp
joewallwork: Good point, this is actually redundant. In edf8d577 I changed the logic so that it only sets the field operation if it hasn't already been set, which is consistent with the comment above.

PR #1024, opened 2026-01-20
Fix logic for pushing Docker build to ghcr

joewallwork:
Again, the Docker dev-env image https://github.com/nextsimhub/nextsimdg/pkgs/container/nextsimdg-dev-env wasn't updated by recent CI jobs on `develop`. It appears to me that this is because the CI is only configured to push to gchr for runs on `main`.

This PR changes the logic to address this and also allows scheduled runs to update the image.


TomMelt (approved):
Thanks for fixing this

PR #1025, opened 2026-01-27
Only trigger Docker workflow if related files change

joewallwork:
Currently, the Docker build workflow will trigger when *any* PR is merged into `main` or `develop`. This is excessive because in most cases those PRs don't touch the relevant files. This PR configures the CI workflow triggers to avoid this behaviour.


TomMelt (approved):
Thanks for this @joewallwork

PR #1023, opened 2026-01-20
Test `stateFromFile` with XIOS

joewallwork:
# Test `stateFromFile` with XIOS

Fixes #1022
Merges into #1017 

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

Added some preprocessor conditionals in `StructureFactory::stateFromFile` to close the XIOS context before calling `getModelState`, as this is required.

Added logic to track the XIOS context status so that we can call `close_context_definition` multiple times and it will only do something the first time.

---
# Test Description

Updated all XIOS tests that call `ParametricGrid::getModelState` to instead call `StructureFactory::stateFromFile`. This is done because it's how restarts are actually read in the model. It's worth noting that the read-write test does this and then creates a separate `ParametricGrid` to handle file writing with no issue.

---
# Documentation Impact

The XIOS doc page has been updated to reflect this change.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] This change conforms to the conventions described in the README

PR #1028, opened 2026-01-27
XIOS: test `StructureFactory::fileFromState()` and `Model::configure()`

### LINKED ISSUE #1027: Enable `fileFromState` with XIOS
Issue #1022 checks we can use `StructureFactory::stateFromFile` with XIOS because this gets called in the `Model::run()` driver. Similarly, we should check we can use `StructureFactory::fileFromState`.

joewallwork:
# XIOS: test `StructureFactory::fileFromState()` and `Model::configure()`

Fixes #1027
Merges into #1023

### Task List

---
# Change Description

Minor code change: call `close_context_definition` in XIOS' finalizer, if it hasn't already been called before.

---
# Test Description

The XIOS tests are reworked to make use of `StructureFactory::fileFromState` instead of `ParametricGrid::dumpModelState` in a couple of places. This is done because it's how restarts are actually written in the model.

For further consistency with the model, we call `Model::configure()` in the restart tests rather than `Model::configureRestarts()` followed by `Model::configureTime()`. To enable that, I needed to rework how the fields are defined, e.g., the `Model` expects `cice`, `hice`, `damage`, and `hsnow` to be DG fields. Also, the grid fields need to be defined correctly.

---
# Documentation Impact

The XIOS docs page has been updated to mention `Model::configure()` and also the fact that the XIOS I/O implementation doesn't currently support Cartesian grids, i.e., those based on `x_dim` and `y_dim` rather than `longitude` and `latitude`. (This limitation is because we haven't defined XIOS domains that `x_dim` and `y_dim` could live on as variables.)

---
### Pre-Request Checklist

### DIFF
--- a/core/test/XiosReadDiagnostic_test.cpp
+++ b/core/test/XiosReadDiagnostic_test.cpp
@@ -59,7 +59,6 @@ MPI_TEST_CASE("TestXiosReadDiagnostic", 2)
     auto& modelMPI = ModelMPI::getInstance(test_comm);
 
     // Create a Model and configure it so that time options are parsed
-    // TODO: Use Model.configure for consistency with the rest of the model
     Model model;
     model.configureRestarts();
     model.configureTime();
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -176,7 +176,13 @@ void Xios::close_context_definition()
 void Xios::context_finalize()
 {
     if (isEnabled) {
+        // Close the context definition, if it wasn't already closed
+        close_context_definition();
+
+        // Ensure the dimension and variable names are correct in the output files
         postprocessOutputFiles();
+
+        // Finalize the XIOS context
         cxios_context_finalize();
     }
 }
--- a/core/test/XiosAxis_test.cpp
+++ b/core/test/XiosAxis_test.cpp
@@ -56,7 +56,6 @@ MPI_TEST_CASE("TestXiosAxis", 3)
     REQUIRE(xiosHandler.getAxisSize("DGSAxis") == 8);
     REQUIRE(xiosHandler.getAxisSize("VertexAxis") == 2);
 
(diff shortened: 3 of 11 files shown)


core/test/XiosReadDiagnostic_test.cpp
```diff
@@ -59,7 +59,6 @@ MPI_TEST_CASE("TestXiosReadDiagnostic", 2)
     auto& modelMPI = ModelMPI::getInstance(test_comm);
 
     // Create a Model and configure it so that time options are parsed
-    // TODO: Use Model.configure for consistency with the rest of the model
```
TomMelt, line 62: just checking this wasn't missed by mistake, but in this file and `core/test/XiosReadForcing_test.cpp` you mentioned using `model.configure()` instead of `model.configureRestarts()` and `model.configureTime()`.

In `core/test/XiosReadRestart_test.cpp` you have used `model.configure()`.

Is that deliberate?

TomMelt (approved):
Thanks @joewallwork , looks good to me. I have a question on `core/test/XiosReadDiagnostic_test.cpp` and `core/test/XiosReadForcing_test.cpp`. Assuming you didn't mean to use `model.configure()` then it's good to go.

core/test/XiosReadDiagnostic_test.cpp
joewallwork: Yeah this was something I decided not to use it in the diagnostic and forcing tests in the end, otherwise we'd need to worry about other fields that define the coordinates such as latitude, longitude, etc like we now need to do in the restart tests. I think it's sufficient to check we can set things up correctly with `model.configure()` in the restart tests.

TomMelt: Thanks @joewallwork ,  that makes sense. No need to over complicate the other tests.

PR #1030, opened 2026-01-28
Adapt `install-xios.sh` to work locally

joewallwork:
I've had versions of this branch lying around for my personal use for a while so I thought I might as well open a PR to get it merged.

The PR specifies `INSTALL_DIR` as `/xios` if running in a Docker container, and just `XIOS3` (the default) otherwise. The logic in rest of the script is updated.

PR #1031, opened 2026-01-28
Tweak CI test triggers

joewallwork:
Currently, if the `install-xios.sh` script is changed (e.g., in #1030) then the test suite gets triggered because of its `.sh` extension. However, this is redundant for the following reasons:

1. The `install-xios.sh` script updates XIOS in the Docker image.
2. However, before a PR is merged, the Docker build CI workflow only checks that it can be built. It doesn't push the image to ghcr.
3. If the tests are run before the PR is merged, they use the old version of the Docker image and don't pick up any updates to it.

A similar thing happens with changes to shell scripts in the `docs` subdirectory.

This PR changes the triggers so that the test suite gets triggered by changes to shell scripts in the `core`, `run`, and `dynamics` subdirectories, but not `docs` or `Dockerfiles`.

I added a periodic trigger two hours after the 3-monthly Docker build run so that the suite can test that the new image is working as expected.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -74,8 +74,7 @@ jobs:
         apt install -y python3-venv
         python3 -m venv ~/nextsim
         . $HOME/nextsim/bin/activate
-        pip install numpy
-        pip install netCDF4
+        pip install netCDF4 numpy
 
     - name: run serial tests
       run: |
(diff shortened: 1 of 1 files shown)


joewallwork:
I just added in a couple of further tweaks:
* `pip install netCDF4` in the XIOS workflow. Currently the build hits Python `ModuleNotFound` errors, even if it does just continue past them.
* Provide a commit hash for the `tmate` action. Otherwise, when you uncomment it Zizmor complains (see #1036).

TomMelt (approved):
Thanks this looks good to me.

PR #1035, opened 2026-02-04
Check XIOS can handle NaN values

joewallwork:
# Check XIOS can handle NaN values

Fixes #1033 
~~Merges into #1028~~

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

n/a

---
# Test Description

This PR updates the XIOS restart tests to confirm that it can handle NaN values in the expected manner. NaN values are used for land-masked parts of the domain in the main model, so the XIOS testing examples have been updated to account for this.

~~I also extended the testing to apply the `=` operator overload of `ModelArray`, so that the `DGField` variable `hsnow` is initialised by "projecting" the `HField` variable `mask`. This will be useful when looking into supporting reading and writing a field using different discretisations in #1034.~~ Edit: defer for now to maintain separation-of-concerns.

---
# Documentation Impact

n/a

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] This change conforms to the conventions described in the README

Issue #1034, opened 2026-02-04
XIOS: allow a field to be read and written with different types


joewallwork:
In nextSIM-DG, fields that are treated as `DGField`s by the model are often initialised from `HField` data in an initial condition file. They may subsequently be written out as `DGField` restarts and later runs can read in the restart from `DGField`. This means that we need to separate out the reading and writing part of a field to improve generality. The overloaded `=` operator of `ModelArray` may be used to "project" from `HField` to `DGField` so we'll probably need to make use of this.

joewallwork:
XIOS error from the June '23 config illustrating the issue:
```
In file "model_to_client_source_filter.cpp", function "void xios::CModelToClientSourceFilter::streamData(xios::CDate, const xios::CArray<double, N>&) [with int N = 3]",  line 34 -> [ Awaiting data of size = 4503, Received data size = 13509 ] The data array does not have the right size! grid =  --> HGrid

(1) **************** void cxios_write_data_k83_hdl(xios::CField*, double*, int, int, int)
```

PR #1036, opened 2026-02-04
DO NOT MERGE - debugging XIOS

joewallwork:
Debugging for #1035


joewallwork:
## Record of debugging in this PR

* Set up a `tmate` session that'd built in `Debug` mode and installed debuggers, ready to do some debugging. Joined the session but when I ran the XIOS tests, they all passed.
* Exited the `tmate` session and checked the tests still pass if I run in the usual way (not interactive session).
* Tried switching back to a `Release` mode build and this gave the same error that crops up in #1035.

joewallwork:
Back in the original PR, I removed the conversion from `HField` to `DGField`, deferring it to when I work on #1034 for better separation-of-concerns. It turns out this was the culprit. Given that @Thanduriel has also reported issues with the overloaded `=` operator for `ModelArray` that gets used under the hood, this is something we should look into.

PR #1037, opened 2026-02-05
XIOS: allow a field to be read and written with different types

### LINKED ISSUE #1034: XIOS: allow a field to be read and written with different types

joewallwork:
# XIOS: allow a field to be read and written with different types

Fixes #1034

### Task List

---
# Change Description

For any fields that are configured to be read, their type is determined upon the initial call to `ModelMetadata::setDimensionsFromFile`. For any fields to be written out, their type is configured through a call to `setFieldType`. This breaks down if a field is configured differently in the two cases, but this can and does occur in nextSIM-DG.

This PR modifies the "base" and "inherited" field relationship. The latter is renamed "input" for clarity. In cases where the two have the same `ModelArray::Type`, they are configured to share the same data using an XIOS field reference. If the input type is `HField` and the base type is `DGField` then this is recorded so that we know to do the conversion upon reading that field. In other cases, an error is thrown.

Other changes:
* Set the grid reference for a field if it never gets set.
* Throw an error if a field configured to be read as a forcing is not of `HField` type.
* Throw an error if `getFieldType` is called but the field type hasn't been set.
* Privatise `setFieldType` and instead provide public API variants `setPrognosticFieldType` and `setDiagnosticFieldType` that call it. (This only needs to be usable for fields being written out.)

---
# Test Description

The XIOS restart tests are updated so that the `hsnow` variable is a `HField` in the `xios_test_input.nc` file but gets interpreted as a `DGField` in the model.

My only slight concern is that the read-write test converts the `hsnow` variable from its `HField` input
```
 hsnow =
  0, 1, 2, 3,
  4, 5, 6, 7 ;
```
and writes it out as `DGField`
```
 hsnow =
  0, 1, 2, 3, 4, 5,
  6, 7, _, _, _, _,
  _, _, _, _, _, _,
  _, _, _, _, _, _,
  _, _, _, _, _, _,
  _, _, _, _, _, _,
  _, _, _, _, _, _,
  _, _, _, _, _, _ ;
```
Is this right?

---
# Documentation Impact

---
### Pre-Request Checklist


timspainNERSC:
> Test Description
> … 
> Is this right?

1. The absent DG components should be 0, not missing or whatever else `_` is representing here.

2. The in-memory representation of a DG field would be to have the DG component varying fastest. *If* this is reflected in the output of the test, then I'd expect the output to be:

```
hsnow =
1, 0, 0, 0, 0, 0,
2, 0, 0, 0, 0, 0,
3, 0, 0, 0, 0, 0,
4, 0, 0, 0, 0, 0,
5, 0, 0, 0, 0, 0,
6, 0, 0, 0, 0, 0,
7, 0, 0, 0, 0, 0,
8, 0, 0, 0, 0, 0;
```

joewallwork:
>     1. The absent DG components should be 0, not missing or whatever else `_` is representing here.
> 
>     2. The in-memory representation of a DG field would be to have the DG component varying fastest. _If_ this is reflected in the output of the test, then I'd expect the output to be:
> 
> 
> ```
> hsnow =
> 1, 0, 0, 0, 0, 0,
> 2, 0, 0, 0, 0, 0,
> 3, 0, 0, 0, 0, 0,
> 4, 0, 0, 0, 0, 0,
> 5, 0, 0, 0, 0, 0,
> 6, 0, 0, 0, 0, 0,
> 7, 0, 0, 0, 0, 0,
> 8, 0, 0, 0, 0, 0;
> ```

Thanks @timspainNERSC, I'll look into the code here and see what might need changing.

joewallwork:
> Thanks @timspainNERSC, I'll look into the code here and see what might need changing.

I've now fixed this issue. It will be addressed by #1042 (which supersedes this PR) and a follow-up one that will actually implement the test.

PR #1040, opened 2026-02-09
Full GPU support

### LINKED ISSUE #672: GPU thermodynamics implementation

Thanduriel:
# Full GPU support

related Issues: #671, #672

---

# Change Description
Ported all major parts of neXtSIM-DG to Kokkos, enabling fast execution on a GPU. This involves major changes to the dynamics and physics (thermodynamics) but also the core infrastructure related to data exchange between `ModelComponent`s.

## Core
The most invasive change is the introduction of `ModelArrayAccessor`. This handle replaces `ModelArrayRef` as a mechanism to share fields between `ModelComponent`s but also takes care of host/device data transfers in a transparent manner. To achieve the latter, the new handle as different ergonomics. In order to a access field, one has to explicitly request read or write access for either host or device each time. The resulting `ModelArray` (or `ModelArrayDevice`) reference should only be used within the current scope to ensure that host and device buffers can be synced in between uses as needed.

### Portable kernels
In order to write code just once that runs with and without Kokkos, a thin wrapper is provided in `KernelAlternatives.hpp`. These utilities are meant to be used in combination with the `Auto` variants that map to the chosen backend, e.g.
- `ModelArrayAccessor::getAutoRW()`/ `ModelArrayAccessor::getAutoRO()` to get the buffer
- `overElementsAuto` to run the computations,
- `OVER_ELEMENTS_LAMBDA` to define an appropriate lambda.

Additionally, device code has a number of restrictions which have to be kept in mind in order to make a computation kernel  portable. These are:

- Every variable used needs to be captured so that it can be copied to the device. This is mostly done through the implicit capture mechanism of lambdas. For device execution, everything is captured by copy and than automatically transferred to the device. Most notably, this includes local `ModelArrayDevice`  objects, acquired through a `ModelArrayAccessor` which are pointers to already existing buffers.  Static class variables on the other hand need to be assigned to a local variable first in order to capture them.
- Functions called within a kernel need to be marked for device execution and satisfy the same constrains as the kernel. They have to be defined inline and should probably be pure functions. Member functions are fine only if the whole object can be captured by the lambda and if they are not virtual.
- Standard math and utility functions should be taken from the namespace `Utils` which maps to either `std` or `Kokkos`.
- Global constants can be used directly, unless they only occur as argument to a template function (e.g. `min`, `max`). In the latter case, just assign the constant to a local variable first.

## Physics
The major thermodynamics modules where rewritten as portable kernels. To get around the kernel restrictions, two modules had to be converted into proper `ModelComponent`s since their implementation as virtual function that operates on a single element does not work anymore.  `IIceAlbedo` introduces new fields which are updated internally. The albedo parameter `i0` (previously also `I_0` in some cases) now also resides with the concrete `IIceAlbedo` implementation instead of the calling module. The other module with a changed interface is `IFreezingPoint`. Since the update was already done on existing fields fewer changes where needed and `FreezingPointImpl` provides a convenient mechanism to generate the needed implementation and the old interface while only requiring the definition of a single element function, as before.
 
With disciplined use of accessors it is still easily possible to mix host and device code. Adding new modules in the old style is therefore not an issue. However, switching the execution space within a single module is discouraged since this incurs a significant cost from memory transfers and it is easier to introduce synchronization bugs.

## Dynamics
Kokkos variants of the dynamics are built on top of the old ones. The interface was kept mostly the same with some additions to minimize needed host/device data transfers when interacting with the other parts of neXtSIM-DG. To port the computations additional tools where needed compared to the thermodynamics. Most important is the Kokkos / Eigen interface provided in `KokkosUtils.hpp` that makes switching between Kokkos views and Eigen matrices effortless while preserving compile-time information.

In principle, it would be possible to roll the remaining parts that are used from the old variants into the Kokkos dynamics and keep only the latter to ease future maintenance. However, this is out of scope for the current PR as their are some remaining roadblocks:
* either Kokkos become mandatory for the build or an additional abstraction layer similar to the thermodynamics is needed
* the Kokkos code is more difficult to work with, presenting a barrier of entry for future developments
* some of the Kokkos kernels are not deterministic; while it should be possible to change this at some performance cost, achieving good performance under this constrain would likely require specialized kernels for CPU and GPU
* MPI support

---
# Test Description

Many existing tests had to be updated to work with the new `ModelArrayAccessor` infrastructure. When Kokkos is enabled, tests automatically use the selected backend. For full code coverage, it is therefore best to run the whole suite with and without Kokkos enabled.

---
# Documentation Impact

Build instructions where added to the `installation.rst`. After the build, there should be little difference for users.


core/src/modules/DynamicsModule/include/BBMDynamics.hpp
```diff
@@ -29,6 +30,12 @@ class BBMDynamics : public IDynamics, public Configured<BBMDynamics> {
         double lowerLimit = -std::numeric_limits<double>::infinity(),
         double upperLimit = std::numeric_limits<double>::infinity()) override;
 
+#ifdef USE_KOKKOS
+    void advectField(double timestep, const DeviceViewMA& field,
+        double lowerLimit = -std::numeric_limits<double>::infinity(),
+        double upperLimit = std::numeric_limits<double>::infinity()) override;
+#endif
+
```
einola, line 38: Why override the base class function?

physics/src/modules/AtmosphereBoundaryModule/MU71Atmosphere.cpp
```diff
@@ -63,46 +50,63 @@ MU71Atmosphere::MU71Atmosphere()
...
-
-    // Not needed/specified by M&U '71
-    qow[i] = 0.;
-    subl[i] = 0.;
-    rain[i] = 0.;
-    evap[i] = 0.;
-    uwind[i] = 0.;
-    vwind[i] = 0.;
+    overElements(
+        [&](size_t i, const TimestepTime& tsTime) {
```
einola: Why not use ``overElementsAuto`` here?

core/test/ModelArrayAccessor_test.cpp
```diff
@@ -0,0 +1,354 @@
...
+//     REQUIRE(differenceRWScalar(0, 0) == 5. - 7.);
+//
+//     ModelArray productRWScalar = rwRef * scalar;
+//     REQUIRE(productRWScalar.getType() == ModelArray::Type::H);
+//     REQUIRE(productRWScalar(0, 0) == 5. * 7.);
+//
+//     ModelArray ratioRWScalar = rwRef / scalar;
+//     REQUIRE(ratioRWScalar.getType() == ModelArray::Type::H);
+//     REQUIRE(ratioRWScalar(0, 0) == 5. / 7.);
+// }
```
timspainNERSC: If this is no longer needed, then it can be removed rather than commented out.

dynamics/src/include/DynamicsKernel.hpp
```diff
@@ -139,6 +138,23 @@ template <int DGadvection, int DGstress> class DynamicsKernel {
...
+    // bad interface: copy can't be avoided, no overload possible
+    // virtual DeviceViewMA getDG0Data(const std::string& name) const;
+    // alternative:
+    // virtual void getDG0Data(const std::string& name, const DeviceViewMA& destination) const;
+    // other thinks to consider:
+    // * should make datatransfers in kernel and mirrored views obsolete
+    // * to completely eliminate host views have to also port computeSeaSurfaceHeight?
+    // * no data transfer capability in kernel would it more difficult to debug / develop new
+    // features
+
```
timspainNERSC, line 157: Is this a TODO, a FIXME or just some musings?

dynamics/src/CGDynamicsKernel.cpp
```diff
@@ -336,6 +336,8 @@ template <int DGadvection> void CGDynamicsKernel<DGadvection>::projectVelocityTo
             // by integrating rhs and inverting with dG(stress) mass matrix
             //
             e11.row(dgi) = pmap->iMgradX[dgi] * vx_local;
+            //    for(int i = 0; i < 8; ++i)
+            //        e11.row(dgi)(i) = pmap->iMgradX[dgi](i,0);
```
timspainNERSC: Is this a TODO or for some future use?

timspainNERSC:
To see if I understand the changes to how `ModelArray` works with Kokkos:
Just like before there was a store of `ModelArrayReference`s, now all `ModelArray`s are now accessed through the `ModelArrayStore`. This allows a store to exist both in the CPU (host) memory space and in the GPU (device) memory space. When running with a GPU, data is copied between the two when necessary. If running without a GPU, then there is only one store in the CPU. This gives all access to `ModelArray`s a consistent interface, using the `Accessors` with a similar syntax to how `ModelArrayReference`s were previously accessed.  A `Module` 'owning' an array still means that the class allocates memory for that array, but now it does so in the `ModelArrayStore` as opposed to in a class member variable.

dynamics/src/kokkos/KokkosDGTransport.cpp
```diff
@@ -0,0 +1,825 @@
...
+                    dt * (Kokkos::fmax(vel, zero) * left + Kokkos::fmin(vel, zero) * right));
+                Kokkos::atomic_add(&phiupDevice(c2),
+                    dt * (Kokkos::fmax(vel, zero) * left + Kokkos::fmin(vel, zero) * right));
+            });
+    } else {
+        const auto PSIe1D = PSIe<EDGE_DOFS, GP1D>;
+        const auto PSIew1 = PSIe_w<DG, GP1D, 1>;
+        const auto PSIew3 = PSIe_w<DG, GP1D, 3>;
+        Kokkos::parallel_for(
+            "edgeTermX", phiDevice.extent(0), KOKKOS_LAMBDA(const DeviceIndex eid) {
```
timspainNERSC: Perhaps should be `edgeTermY`?

timspainNERSC (approved):
There seem to be several large chunks of commented-out code. If this are to be retained, it would be good to have some kind of TOD noting why they are being kept.
Most of the changes look to be changing the array access from CPU `ModelArray`s and `C/DGVector`s  to the new Kokkos arrays. If the unit tests pass, then I assume that the conversion have been written correctly for the CPU only case, at least.
Also, github doesn't much like it when there isn't a newline at the end of a file It would be nice to have fewer of those red symbols.

Since these are minor changes to comments and formatting, I'll approve this, leaving it up to you to delete/reformat what you see fit.

Thanduriel:
> To see if I understand the changes to how `ModelArray` works with Kokkos: Just like before there was a store of `ModelArrayReference`s, now all `ModelArray`s are now accessed through the `ModelArrayStore`. This allows a store to exist both in the CPU (host) memory space and in the GPU (device) memory space. When running with a GPU, data is copied between the two when necessary. If running without a GPU, then there is only one store in the CPU. This gives all access to `ModelArray`s a consistent interface, using the `Accessors` with a similar syntax to how `ModelArrayReference`s were previously accessed. A `Module` 'owning' an array still means that the class allocates memory for that array, but now it does so in the `ModelArrayStore` as opposed to in a class member variable.

Yes, that's basically it, albeit ownership technically always resides with `ModelArrayStore` now.  To preserve the semantics of `ModelArrayRefStore` regarding multiple registrations,  the `ModelArray` itself will always be default constructed when the first `ModelArrayAccessor` asks for it. The only differences for the "owning" module are that it provides proper arguments for the `ModelArray` constructor and that this sets an internal flag in `ModelArrayStore`.

core/src/modules/DynamicsModule/include/BBMDynamics.hpp
Thanduriel: I am not sure if I understood your question. The `DeviceViewMA` variant of `advectField` is needed to advect fields outside of the dynamics directly on the GPU. Just like the host version operating on `ModelArray` the implementation has to reside in the concrete dynamics module because it calls the dynamics kernel implementation. This is one of the few places where virtual functions are actually needed right now in the dynamics because `FieldAdvection` only knows the base type `IDynamics`.

physics/src/modules/AtmosphereBoundaryModule/MU71Atmosphere.cpp
Thanduriel: I decided that it was not worth the effort to port `MU71Atmosphere` because it would require rewriting `monthlyCubicBSpline` and I assumed that this module would not see much use anyway. If you think it would be worthwhile I can still port the module.

For now, I just replaced the `overElements` call with the new template based host variant to get rid of the old variant using `std::function`.

PR #1041, opened 2026-02-10
Custom log formatting

### LINKED ISSUE #1039: Log formatting severity level
The log does not show the correct severity level which makes it difficult to identify important messages. For example, the output of a successful run of  `testConfigOutput` contains
```
[2026-02-09 16:20:55.467942] [0x00007e33aba36b80] [info]    Accessing the ModelComponent::getStore() after it was destroyed. A new store will be constructed. In a test with multiple cases this is the desired behaviour.
[2026-02-09 16:20:55.468317] [0x00007e33aba36b80] [info]    ConfigOutput: No field with the name "top_melt" was found.
[2026-02-09 16:20:55.468868] [0x00007e33aba36b80] [info]    ConfigOutput: Outputting 4 fields to diag0111.nc at 2020-01-11T00:00:00Z
```
Everything is tagged as [info] even though the first two messages are actually warnings. This is the expected behavior since we use custom severity levels in `boost::log`.  According to the [docs](https://www.boost.org/doc/libs/latest/libs/log/doc/html/log/tutorial/attributes.html) one has to define a suitable `operator<<` for the severity type in order to get customized formatting. However, my initial attempt at this did not work and maybe it would be better to fully customize the outputs. The second argument `[0x00007e33aba36b80]` of the standard format is also not helpful with our current logging setup.

Thanduriel:
# Improved log formatting

Fixes #1039

---
# Change Description

Logged messages now also display current local time and the severity level. This is implemented using standard `boost::log` customization points in `Logged::configure`. Thus, improved formatting is not used in some tests (like the one referenced in the issue) where `Model::configure` is not called.

# Test Description
I run some cases with a verbose log level, e.g.:
```
[Logged]
minimum_log_level = all
console_log_level = all
```


CMakeLists.txt
```diff
@@ -57,22 +56,21 @@ find_package(Eigen3 3.4 REQUIRED)
...
 set(ScriptDirectory "${PROJECT_SOURCE_DIR}/scripts")
 set(ModuleBuilderScript "${ScriptDirectory}/module_builder.py")
 set(ModuleHeaderScript "${ScriptDirectory}/module_header.py")
 
-# Dynamics type. Defaults to DG1.
-# Available options are:
-# DG1 Discontinuous Galerkin, degree 1 (3 components)
-# DG2 Discontinuous Galerkin, degree 2 (6 components)
+# Dynamics type. Defaults to DG1. Available options are: DG1 Discontinuous Galerkin, degree 1 (3
+# components) DG2 Discontinuous Galerkin, degree 2 (6 components)
```
timspainNERSC: What or whoever has reformatted the CMakeLists has definitely made this comment less readable.

timspainNERSC (dismissed):
Good that the logging is finally getting looked at!
The CMake formatter has made a mess of the layout of one comment. The rest looks fine, though.

CMakeLists.txt
Thanduriel: Yep, that should not happen. This is `cmake-format` at work, the tool we also have in the CI. I reformatted the comment in a list style that is preserved.

einola (approved):

Issue #1026, opened 2026-01-27
Implement interfaces to get restart field names out of modules


joewallwork:
XIOS needs to know which fields to write as restarts. Beyond the core fields, we need interfaces for the different modules to tell it which fields they need to write out, too.

PR #1042, opened 2026-02-10
XIOS: allow a field to be read and written with different types (take 2)

### LINKED ISSUE #1034: XIOS: allow a field to be read and written with different types

joewallwork:
# XIOS: allow a field to be read and written with different types

Fixes #1034
Supersedes #1037

### Task List

---
# Change Description

For any fields that are configured to be read, their type is determined upon the initial call to `ModelMetadata::setDimensionsFromFile`. For any fields to be written out, their type is configured through a call to `setFieldType`. This breaks down if a field is configured differently in the two cases, but this can and does occur in nextSIM-DG.

This PR modifies the "base" and "inherited" field relationship. The latter is renamed "input" for clarity. In cases where the two have the same `ModelArray::Type`, they are configured to share the same data using an XIOS field reference. If the input type is `HField` and the base type is `DGField` then this is recorded so that we know to do the conversion upon reading that field. In other cases, an error is thrown.

Other changes:
* Set the grid reference for a field if it never gets set.
* Throw an error if a field configured to be read as a forcing is not of `HField` type.
* Throw an error if `getFieldType` is called but the field type hasn't been set.
* Privatise `setFieldType` and instead provide public API variants `setPrognosticFieldType` and `setDiagnosticFieldType` that call it. (This only needs to be usable for fields being written out.)

---
# Test Description

In this version, the tests have not yet been modified such that a conversion would actually take place. This is because the field type needs to be set inside the call to `Model::configure()`. I'll do that in a follow-up PR while addressing #1026.

---
# Documentation Impact

---
### Pre-Request Checklist

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -1588,7 +1588,7 @@ void Xios::write(const std::string& fieldId, const ModelArray& modelarray)
     const ModelArray::Type& expectedType = getFieldType(fieldId);
     if (expectedType != type) {
         throw std::runtime_error(
-            "Xios::read: field '" + fieldId + "' does not have the expected type");
+            "Xios::write: field '" + fieldId + "' does not have the expected type");
     }
 
     // Write out according to field type
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -172,8 +172,11 @@ void Xios::close_context_definition()
                 }
 
                 // Link the input field to the base field if their types align
-                const ModelArray::Type& inputType = getFieldType(inputFieldId);
                 const ModelArray::Type& baseType = getFieldType(fieldId);
+                if (ioType == FORCING && baseType != ModelArray::Type::H) {
+                    throw std::runtime_error("Xios: Forcing fields must be treated as HFields");
+                }
+                const ModelArray::Type& inputType = getFieldType(inputFieldId);
                 if (inputType == baseType) {
                     cxios_set_field_field_ref(
                         getField(inputFieldId), fieldId.c_str(), fieldId.length());
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -1616,7 +1616,6 @@ void Xios::read(const std::string& fieldId, ModelArray& modelarray)
     if (modelarray.nDimensions() != 2) {
         throw std::invalid_argument("Only ModelArrays of dimension 2 are supported");
     }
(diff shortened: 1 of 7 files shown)


core/src/Xios.cpp
```diff
@@ -1524,6 +1583,15 @@ void Xios::write(const std::string& fieldId, const ModelArray& modelarray)
     auto& dims = modelarray.dimensions();
     const ModelArray::Type& type = modelarray.getType();
     domainWritten[domainIds[type]] = true;
+
+    // Check the field type
+    const ModelArray::Type& expectedType = getFieldType(fieldId);
+    if (expectedType != type) {
+        throw std::runtime_error(
+            "Xios::read: field '" + fieldId + "' does not have the expected type");
```
joewallwork: Copy-paste error
```suggestion
            "Xios::write: field '" + fieldId + "' does not have the expected type");
```

core/src/Xios.cpp
```diff
@@ -1551,14 +1619,41 @@ void Xios::write(const std::string& fieldId, const ModelArray& modelarray)
 void Xios::read(const std::string& fieldId, ModelArray& modelarray)
 {
     if (!getFieldReadAccess(fieldId)) {
-        throw std::runtime_error("Xios::read: field " + fieldId
-            + " is not configured for reading, but is being read from file.");
+        throw std::runtime_error("Xios::read: field '" + fieldId
+            + "'' is not configured for reading, but is being read from file.");
```
timspainNERSC: I think this needs to only be one single quotation mark

joewallwork: Good spot! Thanks
```suggestion
            + "' is not configured for reading, but is being read from file.");
```

timspainNERSC (approved):
Without getting into the depths of XIOS, this all looks like it is doing what it should be doing.

joewallwork:
> Without getting into the depths of XIOS, this all looks like it is doing what it should be doing.

Thanks Tim!

PR #1044, opened 2026-02-16
Halo add pb test for halo exchange

TomMelt:
# Add periodic boundary test for Halo Exchange

### Task List
- [ ] Defined the tests that specify a complete and functioning change
- [ ] Commented all code so that it can be understood without additional context
- [ ] No new warnings are generated or they are mentioned below
- [ ] The documentation has been updated (or an issue has been created to do so)

---
# Change Description

- Adds a test for halo exchange with periodic boundaries

---
# Test Description

### `HaloExchangePB_test.cpp`

`HaloExchangePB_test.cpp` is based on `HaloExchangeCB_test.cpp`. It creates test data for all the array types:
- `HField`
- `VertexField`
- `DGField`
- `DGVector`
- `CGVector`

They  are created in such a way that the value of each cell can be calculated based on its indices. Therefore we can check after exchange that all of the cells have been exchanged successfully.

This would be a good PR for @niravshah241 to take over.

### DIFF
--- a/core/test/HaloExchangePB_test.cpp
+++ b/core/test/HaloExchangePB_test.cpp
@@ -1,0 +1,313 @@
+/*!
+ * @author  Tom Meltzer <tdm39@cam.ac.uk>
+ * @brief   Test halo exchange class for Periodic Boundaries (PB)
+ * @details
+ *
+ * Test halo exchange class for Periodic Boundaries (PB) on following fields:
+ * - HField
+ * - VertexField
+ * - DGField
+ * - DGVector
+ */
+
+#include <doctest/extensions/doctest_mpi.h>
+
+#include "ModelMPI.hpp"
+#include "ModelMetadata.hpp"
+#include "include/DGModelArray.hpp"
+#include "include/Halo.hpp"
+
+namespace Nextsim {
+
+const std::string testFilesDir = TEST_FILES_DIR;
+const std::string file = testFilesDir + "/partition_metadata_3_pb.nc";
+
+static const int DG = 3;
+static const bool debug = false;
+
+// reference data for each process
+static std::array<std::vector<double>, 3> refDataAllProcs = {
+    std::vector<double>(
+        { 0, 208, 218, 228, 238, 248, 258, 268, 0, 390, 100, 110, 120, 130, 140, 150, 160, 370, 391,
+            101, 111, 121, 131, 141, 151, 161, 371, 392, 102, 112, 122, 132, 142, 152, 162, 372,
+            393, 103, 113, 123, 133, 143, 153, 163, 373, 0, 204, 214, 224, 234, 244, 254, 264, 0 }),
+    std::vector<double>({ 0, 103, 113, 123, 133, 143, 153, 163, 0, 394, 204, 214, 224, 234, 244,
+        254, 264, 374, 395, 205, 215, 225, 235, 245, 255, 265, 375, 396, 206, 216, 226, 236, 246,
+        256, 266, 376, 397, 207, 217, 227, 237, 247, 257, 267, 377, 398, 208, 218, 228, 238, 248,
+        258, 268, 378, 0, 100, 110, 120, 130, 140, 150, 160, 0 }),
+    std::vector<double>({ 0, 378, 388, 398, 0, 160, 370, 380, 390, 100, 161, 371, 381, 391, 101,
+        162, 372, 382, 392, 102, 163, 373, 383, 393, 103, 264, 374, 384, 394, 204, 265, 375, 385,
+        395, 205, 266, 376, 386, 396, 206, 267, 377, 387, 397, 207, 268, 378, 388, 398, 208, 0, 370,
+        380, 390, 0 }),
+};
+
+void initializeTestData(ModelArray::DataType& data, size_t localNx, size_t localNy, size_t offsetX,
+    size_t offsetY, size_t numComps, int test_rank)
+{
+    // initialize with test data
+    for (size_t j = 0; j < localNy; ++j) {
+        for (size_t i = 0; i < localNx; ++i) {
+            for (size_t d = 0; d < numComps; ++d) {
+                data(d + i * numComps + j * localNx * numComps)
+                    = (d + 1) * 1000 + (test_rank + 1) * 100 + (i + offsetX) * 10 + (j + offsetY);
+            }
+        }
+    }
+}
+
+TEST_SUITE_BEGIN("Halo exchange tests");
+MPI_TEST_CASE("test halo exchange on 3 proc grid with periodic boundary conditions", 3)
+{
+    // test for a 3 proc grid
+    // ┌─┬─┐
+    // │1│ │
+    // ├─┤2│
+    // │0│ │
+    // └─┴─┘ (proc id)
+
+    auto& modelMPI = ModelMPI::getInstance(test_comm);
+    auto& metadata = ModelMetadata::getInstance(file);
+
+    const size_t nx = metadata.getGlobalExtentX();
+    const size_t ny = metadata.getGlobalExtentY();
+    const size_t localNx = metadata.getLocalExtentX() + 2 * Halo::haloWidth;
+    const size_t localNy = metadata.getLocalExtentY() + 2 * Halo::haloWidth;
+    const size_t offsetX = metadata.getLocalCornerX();
+    const size_t offsetY = metadata.getLocalCornerY();
+
+    ModelArray::setDimension(ModelArray::Dimension::X, nx, localNx, offsetX);
+    ModelArray::setDimension(ModelArray::Dimension::Y, ny, localNy, offsetY);
+    ModelArray::setDimension(ModelArray::Dimension::XVERTEX, nx + 1, localNx + 1, offsetX);
+    ModelArray::setDimension(ModelArray::Dimension::YVERTEX, ny + 1, localNy + 1, offsetY);
+
+    // create example 2D field on each process
+    auto testData = ModelArray::HField();
+    testData.resize();
+    testData = 0.;
+
+    // create halo for testData model array
+    Halo halo(testData);
+
+    // create and allocate temporary Eigen array
+    ModelArray::DataType innerData;
+    innerData.resize(halo.getInnerSize(), testData.nComponents());
+    initializeTestData(
+        innerData, localNx - 2, localNy - 2, offsetX, offsetY, testData.nComponents(), test_rank);
+
+    // populate inner block of modelarray
+    halo.setInnerBlock(innerData, testData.getDataRef());
+    // exchange halos
+    halo.exchangeHalos(testData.getDataRef());
+
+    // initialize reference data for each proc
+    ModelArray::DataType refData;
+    refData.resize(localNx * localNy, 1);
+    refData
+        = Eigen::Map<ModelArray::DataType>(refDataAllProcs[test_rank].data(), refData.size(), 1);
+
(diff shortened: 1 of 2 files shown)


Mikolaj-A-Kowalski:
Resurrecting and taking over this PR. 

To get started rebased on the current `develop`.

Mikolaj-A-Kowalski:
This one is now ready to go in my opinion. 

The main things to discuss are 
- How clear is the wraparound logic and the mapping `local` -> `global` for the cells and points coordinates? I found it to be the most confusing part of the `CB` version of the test and have changed it significantly in the `PB` version. 
- Should we refactor `CB` test as well to work with signed integers as coordinates?

Mikolaj-A-Kowalski:
Ok. We are running into issues with the failing CI but these are related by some issues with the `entrypoint.sh` scripts.

Mikolaj-A-Kowalski:
With the Docker image being back online I think this one is ready for review now.

niravshah241:
I tried to run the tests ``testHaloExchangeCB_MPI3`` and ``testHaloExchangePB_MPI3`` an both are failing. Is this PR ready for review? I also think that the task list is probably done but not marked with a check sign.

Mikolaj-A-Kowalski:
> I tried to run the tests testHaloExchangeCB_MPI3 and testHaloExchangePB_MPI3 an both are failing. Is this PR ready for review? I also think that the task list is probably done but not marked with a check sign.

It is ready for review and the tests were working fine on my side (CI also seems to be happy). What kind of error message are you getting?

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+ * @brief   Test halo exchange class for Periodic Boundaries (PB)
+ * @details
+ *
+ * The purpose of this test is to verify the Halo exchange.
+ *
+ * We accomplish that by:
+ *  - Assigning each value in the grid with a unique id (hash) based on its global
+ *    coordinates
+ *  - We take care to not assign any values in the halo region for each rank
+ *  - We then perform tha halo exchange
```
jatkinson1000, line 12:
```suggestion
 *  - We then perform the halo exchange
```

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+ *  - For each rank we verify that the values stored in the field match the
+ *    hash
```
jatkinson1000, line 14: Not clear what this means.
You check that the hash recomputed using the remapped variables matches the original? You check that the values at the new location match those at the origin hash? You check that the hash recomputed at the new location matches the origin and that the values are the same?

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+static const int CGDEGREE = 2;
+static const bool debug = false;
+
+/**
+ * Helper class to group all domain-dependent calculations
+ *
+ * It is hardcoded for a periodic topology both in X and Y.
+ * The interpretation of  wraparound for the cell based fields (e.g. H or DG) is
+ * quite straightforward. It is just that on the right edge, the right neighbours
+ * of the cells are the cells on the left edge of the domain.
```
jatkinson1000, line 51: Consider a diagram to be explicit?

```suggestion
 * of the cells are the cells on the left edge of the domain.
 * 
 * | 0 | 1 | 2 | . . . | n-1 | n | n+1 = 0
```

Also you state that this is in X and Y, so also upper neighbours of the top cells are the bottom cells in the domain? Maybe just be explicit about this too?

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+ *
+ * For point based fields there is only one caveat. The points on the right edge
+ * ARE the points on the left edge. There is continuity and they need to have
+ * the same identity.
+ *
```
jatkinson1000, line 56: I don't understand this.

Do you mean fields that are defined on vertices or edges?

To me this does not need saying as it is already included in the above definition and this just serves to confuse, but perhaps that's just me.

If you want you could just leave as "Note: that this means that the vertices and edges on the left boundary are identical to those on the right."

jatkinson1000: Ah, it relates to the below statement about number of cells. I think I would wrap this up in a "Note" rather than calling it a caveat. To me it is not a caveat, but follows from what you have already defined and is just a nuance to be aware of.

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+        , ny(ny)
+        , localNx(localNx)
+        , localNy(localNy)
+        , offsetX(offsetX)
+        , offsetY(offsetY)
+        , haloSize(1)
+        , type(type)
+    {
+        if (type == ModelArray::Type::VERTEX) {
+            //  Vertex fields have an extra entries (e.g. 3 cells in 1D have 4 vertices)
```
jatkinson1000, line 95: Ie, not periodic, per the above description? Maybe reinforce that?

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+                offset *= 10;
+            }
+            return offset;
+        }(domain_size);
+
+        return component_offset * (component + 1) + globalI * this->ny + globalJ;
+    }
+
+    /**
+     * @brief Map the local coorinates of the domain to the global coordinates
```
jatkinson1000, line 138:
```suggestion
     * @brief Map the local coordinates of the domain to the global coordinates
```

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+    verifyTestData(testData, DGVectorDomain);
+}
+
+MPI_TEST_CASE("CGVector", 3)
+{
+    auto& modelMPI = ModelMPI::getInstance();
+    auto& metadata = ModelMetadata::getInstance();
+
+    // We work on signed integers to represent extents and grid
+    // coordinates
```
jatkinson1000, line 367: Is this significant? Different to what neXtSIM usually does?

core/test/HaloExchangePB_test.cpp
jatkinson1000: From discussion:

For subdomains points on the left/bottom edge belong to the domain, and points on the right/top belong to the right/top-neighbouring domain.

For global periodic wrap-around means left/bottom edge is also the right/top.

But for non-periodic a separate top/right set of values have to be defined to set the boundary condition.

core/test/HaloExchangePB_test.cpp
jatkinson1000: From discussion, comment is perhaps stale and needs clarification for local subdomain which always includes halos, so is always n+1 vertices for n cells.

core/test/HaloExchangePB_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+ *  - We then perform tha halo exchange
+ *  - For each rank we verify that the values stored in the field match the
+ *    hash
+ *
+ * Test halo exchange class for Periodic Boundaries (PB) on following fields:
+ * - HField
+ * - VertexField
+ * - DGField
+ * - DGVector
+ * - CGVector
```
jatkinson1000, line 21: Clarify here or in test suite that the same test (3 subdomains) is run on each of these types?

jatkinson1000 (commented):
Hey @Mikolaj-A-Kowalski generally looking pretty good.

A couple of typos and some suggestions for clarifying the explanation (to me at least).

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: It is different with respect to the sibling CB version of the test that works on the unsigned integers, which I am feeling makes the global position calculations more difficult to perform. 

I will expand the comment to elaborate this.

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: I will try to reword it. The gist that I was trying to get across is that: 
- The hash value is dependent only on the position of the cell in the global coordinates and the component index
- We store the value of the hash in the field itself. But do not assign values in the halo regions
- We perform halo exchange 
- We can now go through each rank and compare the value in the field against the hash. If all matches we know that the exchange was correct and all values transferred to their correct location.

Mikolaj-A-Kowalski: Changed in e8c8a8cef78c

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: Changed in  e8c8a8cef78

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: Expanded this remark in 0b81fdccf7

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: Expanded in 7efcec3316

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: I have added the diagrams in 50f8dc7a7e68d04

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: I have tried to clarify these points in the expanded discussion in 50f8dc7a7e68d04 

Hopefully they become much cleared without the unnecessary new paragraph (that confused you) and more visual illustration with the diagrams.

Mikolaj-A-Kowalski:
@jatkinson1000 I have tried to address all of your comments. Please mark the threads as resolved if the changes fixes the points to your satisfaction ;-)

core/test/HaloExchangePB_test.cpp
jatkinson1000: I think that reads better, thanks. One last point to clarify - when verifying you do so by recomputing the hash based on global coordinate and checking it matches the value that was transferred during halo exchange? This last step wasn't explicitly clear.

core/test/HaloExchangePB_test.cpp
jatkinson1000: Yes, diagrams are great, thanks!

jatkinson1000 (approved):
Thanks @Mikolaj-A-Kowalski one last query, but I think this is good to go.

core/test/HaloExchangePB_test.cpp
Mikolaj-A-Kowalski: > One last point to clarify - when verifying you do so by recomputing the hash based on global coordinate and checking it matches the value that was transferred during halo exchange?

Exactly. I think the point that is not 100% clear is that we do it for all values in the field (halo and exterior).  This is to detect cases when e.g. we mess up the indexing and corrupt data in the interior. 

Will expand it in a sec.

Mikolaj-A-Kowalski: Added the motivation for checking the whole field.

Mikolaj-A-Kowalski:
> Thanks @Mikolaj-A-Kowalski one last query, but I think this is good to go.

@jatkinson1000  Reworded the remark one more time ;-) 
Also rebased on the current develop. If you are happy I will proceed and merge when the tests pass.

jatkinson1000:
Gah, fixups 😛 

LGTM, you may merge when ready.

PR #1043, opened 2026-02-10
Determine XIOS field types

joewallwork:
# Determine XIOS field types

Fixes #1026
~~Merges into #1042~~

### Task List
- [x] Defined the tests that specify a complete and functioning change (*It may help to create a [design specification & test specification](../../../wiki/Specification-Template)*)
- [x] Implemented the source code change that satisfies the tests
- [x] Documented the feature by providing worked example
- [x] Updated the README or other documentation
- [x] Completed the pre-Request checklist below

---
# Change Description

The crux of this PR is the following two changes:
* This PR sets core field types to be used by XIOS in the call to `PrognosticData::configure()`.
* If a field type isn't set then I've modified `Xios::getFieldType()` to return `ModelArray::Type::H`, which is the default.

To facilitate easy handling of field names, I also switched from repeated calls to getters to saving the sets of field names when XIOS is set up.

Some warnings are also added to make it clearer what's going on.

---
# Test Description

The XIOS test input file is modified so that `cice` is given by the `HField`
```
 cice =
  0, 1, 2, 3,
  4, 5, 6, 7 ;
```
However, this field is specified as being a `DGField` in `PrognosticData::configure()`, which is called by `Model::configure()`. Commenting out the lines in `XiosReadWriteRestart_test` that remove the output files produced by this test and interrogating them, I now find that
```
 cice =
  0, 0, 0, 0, 0, 0,
  1, 0, 0, 0, 0, 0,
  2, 0, 0, 0, 0, 0,
  3, 0, 0, 0, 0, 0,
  4, 0, 0, 0, 0, 0,
  5, 0, 0, 0, 0, 0,
  6, 0, 0, 0, 0, 0,
  7, 0, 0, 0, 0, 0 ;
```
as required.

---
# Documentation Impact

The XIOS docs page has been updated accordingly.

---
### Pre-Request Checklist

- [x] The requirements of this pull request are fully captured in an issue or design specification and are linked and summarised in the description of this PR
- [x] No new warnings are generated
- [x] The documentation has been updated (or an issue has been created to track the corresponding change)
- [x] Methods and Tests are commented such that they can be understood without having to obtain additional context
- [x] This PR/Issue is labelled as a bug/feature/enhancement/breaking change
- [x] This change conforms to the conventions described in the README

Issue #981, opened 2025-11-10
Align XIOS diagnostics approach with the rest of the model


joewallwork:
In particular, the `ConfigOutput` module.

Note that, unlike for `ParaGrid_test` (see #978), it seems like `ConfigOutput_test` is already laid out nicely for enabling XIOS.

PR #1050, opened 2026-02-24
XIOS: Avoid repeated getting of I/O field names configs

### LINKED ISSUE #1049: XIOS: Stash config rather than re-reading
Currently the XIOS handler class repeatedly parses the config to get fields for different I/O types. We should refactor this to read once and store the values for efficiency and ease.

joewallwork:
# XIOS: Avoid repeated getting of I/O field names configs

Fixes #1049
Refinement of #1043

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -52,13 +52,17 @@ static const std::string xInputPfx = "XiosInput";
 static const std::string xDiagnosticPfx = "XiosDiagnostic";
 static const std::string xForcingPfx = "XiosForcing";
 static const std::map<int, std::string> keyMap = { { Xios::ENABLED_KEY, "xios.enable" },
+    // TODO: Avoid having to parse restart fields (#1056)
     { Xios::OUTPUT_FIELD_NAMES_KEY, xOutputPfx + ".field_names" },
+    // TODO: Avoid having to parse input fields (#1056)
     { Xios::INPUT_FIELD_NAMES_KEY, xInputPfx + ".field_names" },
     { Xios::DIAGNOSTIC_PERIOD_KEY, xDiagnosticPfx + ".period" },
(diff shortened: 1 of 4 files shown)


TomMelt (approved):
Thanks @joewallwork LGTM

core/src/Xios.cpp
```diff
@@ -242,6 +241,42 @@ void Xios::configure()
...
+{
+    inputRestartFieldNames
+        = str2set(Configured::getConfiguration(keyMap.at(INPUT_FIELD_NAMES_KEY), std::string()));
+    outputRestartFieldNames
+        = str2set(Configured::getConfiguration(keyMap.at(OUTPUT_FIELD_NAMES_KEY), std::string()));
+    forcingFieldNames
+        = str2set(Configured::getConfiguration(keyMap.at(FORCING_FIELD_NAMES_KEY), std::string()));
+    diagnosticFieldNames = str2set(
+        Configured::getConfiguration(keyMap.at(DIAGNOSTIC_FIELD_NAMES_KEY), std::string()));
+
```
timspainNERSC, line 280: This is reading a list of field for each type from the config file, right? That's fine for now, but I don't want the user to have to guess what fields the model wants for every different configuration. And the set of output restart fields should certainly be determined only by the configuration of the model, so as to capture all the state required to perfectly restart the model.

timspainNERSC (approved):
This looks fine as it is, but there should be a strong TODO to not have the user have to specify the restart fields in the config file.

core/src/Xios.cpp
joewallwork: Indeed, these are temporary measures. I am currently working on avoiding these config sections in resolving #1056, #1052, and #981. Once done, the config file format will be the same as for running without XIOS.

joewallwork: Added some TODO notes in d8a84070.

PR #1051, opened 2026-02-24
Determine XIOS field types (take 3)

### LINKED ISSUE #1026: Implement interfaces to get restart field names out of modules

joewallwork:
# Determine XIOS field types (take 3)

Fixes #1026 
Refinement of #1043
~~Merges into #1050~~

### Task List

---
# Change Description

Main changes:
* Modify build system to allow including XIOS outside of `core/src`.
* Modify build system to make `ModelArray::AdvectionType` always available.
* Set core field types to be used by XIOS in the `IDynamics` and `IColumnPhysics` modules.
* Default field type to `ModelArray::Type::H`.

Some warnings are also added to make it clearer what's going on.

---
# Test Description

Some of the field names used in the XIOS tests needed tweaking to avoid conflicts between restarts and forcings.

---
# Documentation Impact

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ if(ENABLE_XIOS)
     target_link_libraries(nextsimlib PUBLIC xios HDF5::HDF5)
     target_include_directories(
         nextsimlib
-        PRIVATE ${xios_INCLUDES} ${xios_EXTERNS}/blitz ${xios_EXTERNS}/boost_extraction/include
+        PUBLIC ${xios_INCLUDES} ${xios_EXTERNS}/blitz ${xios_EXTERNS}/boost_extraction/include
                 ${xios_EXTERNS}/rapidxml/include
     )
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -839,6 +839,7 @@ void Xios::createField(const std::string& fieldId, const std::string& fileId)
         setFieldReadAccess(fieldId, false);
         fileAddField(fileId, fieldId);
     } else {
+        // TODO: Allow unused fields in input files (#1060)
         throw std::runtime_error("Xios: Unknown I/O type for field '" + fieldId + "'");
     }
 }
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -801,32 +801,6 @@ xios::CField* Xios::getField(const std::string& fieldId)
  */
 void Xios::createField(const std::string& fieldId, const std::string& fileId)
 {
-    int ioType = getFileIOType(fileId);
-
-    // Check that the field is in the expected config section
-    std::string ioName;
-    if (ioType == INPUT_RESTART
-        && inputRestartFieldNames.find(fieldId) == inputRestartFieldNames.end()) {
-        throw std::runtime_error(
-            "Xios: Field '" + fieldId + "' cannot be found in the input restart config section");
-    } else if (ioType == OUTPUT_RESTART
-        && outputRestartFieldNames.find(fieldId) == outputRestartFieldNames.end()) {
-        throw std::runtime_error(
-            "Xios: Field '" + fieldId + "' cannot be found in the output restart config section");
-    } else if (ioType == ERA5_FORCING
-        && era5ForcingFieldNames.find(fieldId) == era5ForcingFieldNames.end()) {
-        throw std::runtime_error(
-            "Xios: Field '" + fieldId + "' cannot be found in the ERA5 forcing config section");
-    } else if (ioType == TOPAZ_FORCING
-        && topazForcingFieldNames.find(fieldId) == topazForcingFieldNames.end()) {
-        throw std::runtime_error(
-            "Xios: Field '" + fieldId + "' cannot be found in the TOPAZ forcing config section");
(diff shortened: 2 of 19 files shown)


core/src/include/ModelArrayRef.hpp
```diff
@@ -327,6 +327,8 @@ template <const TextTag& fieldName> class ModelArrayRef<fieldName, RW> {
         outArr = dataReference->component(0) / divisor;
         return outArr;
     }
+    //! Returns the (enum of) the ModelArray::Type of the referenced ModelArray.
+    ModelArray::Type const getType() { return dataReference->getType(); };
```
joewallwork: This is required to get the advection type in the `IDynamics` module.

core/src/Xios.cpp
```diff
@@ -280,6 +280,7 @@ void Xios::parseConfig()
     fieldNames = inputRestartFieldNames;
     fieldNames.insert(outputRestartFieldNames.begin(), outputRestartFieldNames.end());
     fieldNames.insert(era5ForcingFieldNames.begin(), era5ForcingFieldNames.end());
+    fieldNames.insert(topazForcingFieldNames.begin(), topazForcingFieldNames.end());
```
joewallwork, line 283: This was missed in #1052.

CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ if(ENABLE_XIOS)
     target_link_libraries(nextsimlib PUBLIC xios HDF5::HDF5)
     target_include_directories(
         nextsimlib
-        PRIVATE ${xios_INCLUDES} ${xios_EXTERNS}/blitz ${xios_EXTERNS}/boost_extraction/include
+        PUBLIC ${xios_INCLUDES} ${xios_EXTERNS}/blitz ${xios_EXTERNS}/boost_extraction/include
```
joewallwork, line 148: This is needed to be able to `#include "includes/Xios.hpp"` outside of `core/src`.

core/src/include/ModelArrayRef.hpp
joewallwork: Actually it wasn't needed in the end.

core/src/Xios.cpp
```diff
@@ -803,29 +804,27 @@ void Xios::createField(const std::string& fieldId, const std::string& fileId)
...
-        fieldNames = diagnosticFieldNames;
-        ioName = "diagnostic";
-    } else {
-        throw std::runtime_error("Xios: Unknown I/O type for field '" + fieldId + "'");
-    }
-    if (fieldNames.find(fieldId) == fieldNames.end()) {
+    if (ioType == INPUT_RESTART
+        && inputRestartFieldNames.find(fieldId) == inputRestartFieldNames.end()) {
+        throw std::runtime_error(
+            "Xios: Field '" + fieldId + "' cannot be found in the input restart config section");
```
timspainNERSC: Does this means that if a field is present in the restart file that is not requested by the current configuration of the model Modules, then an error is raised? Because the model should definitely allow extra fields to exist in the restart file. For example the a restart file from a brittle rheology run, which contains stress component fields, could be used to initialize a visco-plastic rheology run, which will not ask for stress component fields. Logging a warning would certainly be reasonable.

joewallwork: No, these checks are just to ensure that if a field is being handled with a particular I/O type (e.g., restart) then it's included in the appropriate list of fields.

As it turns out, I just realised this code was actually redundant so I removed it in ed9e9c0f.

core/src/Xios.cpp
```diff
@@ -803,29 +804,27 @@ void Xios::createField(const std::string& fieldId, const std::string& fileId)
...
-            "Xios: Field '" + fieldId + "' cannot be found in the " + ioName + " config section");
+            "Xios: Field '" + fieldId + "' cannot be found in the output restart config section");
+    } else if (ioType == ERA5_FORCING
+        && era5ForcingFieldNames.find(fieldId) == era5ForcingFieldNames.end()) {
+        throw std::runtime_error(
+            "Xios: Field '" + fieldId + "' cannot be found in the ERA5 forcing config section");
+    } else if (ioType == TOPAZ_FORCING
+        && topazForcingFieldNames.find(fieldId) == topazForcingFieldNames.end()) {
+        throw std::runtime_error(
+            "Xios: Field '" + fieldId + "' cannot be found in the TOPAZ forcing config section");
```
timspainNERSC: As above, there might be fields in the forcing files that are not required by the model, which should not cause an error. This is especially true when using the real climatology files (as opposed to the current pre-processed files): there will be dozens of fields we are not interested in.

timspainNERSC (approved):
This looks good, but there should be a TODO where the XIOS subsystem checks that fields in the file have been specified in the configuration. Throwing an error at this point is too strict.

core/src/Xios.cpp
joewallwork: Thanks. As above, made redundant in https://github.com/nextsimhub/nextsimdg/commit/ed9e9c0f92915452630546f78a658ed55006a923.

joewallwork:
> This looks good, but there should be a TODO where the XIOS subsystem checks that fields in the file have been specified in the configuration. Throwing an error at this point is too strict.

Opened issue #1060 to track this and put in a TODO that references it in 9f7701bc.

PR #1055, opened 2026-03-02
Read forcings only when necessary

### LINKED ISSUE #1054: Avoid reading forcings every timestep
Currently, if the ERA5 and TOPAZ modules are included then they are read every timestep. This is unnecessary given that ERA5 data are only available once per hour and TOPAZ once per day.

joewallwork:
# Read forcings only when necessary

Fixes #1054

### Task List

---
# Change Description

Currently, if the ERA5 and TOPAZ modules are included then they are read every timestep. This is unnecessary given that ERA5 data are only available once per hour and TOPAZ once per day.

This PR makes it so that ERA5 forcings are only read at the top of the hour and TOPAZ forcings are only read at midnight.

## Caveats

1. Assumes that the simulation begins at midnight.
2. Assumes that timesteps are aligned with the hour.

Are we happy with these assumptions? If so, I can add config checks to ensure that they are satisfied. If not then I will have to modify how the XIOS calendar is handled.

(This change is also required to enable ERA5 and TOPAZ forcings with XIOS.)

---
# Test Description

XIOS tests are updated accordingly.

---
# Documentation Impact

A note on the assumptions has been added to the XIOS doc page.

### DIFF
--- a/physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
+++ b/physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
@@ -73,15 +73,18 @@ void ERA5Atmosphere::update(const TimestepTime& tst)
     std::set<std::string> forcings
         = { "tair", "dew2m", "pair", "sw_in", "lw_in", "wind_speed", "u", "v" };
 
-    ModelState state = ParaGridIO::readForcingTimeStatic(forcings, tst.start, filePath);
-    tairAccessor.getHostRW() = state.data.at("tair");
-    tdewAccessor.getHostRW() = state.data.at("dew2m");
-    pairAccessor.getHostRW() = state.data.at("pair");
-    sw_inAccessor.getHostRW() = state.data.at("sw_in");
-    lw_inAccessor.getHostRW() = state.data.at("lw_in");
-    windAccessor.getHostRW() = state.data.at("wind_speed");
-    uwindAccessor.getHostRW() = state.data.at("u");
-    vwindAccessor.getHostRW() = state.data.at("v");
+    // Read ERA5 forcings at the top of the hour
+    if (std::fmod((tst.start - TimePoint()).seconds(), 3600.) == 0.) {
+        forcingState = ParaGridIO::readForcingTimeStatic(forcings, tst.start, filePath);
+    }
+    tairAccessor.getHostRW() = forcingState.data.at("tair");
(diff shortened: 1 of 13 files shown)


physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
```diff
@@ -79,15 +79,18 @@ void ERA5Atmosphere::update(const TimestepTime& tst)
...
-    tdew = state.data.at("dew2m");
-    pair = state.data.at("pair");
-    sw_in = state.data.at("sw_in");
-    lw_in = state.data.at("lw_in");
-    wind = state.data.at("wind_speed");
-    uwind = state.data.at("u");
-    vwind = state.data.at("v");
+    // Read ERA5 forcings at the top of the hour
+    if (tst.start.format(TimePoint::msFormat) == "T00:00Z") {
+        forcingState = ParaGridIO::readForcingTimeStatic(forcings, tst.start, filePath);
```
timspainNERSC: To avoid string parsing every time, I suggest getting the duration since the epoch in seconds and  seeing whether it is the top of the hour:
`(tst.start - TimePoint()).seconds() % 3600. == 0.`
(This should work: if it doesn't that's a bug in the Time classes).

timspainNERSC (changes requested):
This will be fine. There is an assumption that the time step length will either evenly divide an hour or be a multiple number of hours that evenly divide the day. But this seems to be a reasonable assumption.

I'd recommend not printing the time string every time you want to check for fresh data, though.

joewallwork:
[Rebased on top of `develop` after merging #1051]

physics/src/modules/AtmosphereBoundaryModule/ERA5Atmosphere.cpp
joewallwork: Done in https://github.com/nextsimhub/nextsimdg/pull/1055/commits/74d78916cf1ca5a8621a1ba6a4267783d7b86092. Similarly for TOPAZ in [ac44e60](https://github.com/nextsimhub/nextsimdg/pull/1055/commits/ac44e60d942c3bdb358e64978b414ef92422e707).

TomMelt (approved):
Thanks @joewallwork . It looks good to me

timspainNERSC (approved):
Looking good, now

PR #1057, opened 2026-03-03
XIOS: read all inputs

joewallwork:
# XIOS: read all inputs

Towards #1056

### Task List
- [x] Linked an issue above that captures the requirements of this PR
- [x] Defined the tests that specify a complete and functioning change
- [x] Implemented the source code change that satisfies the tests
- [x] Commented all code so that it can be understood without additional context
- [x] No new warnings are generated or they are mentioned below
- [x] The documentation has been updated (or an issue has been created to do so)
- [x] Relevant labels (e.g., enhancement, bug) have been applied to this PR
- [x] This change conforms to the conventions described in the README

---
# Change Description

This PR avoids the need for an `XiosInput` configuration section. Instead, XIOS reads all field variables that it finds in any input files whose names are provided.

---
# Test Description

XIOS tests are updated to account for this change. Some tweaks were required to avoid the same variable name (`u`) being read from both restarts and forcings.

---
# Documentation Impact

The XIOS docs page has been update accordingly.

PR #1063, opened 2026-03-13
Drop `XiosOutput` section

### LINKED ISSUE #1056: Avoid `XiosInput` and `XiosOutput` config sections
In #1052, the `XiosForcing` config section gets removed, with forcings being configured in the `ERA5Atmosphere` and `TOPAZOcean` modules, as with the existing I/O approach.

With #981, the `XiosDiagnostic` config section will be removed, with diagnostics being configured in the `ConfigOutput` module.

We should similarly drop the `XiosInput` and `XiosOutput` sections to align the user experience when building with and without XIOS. For the inputs, we can probably just load all the fields that are found in the input file, but for outputs it's not immediately obvious what to do.

joewallwork:
# Drop `XiosOutput` section

Fixes #1056
Merges into #1057

### Task List

---
# Change Description

This PR allows us to drop the temporary `XiosOutput` config section. In order to achieve this, I needed to ensure that each module that defines a `getStatePrognostic` member function also sets XIOS field types appropriately.

---
# Test Description

I noticed that some fields were slipping through the cracks in the XIOS tests so I made them more robust. That is, I added checks that the expected fields are all found in any input files.

I also dropped `damage` from the test input file to check that it gets set up automatically during model configuration, as expected.

---
# Documentation Impact

### DIFF
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -134,7 +134,6 @@ Xios::~Xios() { finalize(); }
 void Xios::close_context_definition()
 {
     if (isEnabled && contextStatus == DEFINITION_OPEN) {
-        // TODO: Check moving this here doesn't break anything else
         setupFiles();
 
         // Special handling of input fields
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -284,6 +284,23 @@ void Xios::parseConfig()
     // TODO: Avoid having to parse diagnostic fields for XIOS specifically (#981)
     diagnosticFieldNames = str2set(
         Configured::getConfiguration(keyMap.at(DIAGNOSTIC_FIELD_NAMES_KEY), std::string()));
+
+    // Ensure the coordinate variables are included in diagnostic files
+    if (spherical) {
+        if (diagnosticFieldNames.count(longitudeName) == 0) {
+            diagnosticFieldNames.insert(longitudeName);
+        }
+        if (diagnosticFieldNames.count(latitudeName) == 0) {
+            diagnosticFieldNames.insert(latitudeName);
+        }
+    } else {
+        if (diagnosticFieldNames.count(xName) == 0) {
+            diagnosticFieldNames.insert(xName);
+        }
+        if (diagnosticFieldNames.count(yName) == 0) {
+            diagnosticFieldNames.insert(yName);
+        }
+    }
 }
 
 //! Initialize the XIOS context with ID contextId
--- a/core/src/Xios.cpp
+++ b/core/src/Xios.cpp
@@ -49,8 +49,6 @@ namespace Nextsim {
 
 static const std::string xDiagnosticPfx = "XiosDiagnostic";
 static const std::map<int, std::string> keyMap = { { Xios::ENABLED_KEY, "xios.enable" },
-    // TODO: Avoid having to parse restart fields (#1056)
-    { Xios::OUTPUT_FIELD_NAMES_KEY, "XiosOutput.field_names" },
     { Xios::DIAGNOSTIC_PERIOD_KEY, xDiagnosticPfx + ".period" },
     { Xios::DIAGNOSTIC_FILE_KEY, xDiagnosticPfx + ".filename" },
     // TODO: Avoid having to parse diagnostic fields for XIOS specifically (#981)
@@ -66,10 +64,6 @@ Xios::HelpMap& Xios::getHelpText(HelpMap& map, bool getAll)
             "to be modifed by the user. Build nextSIM-DG with XIOS support with the CMake argument "
             "-DENABLE_XIOS=ON, passing the path to your XIOS installation with "
             "-Dxios_DIR=/path/to/xios." },
-    };
-    map["XiosOutput"] = {
-        { keyMap.at(OUTPUT_FIELD_NAMES_KEY), ConfigType::STRING, {}, "", "",
-            "Comma-separated list of field names to be written to the output file." },
     };
     map["XiosDiagnostic"] = {
         { keyMap.at(DIAGNOSTIC_PERIOD_KEY), ConfigType::STRING, {}, "0", "",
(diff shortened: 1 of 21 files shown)


core/src/modules/include/IDynamics.hpp
```diff
@@ -53,11 +53,16 @@ class IDynamics : public ModelComponent {
...
+        // Advected fields
         xiosHandler.setPrognosticFieldType(hiceName, ModelArray::AdvectionType);
         xiosHandler.setPrognosticFieldType(ciceName, ModelArray::AdvectionType);
-        xiosHandler.setPrognosticFieldType(damageName, ModelArray::AdvectionType);
+        xiosHandler.setPrognosticFieldType(damageName, ModelArray::Type::H);
         xiosHandler.setPrognosticFieldType(hsnowName, ModelArray::AdvectionType);
+
+        // FIXME: We need to write out u and v to get a perfect restart
+        // xiosHandler.setPrognosticFieldType(uName, ModelArray::Type::H);
+        // xiosHandler.setPrognosticFieldType(vName, ModelArray::Type::H);
```
joewallwork: Note to self to complete this before marking the PR as ready for review.

core/src/modules/include/IStructure.hpp
```diff
@@ -27,7 +31,24 @@ namespace Nextsim {
...
 public:
-    IStructure() { }
+    IStructure()
+    {
+#ifdef USE_XIOS
+        // Set XIOS field types for core grid-related fields
+        Xios& xiosHandler = Xios::getInstance();
+        xiosHandler.setPrognosticFieldType(coordsName, ModelArray::Type::VERTEX);
+        xiosHandler.setPrognosticFieldType(gridAzimuthName, ModelArray::Type::H);
+        const bool spherical = true; // TODO: determine this
```
joewallwork: Note to self to do this before marking the PR as ready for review.

core/src/Xios.cpp
```diff
@@ -140,6 +134,8 @@ Xios::~Xios() { finalize(); }
 void Xios::close_context_definition()
 {
     if (isEnabled && contextStatus == DEFINITION_OPEN) {
+        // TODO: Check moving this here doesn't break anything else
```
joewallwork: Note to self to do this before marking the PR as ready for review.

joewallwork: Checked in the unit and integration tests. Docs updated in 2209e61e.

core/src/modules/StructureModule/include/ParametricGrid.hpp
```diff
@@ -24,6 +24,21 @@ class ParametricGrid : public IStructure {
...
+        xiosHandler.setPrognosticFieldType(coordsName, ModelArray::Type::VERTEX);
+        xiosHandler.setPrognosticFieldType(gridAzimuthName, ModelArray::Type::H);
+        const bool spherical = true; // TODO: Determine this
+        if (spherical) {
+            xiosHandler.setPrognosticFieldType(latitudeName, ModelArray::Type::H);
+            xiosHandler.setPrognosticFieldType(longitudeName, ModelArray::Type::H);
+        } else {
+            xiosHandler.setPrognosticFieldType(xName, ModelArray::Type::H);
+            xiosHandler.setPrognosticFieldType(yName, ModelArray::Type::H);
+        }
```
joewallwork: Question for reviewer: I see that there's an enum for this in `ParametricMesh` but not in `ParametricGrid`. It feels like it would make sense to align this - would that be agreeable? The difficulty I'm having is that I want to determine the coordinate system upon constructing `ParametricGrid`, when there likely isn't a `ModelState` I can query.

core/src/modules/include/IStructure.hpp
joewallwork: Deferred to https://github.com/nextsimhub/nextsimdg/pull/1063/files/c2d9c6144f76e416a55139fb5f05b84b21cec693#diff-6b00b9e8a65c8ffd19e30c6df0373155cda7330af427f58a8ccdaf44c582ab11.

core/src/modules/include/IDynamics.hpp
joewallwork: The issue I was having here has to do with the `u` and `v` fields being expected to be type `Model::Array::U` and `Model::Array::V`, which I hadn't accounted for. Working on a fix.

@timspainNERSC: It seems to me that these types are duplicates of `Model::Array::H`. Is that the case or are there distinguishing factors?

timspainNERSC: In the current set up, these are indeed duplicates of `HField`. The design exists to be able to extend nextSIM-DG with other dynamics cores (such as the Lagrangian) where the u and/or v arrays might have different dimensions from the h arrays.

joewallwork: Okay good to know, thanks. I think it would be best to do this separately from this PR. I opened issue #1066 for tackling this and added a reference in the code.

joewallwork: Should be fixed by #1068.

core/src/modules/StructureModule/include/ParametricGrid.hpp
timspainNERSC: The sphericity versus planarity of the coordinate system is currently determined by the restart file. A solution would be to peek inside the restart file to check if there is a `longitude` field (=> spherical coordinates) and if there is not, an `x` field, the lack of which would be an error. The `ModelMetadata` class does something similar to get the dimensions around `ModelMetadata.cpp:200`.

core/src/include/Model.hpp
```diff
@@ -65,6 +65,9 @@ class Model : public Configured<Model> {
     Iterator iterator;
     DevStep modelStep; // Change the model step calculation here
     PrognosticData pData;
+
+    bool configuredRestarts = false; // Guard to avoid repeated configuration
+    bool configuredTime = false; // Guard to avoid repeated configuration
```
timspainNERSC, line 70: Is XIOS configuration not idempotent? The rest of the model can be configured multiple times without ill effect.

timspainNERSC (approved):
This all seems in order.

core/src/include/Model.hpp
joewallwork: Unfortunately not. Parts of it could be made so if we wanted that but the context is problematic because it cannot be overwritten.

core/src/modules/StructureModule/include/ParametricGrid.hpp
joewallwork: Thanks for the suggestion. Done in 1f33b5e5.

joewallwork: Also throw the recommended error in 88a9d749.

PR #1074, opened 2026-05-06
Integrate interpolation support from XIOS

### LINKED ISSUE #979: XIOS: Support forcing files on different grids
@timspainNERSC is writing a new implementation of forcings that allows for reading from input files on different grids to that used in the model and interpolating.

We should be able to implement this using XIOS' `interpolation_domain` and `interpolation_axis` concepts by providing the domain/axis information for each grid, as well as the regridding weights.

TomMelt:
# Integrate interpolation support from XIOS

Fixes #979

## Open Mailing list query with XIOS devs

- I have created a MWE repo [here](https://github.com/TomMelt/xios-interp)
  - There is a PR here which implements interpolation https://github.com/TomMelt/xios-interp/pull/1
- I have emailed XIOS user mailing list for support (see [here](https://listes.ipsl.fr/sympa/arc/xios-users/2026-06/msg00000.html))
- I believe I have implemented all necessary features but I cannot get interpolation working

---

### Task List
- [ ] Commented all code so that it can be understood without additional context
- [ ] No new warnings are generated or they are mentioned below
- [ ] The documentation has been updated (or an issue has been created to do so)

---
# Change Description

This PR exposes the interpolation functionality of `XIOS` into `nextsim`.

**Plan**
- Add new ERA5 fields which exist on the "Forcing" grid/domain
  - add a `field_ref` to this new field in the main nextsim code 
```xml
<field id="tairEra5" grid_ref="HGridEra5" />
<field id="tair" field_ref="tairEra5" grid_ref="HGridEra5Interp" /> <!-- done at runtime instead -->
```  
- Add 2 new grids `HGridEra5` and `HGridEra5Interp`
  - `HGridEra5Interp` will be same size as nextsim grid, but has the `<interpolate_domain/>` property set
  - because `tair` references field `tairEra5` but exists on `grid_ref="HGridEra5Interp"` XIOS will interpolate at runtime
```xml
<grid id="HGridEra5">
  <domain domain_ref="era5"/>
</grid>

<grid id="HGridEra5Interp" name="HGridEra5Interp">
  <domain domain_ref="era5Interp">
    <interpolate_domain/>
  </domain>
</grid>
```
- Add 2 new domains `era5` and `era5Interp`
  - `era5Interp` will be the same size as the nextsim model domain, and `era5` will match the Forcing data
  - we need to set name so that XIOS reads correct dims (I believe dims are labelled `[dim_name]_[domain_name]` which means if we didn't change the name manually it would be the same as the `domain id` leading to `x_era5` rather than `x_dim`
  - We need 2 domains because we need one for the Forcing file grid, and the second in theory matches the size of the nextsim grid... but... it has the interpolate property set to true, which would break the existing nextsim grids which don't need interpolation e.g., restart files and initialization files.
```xml
<domain id="era5" name="dim" ni_glo="3" nj_glo="5" type="rectilinear" />
<domain id="era5Interp" name="dim" type="rectilinear" /> <!-- details set at runtime -->
```

---
# Test Description

- `XiosForcingInterpolation_test.cpp` tests `XIOS` interpolation from a low resolution grid (from file) to the higher resolution grid (in the test)   

---
# Documentation Impact

To be completed

---
# Other Details

N/A

### DIFF
--- a/core/src/include/Xios.hpp
+++ b/core/src/include/Xios.hpp
@@ -129,6 +129,8 @@ private:
 
     /* Axis */
     xios::CAxis* getAxis(const std::string& axisId);
+    const std::string era5interpId = "era5interp";
+    const std::string era5HGridHR = "HGridEra5HR";
 
     /* Domain */
     // NOTE: Dimension names get processed as <dim>_<domainId> by XIOS, so we define the domainIds
@@ -144,6 +146,9 @@ private:
         { ModelArray::Type::VERTEX, "vertex" },
         // CG-based x- and y-dimensions (alt. names x_cg and y_cg)
         { ModelArray::Type::CG, "cg" },
+    };
+    std::map<ModelArray::Type, std::string> EAR5domainIds = {
+        { ModelArray::Type::H, "dim" },
     };
     std::map<std::string, bool> domainWritten = {
         { "dim", false },
@@ -164,6 +169,9 @@ private:
         { ModelArray::Type::DGSTRESS, "DGSGrid" },
         { ModelArray::Type::VERTEX, "VertexGrid" },
         { ModelArray::Type::CG, "CGGrid" },
+    };
+    std::map<ModelArray::Type, std::string> ERA5GridIds = {
+        { ModelArray::Type::H, "HGridEra5HR" },
     };
     void setupGrids();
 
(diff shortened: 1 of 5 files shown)


core/src/include/Xios.hpp
```diff
@@ -145,6 +147,9 @@ class Xios : public Configured<Xios> {
         // CG-based x- and y-dimensions (alt. names x_cg and y_cg)
         { ModelArray::Type::CG, "cg" },
     };
+    std::map<ModelArray::Type, std::string> EAR5domainIds = {
```
timspainNERSC, line 150: 👂

PR #1080, opened 2026-06-01
Issue1079 mixedprec

### LINKED ISSUE #1079: Single/mixed precision support
Small scale experiments indicate that it might be possible to use single precision throughout the model [1].
On GPU the problem size is limited by memory requirements which would be cut in half and there should be a significant performance boost.

[1] Jendersie, R., Lessig, C., & Richter, T. (2025). A GPU parallelization of the neXtSIM-DG dynamical core (v0. 3.1). Geoscientific Model Development, 18(10), 3017-3040.

Thanduriel:
# Single precision support

Fixes #1079, #1083

---
# Change Description

Added a build-time option to switch to `float` (single precision) for all computations. The generic type `FloatType`, previously used inconsistently and in the dynamics only, was introduced everywhere to replace `double`. To make it compile, some double literals had to be wrapped (e.g. `FloatType(0.5)`). There are likely more places where double literals cause unnecessary promotion of an operation to `double`.

Some bugs, unrelated to the float type, where fixed while debugging single precision:

* Kokkos with a host backend was broken because of incorrect buffer lifetimes
* `ssh` field was overwritten in the Kokkos kernel with 0 before computations
* `SlabOcean` computation was incorrect due to parentheses placement 
* `SlabOcean` test initialization was missing some fields since the introduction of Accessors
* kokkos host backend could not handle reallocation the underlying ModelArray buffer which happens in some tests

---
# Test Description
The regular tests all pass when built in single precision. To achieve this, some error tolerances had to be adjusted based on the floating point type because of the reduced accuracy of `float`.  More substantial changes where needed in the `slabOcean`, where the reference computations had to be updated to match the fixed equation in the implementation. @timspainNERSC Can you have a look again at these changes in the test?

Both the benchmark and the NH case where run for a few simulated days in single precision and look fine.

---
# Documentation Impact
This feature is mentioned in the Installation/GPU build section. 

---
# Other Details


einola (dismissed):
This looks fine. A lot of modifications, so it's hard to review thoroughly. I also don't understand why it solves #1082 - but it does.

einola (changes requested):
I had to make some changes to get this to compile with clang + kokkos. They just consist of adding some FloatType casting of hard-coded numbers. Can you check my two commits?

Also, should we add something like

```
constexpr FloatType ft0 = 0
constexpr FloatType ft1 = 1
```

etc to FloatType.hpp and use ``ft0`` et al instead of ``FloatType(0)``?

Thanduriel:
> I had to make some changes to get this to compile with clang + kokkos. They just consist of adding some FloatType casting of hard-coded numbers. Can you check my two commits?

I went over the code again and found a few places where the cast happened too late, causing an unnecessary promotion to double in between. There are probably more of these in the code but it is not a correctness issue.
> 
> Also, should we add something like
> 
> ```
> constexpr FloatType ft0 = 0
> constexpr FloatType ft1 = 1
> ```
> 
> etc to FloatType.hpp and use `ft0` et al instead of `FloatType(0)`?

I have introduced a user defined literal operator instead. This also makes the syntax a bit less verbose and it can be used with any literals:
```C++
const FloatType a = 0.5_ft * b;
```

einola (approved):
Very nice, and the literal operator is a nice touch!

dynamics/src/kokkos/KokkosBrittleCGDynamicsKernel.cpp
```diff
@@ -5,6 +5,7 @@
 #include "include/KokkosBrittleCGDynamicsKernel.hpp"
 #include "../../../core/src/kokkos/include/KokkosTimer.hpp"
 #include "include/KokkosDGLimit.hpp"
+#include <cfenv> // debugging
```
timspainNERSC: Is this include still required?

timspainNERSC (approved):
A left-in debugging include file?
Otherwise, it all looks fine.

Issue #1090, opened 2026-07-23
NetCDF test files not generated when providing Python executable as CMake option


jatkinson1000:
2 files (`core/test/alt_names.nc` and `core/test/ParaGridIO_input_test.nc`) are not generated from their associated Python scripts when `PYTHON_EXECUTABLE` is passed as a CMake option.

I believe this is because the CMake option is not being correctly converted to `Python_EXECUTABLE` within the core `CMakeLists.txt`.

This results in failure to test `testParaGrid_MPI2` due to inability to read/open/find those NetCDF files.

PR #1091, opened 2026-07-23
Fix typo in CMakeLists to set Python_EXECUTABLE

jatkinson1000:
# Fix typo in CMakeLists to set Python_EXECUTABLE

Fixes No #1090 

### Task List
- [-] Defined the tests that specify a complete and functioning change
- [-] Commented all code so that it can be understood without additional context
- [-] No new warnings are generated or they are mentioned below
- [-] The documentation has been updated (or an issue has been created to do so)

---
# Change Description

Currently the CMakeLists doesn't correctly parse the command-line option for specifying PYTHON_EXECUTABLE as described in #1090.

An indication of this was the CMake warning generated by the current code:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    PYTHON_EXECUTABLE
```

This is because it was not being parsed as a variable when setting the internal `Python_EXECUTABLE`.

As a result two NetCDF files did not get generated in `core/test/`

---
# Test Description

Previously `testParaGrid_MPI2` failed due to not locating those files.
It now passes.

---
# Documentation Impact

None

---
# Other Details

None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ target_link_libraries(nextsimlib PUBLIC Eigen3::Eigen)
 
 if(DEFINED PYTHON_EXECUTABLE)
     set(Python_EXECUTABLE
-        PYTHON_EXECUTABLE
+        ${PYTHON_EXECUTABLE}
         CACHE STRING "")
 else()
     set(Python_FIND_VIRTUALENV FIRST)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.15)
 
 # set globally because we have a number of independent targets
 set(CMAKE_CXX_STANDARD 17)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.16)
 
 # set globally because we have a number of independent targets
 set(CMAKE_CXX_STANDARD 17)
@@ -63,14 +63,8 @@ if(NOT TARGET Eigen3::Eigen)
 endif()
 target_link_libraries(nextsimlib PUBLIC Eigen3::Eigen)
 
-if(DEFINED PYTHON_EXECUTABLE)
-    set(Python_EXECUTABLE
-        ${PYTHON_EXECUTABLE}
-        CACHE STRING "")
-else()
-    set(Python_FIND_VIRTUALENV FIRST)
-    find_package(Python COMPONENTS Interpreter)
-endif()
+set(Python_FIND_VIRTUALENV FIRST)
+find_package(Python COMPONENTS Interpreter REQUIRED)
 
 # The location of the module_builder.py scripts
 set(ScriptDirectory "${PROJECT_SOURCE_DIR}/scripts")
--- a/core/test/CMakeLists.txt
+++ b/core/test/CMakeLists.txt
@@ -112,7 +112,8 @@ if(ENABLE_MPI)
         add_custom_command(
             OUTPUT iodef.xml
             COMMAND
-                python3 ${CMAKE_CURRENT_SOURCE_DIR}/../src/generate_iodef.py ${DGComp}
+                "${Python_EXECUTABLE}"
+                ${CMAKE_CURRENT_SOURCE_DIR}/../src/generate_iodef.py ${DGComp}
                 ${DGStressComp} ${CMAKE_CURRENT_SOURCE_DIR}/../src/iodef.xml.jinja
                 ${CMAKE_CURRENT_BINARY_DIR}/iodef.xml
             DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/../src/iodef.xml.jinja)
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -125,6 +125,13 @@ required Python packages with
 
         pip install -r requirements.txt
 
+.. note::
+
+   The build system locates the Python interpreter using CMake's ``FindPython``
+   module, preferring the interpreter of any active ``conda`` or ``virtualenv``
+   environment. If you need to point the build at a specific Python interpreter,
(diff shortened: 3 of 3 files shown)


jatkinson1000:
Discussion points:

- Do we want quotes in there as well for better practice?
- Elsewhere in the file uses `python3` directly, perhaps we should instead standardise?
- Why is `PYTHON_EXECUTABLE` mapped to `Python_EXECUTABLE` instead of passing in `Python_EXECUTABLE` directly?

TomMelt:
> Discussion points:
> 
>     * Do we want quotes in there as well for better practice?

Sorry, what do you mean by this? Can you provide more context please.

>     * Elsewhere in the file uses `python3` directly, perhaps we should instead standardise?

Yes, absolutely. We should use `${Python_EXECUTABLE}` every where. I can see `python3` in `core/test/CMakeLists.txt:110` did you find any others? Please feel free to fix any and add them to this PR :+1: 

>     * Why is `PYTHON_EXECUTABLE` mapped to `Python_EXECUTABLE` instead of passing in `Python_EXECUTABLE` directly?

The reason is that in `CMakeLists.txt` we use `find_package(Python COMPONENTS Interpreter)`, but if this doesn't work, or, the user wants to manually specify a different python, which isn't the first one found in their environment, then we allow the user to specify `-DPYTHON_EXECUTABLE`. The other one (`Python_EXECUTABLE`) is defined by CMake. So it would be overridden. 
https://github.com/nextsimhub/nextsimdg/blob/f063e57e71d64c495f9aa2a6531947ac8f0c5473/CMakeLists.txt#L65-L72

TomMelt (approved):
Thanks @jatkinson1000 for spotting this. I am happy with this PR, but as you suggested please feel free to fix the other python3 references and then we can go ahead and merge 🚀

jatkinson1000:
> > ```
> > * Do we want quotes in there as well for better practice?
> > ```
> 
> Sorry, what do you mean by this? Can you provide more context please.
> 

Some prefer `"${PYTHON_EXECUTABLE}"` in case anyone is passing a filepath with spaces in.

A quick `find | grep` indicates this is the only offender.
Addressed in https://github.com/nextsimhub/nextsimdg/pull/1091/commits/12ffa7bb532081b0e60aad42d3375a456e34c64a


> The reason is that in `CMakeLists.txt` we use `find_package(Python COMPONENTS Interpreter)`, but if this doesn't work, or, the user wants to manually specify a different python, which isn't the first one found in their environment, then we allow the user to specify `-DPYTHON_EXECUTABLE`. The other one (`Python_EXECUTABLE`) is defined by CMake. So it would be overridden.
> 

Makes sense, but in that case, since we require CMake version 3.12, why not remove the if statement and just rely on `find_package(Python COMPONENTS Interpreter)` (with addition of `REQUIRED`?)? They will need the venv Python anyway for the dependencies required by the scripts.
I would argue that if it has taken this long to spot the bug then people are not commonly using this and, moreover, in the current setup it was ignored/unused anyway.

CMakeLists.txt
Mikolaj-A-Kowalski: Unrelated directly to this PR.

But since we are looking at Python CMake setup anyway. 

The [doc](https://cmake.org/cmake/help/latest/module/FindPython.html) says that `Python_FIND_VIRTUALENV` has been added in CMake 3.15. 

Since we require only 3.12 there is a chance this will be silently ignored I think.

Mikolaj-A-Kowalski (commented):
> Makes sense, but in that case, since we require CMake version 3.12, why not remove the if statement and just rely on find_package(Python COMPONENTS Interpreter) (with addition of REQUIRED?)? They will need the venv Python anyway for the dependencies required by the scripts.
I would argue that if it has taken this long to spot the bug then people are not commonly using this and, moreover, in the current setup it was ignored/unused anyway.

Myself I am inclined to agree with this. [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html) provides the `Python_ROOT_DIR` variable that can be provided both via env variable (or CLI with `-DPython_ROOT_DIR`). When I played with it just now it seems to work well. It will only be a question of documenting this to a user (in README perhaps?). 

The main reason I prefer it over the existing solution is that it feel much more 'idiomatic' and might be less likely to cause unexpected problems down the line 🤞

CMakeLists.txt
Mikolaj-A-Kowalski: Bump minimum CMake to 3.15? The question is if there is anybody who relies on the 3.12-3.14?

TomMelt: @jatkinson1000 I wont add noise by adding a review. I agree with @Mikolaj-A-Kowalski 's points. So once you have done that and added the quotes, I am also happy for you to merge :rocket:

jatkinson1000: https://github.com/nextsimhub/nextsimdg/pull/1091/commits/df29abea5d51e7157dcbfdbea49a7695221d4e7f Bumps to CMake version 3.15

jatkinson1000:
@Mikolaj-A-Kowalski I think this should also have the proposed change from you and I above to remove conditional and just rely on `find_package(Python COMPONENTS Interpreter)` with `REQUIRED` and remove the specification of `PYTHON_EXECUTABLE` as a input arg?

If you are happy and in agreement I will make that change.

Mikolaj-A-Kowalski:
~~@jatkinson1000 Thanks for the final touches!~~

~~I would say it is good to go after CI passes.~~

Mikolaj-A-Kowalski:
> @Mikolaj-A-Kowalski I think this should also have the proposed change from you and I above to remove conditional and just rely on find_package(Python COMPONENTS Interpreter) with REQUIRED and remove the specification of PYTHON_EXECUTABLE as a input arg?
>
>If you are happy and in agreement I will make that change.

Sorry about the comment above. I was trying to be too quick. 

I think it is indeed a good idea to do it if it is not much trouble and you have time. The only issue there is searching the docs and update it accordingly with correct instructions. But to be fair it is a perfect task for an LLM so it should be ok. 

So do it if you have time? Otherwise I also think it is ok to merge this one and sort out the more idiomatic way in the followup.

jatkinson1000:
@Mikolaj-A-Kowalski I think that change is now made so ready for re-review.

Mikolaj-A-Kowalski (approved):
Looks good to me! 

I think we can merge.

PR #1093, opened 2026-07-28
bug: fix intel compiler error for invalid conversion

### LINKED ISSUE #1092: invalid narrowing conversion from "int" to "unsigned long"
When building with intel's oneapi compiler `intel-oneapi-compilers@2022.1.0` I get a compiler error: 

```
/home/tdm39/nextsimdg/core/src/RectGridIO.cpp(92): error #3291: invalid narrowing conversion from "int" to "unsigned long"
      const std::vector<size_t> size = { metadata.getLocalExtentY(), metadata.getLocalExtentX() };
``` 

Also newer gnu compilers `gcc >= 16` will issue a warning:

```
/home/melt/sync/cambridge/projects/current/sasip/nextsimdg/core/src/RectGridIO.cpp: In member function ‘virtual
 Nextsim::ModelState Nextsim::RectGridIO::getModelState(const std::string&)’:
 /home/melt/sync/cambridge/projects/current/sasip/nextsimdg/core/src/RectGridIO.cpp:91:65: warning: narrowing conversion of ‘(&
 metadata)->Nextsim::ModelMetadata::getLocalCornerY()’ from ‘int’ to ‘long unsigned int’ [-Wnarrowing]
    91 |     const std::vector<size_t> start = { metadata.getLocalCornerY(), metadata.getLocalCornerX() };
       |                                         ~~~~~~~~~~~~~~~~~~~~~~~~^~
```  

The integers are always positive, so it should be fine to cast them to size_t.

TomMelt:
Fixes #1092 

Make explicit conversion from `int` to `size_t`.

### DIFF
--- a/core/src/RectGridIO.cpp
+++ b/core/src/RectGridIO.cpp
@@ -88,8 +88,10 @@ ModelState RectGridIO::getModelState(const std::string& filePath)
     // Set the origins and extensions for reading 2D data based
     // on MPI decomposition
     const auto& metadata = ModelMetadata::getInstance();
-    const std::vector<size_t> start = { metadata.getLocalCornerY(), metadata.getLocalCornerX() };
-    const std::vector<size_t> size = { metadata.getLocalExtentY(), metadata.getLocalExtentX() };
+    const std::vector<size_t> start = { static_cast<size_t>(metadata.getLocalCornerY()),
+        static_cast<size_t>(metadata.getLocalCornerX()) };
+    const std::vector<size_t> size = { static_cast<size_t>(metadata.getLocalExtentY()),
+        static_cast<size_t>(metadata.getLocalExtentX()) };
 #else
     const std::vector<size_t> start = { 0, 0 };
     const std::vector<size_t> size(ModelArray::dimensions(ModelArray::Type::H).rbegin(),


jatkinson1000 (approved):
This seems reasonable to me @TomMelt based on the error you reported.
We have seen similar in FTorch in the past I think.

For disclosure I have not checked it builds for me, but trust you (and any CI).

Not sure if this is still a draft as you requested a review, so will approve, but lmk if there is more you plan to add.

Mikolaj-A-Kowalski:
This PR made me wonder and I have noticed that we are compiling nextsim only with the default warnings... 

I think the reason newer GCC behaviour is that the `-Wnarrowing` got moved  from an optional to the default warnings lists. 

When I tried  to compile `develop` with `-Wall` (GCC 13) it produces a rather long list of warnings [warnings.log](https://github.com/user-attachments/files/30497057/warnings.log). It is something to potentially discuss on Monday and open an issue about. 

WRT this PR I am same opinion as Jack. It can be just merged in.

Mikolaj-A-Kowalski (approved):

TomMelt:
Thanks @jatkinson1000 and @Mikolaj-A-Kowalski for the review. I will merge it in :rocket:

PR #1105, opened 2026-08-10
Issue1088 slab ocean bug

### LINKED ISSUE #1088: Instability related to SlabOcean
The change to the `SlabOcean` introduced in #1079 causes large visible differences in the ice volume over a year that can result in rapid, nonphysical ice growth.

Image

This difference is caused by the placement of parentheses in [`SlabOcean.cpp`](https://github.com/nextsimhub/nextsimdg/commit/554a7262969d2884df6a6b7c4f9880d31c8c37d8#diff-e3b961e4a38dd1b7c6e7becc678bb41a5e83aa541a77f2af748f20172421ec5e):
```
- sssSlab[i] = sss[i] + (sss[i] * fwFlux[i] - fdw[i] * dt) / denominator;
+ sssSlab[i] = sss[i] + (sss[i] * (fwFlux[i] - fdw[i]) * dt) / denominator;
```
I checked in with @timspainNERSC before about this change, so the issue might be something else and this just makes it visible.

The rapid growth was only observed with 6.25km resolution. To reproduce it more quickly I switched to DG1 but it is independent of the dynamics. All runs where done in double with the Kokkos GPU backend, but the latter should not be necessary.

einola:
# Fix salinity calculation in slab ocean

Fixes #1088 

### Task List

---
# Change Description

The code wasn't using the salt flux calculated by IOCeanBoundary::mergeFluxes. This may be causing problems in the spring. Not fully tested yet!

---
# Test Description

It doesn't crash immediately, but I need @Thanduriel's help to test it on his setup.

---
# Documentation Impact

None

---
# Other Details

None

### DIFF
--- a/core/src/ModelArray.cpp
+++ b/core/src/ModelArray.cpp
@@ -342,7 +342,8 @@ void ModelArray::checkLimits(const ModelArray& mask) const
 
     throw std::runtime_error("Field contains out-of-bounds value(s), " + std::to_string(value)
         + " not in [" + std::to_string(lowerPhysicalLimit) + ","
-        + std::to_string(upperPhysicalLimit) + "]. Error at index " + locStr + ".");
+        + std::to_string(upperPhysicalLimit) + "]. Error at " + locStr + " and index "
+        + std::to_string(i) + ".\n");
 }
 
 void ModelArray::validateMaps()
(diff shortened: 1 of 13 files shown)


Thanduriel (approved):
Works for me. In fact, DG2 looks ok as well. I am still running with a time-step of 5min and 120 BBM sub-steps.

Image

A closer look at the solutions in december:
Image

PR #1112, opened 2026-08-19
Fix the Docker image for the MPI tests

### LINKED ISSUE #1110: Fix the failing CI for MPI
Following the new GitLab-based versions of the dev containers in #1104 the tests that use the MPI are failing. 

The reason is that the `LD_LIBRARY_PATH` variable set in the `entrypoint.sh` is not visible in the GitHab actions jobs steps.

I am not sure why it is exactly since the initialisation seems to work fin with one just runs the docker image locally... 

The GitHub docs are no very helpful in this regard. 

Perhaps the most robust solution will be to patch the `ncgen` binary directly instead of trying to fix it up by `LD_LIBRARY_PATH`?

Mikolaj-A-Kowalski:
# Fix the Docker image for the MPI tests

Fixes #1110 

### Task List
- [ ] ~~Defined the tests that specify a complete and functioning change~~
- [ ] ~~The documentation has been updated (or an issue has been created to do so)~~
- [ ] Relevant labels (e.g., enhancement, bug) have been applied to this PR

---
# Change Description

In 4cc4ec7d a value was set to LD_LIBRARY_PATH in the Docker image to resolve the known Spack issue with the RPATH in the ncgen utility.

However, the value was set in the entrypoint script. While it works fine if running the container interactively, in GitHub CI the value of the environmental variable is not propagated into individual job steps, which causes any MPI build to fail.

The solution is to properly bake the value into the image using the `ENV` Dockerfile instruction.

---
# Test Description

At the moment we don't have any good provision to test Docker containers.

---
# Documentation Impact

The changes are fully internal and do not impact user-facing documentation? 

---
# Other Details

### DIFF
--- a/Dockerfiles/Dockerfile.devenv
+++ b/Dockerfiles/Dockerfile.devenv
@@ -70,12 +70,6 @@ ENV LD_LIBRARY_PATH=/opt/views/view/lib
 # Unanswered prompts may lead to failures
 ENV DEBIAN_FRONTEND=noninteractive
 
-# We need to explicitly point ncgen to the view libraries
-# to resolve a problem of location of HDF5 libraries not being provided
-# in the executable's RPATH. See:
-# https://github.com/spack/spack/issues/52555
-ENV LD_LIBRARY_PATH=/opt/views/view/lib
-
 # Create entrypoint script
 RUN { \
       echo '#!/bin/sh'; \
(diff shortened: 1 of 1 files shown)


einola (approved):
Good work, and it's important to have the CI up and running again.

